PR #43, opened 2018-03-30
Add collection deletion API functions

### LINKED ISSUE #32: Ensure Collection deletion API coverage similar to Android
We support the following Collection deletion methods on Android:

```
AzureData.deleteCollection (collectionId, databaseId) {
    // successfully deleted == it.isSuccessful
}

database.deleteCollection (collection) {
    // successfully deleted == it.isSuccessful
}

database.deleteCollection(collectionId) {
    // successfully deleted == it.isSuccessful
}

collection.delete {
    // successfully deleted == it.isSuccessful
}
```

Need to:

- [ ] Make sure all APIs/overloads are supported in iOS library
- [ ] Add examples to the [docs](https://github.com/Azure/Azure.iOS/wiki/AzureData)

ftchirou:
Fixes #32 .

- Add collection deletion functions
- Update README

### DIFF
--- a/AzureData/README.md
+++ b/AzureData/README.md
@@ -119,12 +119,24 @@ database.get (collectionWithId: id) { r in
 
 #### Delete
 ```swift
-AzureData.delete (collection, from: databaseId) { s in
-    // s == successfully deleted
+AzureData.delete (collection, fromDatabase: databaseId) { r in
+    // successfully deleted == r.isSuccess
 }
 
-database.delete (collection) { s in
-    // s == successfully deleted
+AzureData.delete (collectionWithId: collectionId, fromDatabase: databaseId) { r in
+    // successfully deleted == r.isSuccess
+}
+
+database.delete (collection) { r in
+    // successfully deleted == r.isSuccess
+}
+
+database.delete (collectionWithId: collectionId, fromDatabase: databaseId) { r in
+    // successfully deleted == r.isSuccess
+}
+
+collection.delete { r in
+    // successfully deleted == r.isSuccess
 }
 ```
 
--- a/AzureData/README.md
+++ b/AzureData/README.md
@@ -119,7 +119,7 @@ database.get (collectionWithId: id) { r in
 
 #### Delete
 ```swift
-AzureData.delete (collection, fromDatabase: databaseId) { r in
+AzureData.delete (collection) { r in
     // successfully deleted == r.isSuccess
 }
 
@@ -131,7 +131,7 @@ database.delete (collection) { r in
     // successfully deleted == r.isSuccess
 }
 
-database.delete (collectionWithId: collectionId, fromDatabase: databaseId) { r in
+database.delete (collectionWithId: collectionId) { r in
     // successfully deleted == r.isSuccess
 }
 
--- a/AzureData/Source/AzureData.swift
+++ b/AzureData/Source/AzureData.swift
@@ -104,9 +104,13 @@ public func get (collectionWithId collectionId: String, inDatabase databaseId: S
 }
 
 // delete
-//public func delete (_ collection: DocumentCollection, fromDatabase databaseId: String, callback: @escaping (DataResponse) -> ()) {
-//    return DocumentClient.default.delete (collection, fromDatabase: databaseId, callback: callback)
-//}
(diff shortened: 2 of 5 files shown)


AzureData/README.md
```diff
@@ -119,12 +119,24 @@ database.get (collectionWithId: id) { r in
...
-    // s == successfully deleted
+AzureData.delete (collectionWithId: collectionId, fromDatabase: databaseId) { r in
+    // successfully deleted == r.isSuccess
+}
+
+database.delete (collection) { r in
+    // successfully deleted == r.isSuccess
+}
+
+database.delete (collectionWithId: collectionId, fromDatabase: databaseId) { r in
```
colbylwilliams: Don't think this override is necessary, if we have the `database` object, we should be able to get e the id from there, shouldn't have to pass it in manually.   (if this is the way it is in Android, we should let them know)

AzureData/README.md
```diff
@@ -119,12 +119,24 @@ database.get (collectionWithId: id) { r in
 
 #### Delete
 ```swift
-AzureData.delete (collection, from: databaseId) { s in
-    // s == successfully deleted
+AzureData.delete (collection, fromDatabase: databaseId) { r in
```
colbylwilliams: Don't think this override is necessary, if we have the `collection` object, we should be able to just use the generic `delete(resource)` function and don't need the `databaseId`... correct?

AzureData/Source/AzureData.swift
```diff
@@ -104,9 +104,13 @@ public func get (collectionWithId collectionId: String, inDatabase databaseId: S
 }
 
 // delete
-//public func delete (_ collection: DocumentCollection, fromDatabase databaseId: String, callback: @escaping (DataResponse) -> ()) {
-//    return DocumentClient.default.delete (collection, fromDatabase: databaseId, callback: callback)
-//}
+public func delete (_ collection: DocumentCollection, fromDatabase databaseId: String, callback: @escaping (DataResponse) -> ()) {
```
colbylwilliams: Same here, if we have the `collection` I don't think we should need to pass in the `databaseId`.

AzureData/Source/AzureDataExtensions.swift
```diff
@@ -46,10 +46,13 @@ public extension Database {
     }
     
     //delete
-    public func delete (_ resource: DocumentCollection, callback: @escaping (DataResponse) -> ()) {
-        return DocumentClient.default.delete(resource, callback: callback)
+    public func delete (_ collection: DocumentCollection, callback: @escaping (DataResponse) -> ()) {
+        return DocumentClient.default.delete(collection, fromDatabase: self.id, callback: callback)
```
colbylwilliams: same here

AzureData/Source/DocumentClient.swift
```diff
@@ -187,11 +187,17 @@ public class DocumentClient {
     }
     
     // delete
-//    public func delete (_ collection: DocumentCollection, fromDatabase databaseId: String, callback: @escaping (DataResponse) -> ()) {
-//
-//        return self.delete(DocumentCollection.self, at: resourceLocation, callback: callback)
-//    }
-    
+    public func delete (_ collection: DocumentCollection, fromDatabase databaseId: String, callback: @escaping (DataResponse) -> ()) {
+        return self.delete(collectionWithId: collection.id, fromDatabase: databaseId, callback: callback)
```
colbylwilliams: Same here, if we have the `collection` I don't think we should need to pass in the `databaseId`... can we remove this override all together?

AzureData/Source/AzureData.swift
ftchirou: I agree. They have this override on Android and thought to add it for parity.

AzureData/README.md
ftchirou: Oh! Yes. I actually haven't implemented this override. Just made a mistake while updating the README.

AzureData/README.md
ftchirou: I agree and thought the same. This override exists on Android. Should we keep it for parity with Android?

colbylwilliams: no, we should ping @NateRickard to remove it from android unless he has a reason not to

NateRickard: I think this is the one where the doc was off a bit but let me know if I'm missing something.

PR #70, opened 2018-04-26
Review offline caching and add tests

### LINKED ISSUE #69: Investigate and fix issue with cached resources not being fetched while offline
## Bug Information
Cached resources are not being fetched from the local filesystem when the device is offline.

### Steps to Reproduce
- Create a simple iOS app and add `AzureData.framework` in its `Linked Frameworks and Libraries`
- Run the app on a simulator or a physical device
- Make sure the simulator or the device is connected to the internet
- Call `AzureData.get(collectionsIn: database, callback: callback)` to fetch the collections from the specified database. The fetched collections will be cached locally.
- Disconnect the simulator or the device from the internet
- Call `AzureData.get(collectionsIn: database, callback: callback)` again while the simulator or device is not connected to the internet

### Expected Behavior
`AzureData.get(collectionsIn: database, callback: callback)` should return the collections cached locally through the callback.

### Actual Behavior
`AzureData.get(collectionsIn: database, callback: callback)` returns `nil` through the callback.

ftchirou:
Fixes #69 and #71.

- Fix the issue preventing some cached resources from being loaded when the network is not reachable
- Delete a resource from the cache if the remote database returns `404`
- Refactor `ReachabilityManager` to make it testable
- Add tests for the offline capabilities

### DIFF
--- a/AzureCore/AzureCore.xcodeproj/project.pbxproj
+++ b/AzureCore/AzureCore.xcodeproj/project.pbxproj
@@ -11,6 +11,10 @@
 		3259AE80206C5519002EF09B /* KeychainAccess.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3259AE7F206C5519002EF09B /* KeychainAccess.framework */; };
 		3259AE82206C5523002EF09B /* KeychainAccess.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3259AE81206C5523002EF09B /* KeychainAccess.framework */; };
 		3259AE84206C5531002EF09B /* KeychainAccess.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3259AE83206C5531002EF09B /* KeychainAccess.framework */; };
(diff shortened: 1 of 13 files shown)


ftchirou:
Yes. The underscore was temporary and was just there to signify that this may replace `AzureDataTests` at some point. Is it okay if I rename it `AzureTests` for now (so it doesn't clash with `AzureDataTests`)?

colbylwilliams:
Ah, I thought you already replaced `AzureDataTests`... Can you write up a super brief summary explaining the changes/difference between `AzureDataTests` and `_AzureDataTests` and what would prevent us from going ahead and replacing `AzureDataTests`?

ftchirou:
I haven't yet replaced `AzureDataTests`. I'd need to update practically all the tests. Maybe I create an issue for that? That way, I can write the summary you requested as a comment on that issue.

The main difference is that each test case is now responsible of ensuring that a resource exists by calling one of the `ensure...Exists` functions at the beginning of the test case. As opposed to having a global `ensureDatabase = true` set for all the test cases in the `setup` function.

colbylwilliams:
@ftchirou sounds good.  Please create the issue (and summary) and reference it in this PR.

ftchirou:
Issue #79 created

PR #76, opened 2018-05-02
Custom data encryption

### LINKED ISSUE #72: Data encryption for local persistence
Expose a user-facing API (likely a protocol with `encrypt` and `decrypt` functions) to allow user to implement custom data encryption.

ftchirou:
Fixes #72.

- Allow users to implement custom data encryption

### DIFF
--- a/AzureData/Source/ResourceCache.swift
+++ b/AzureData/Source/ResourceCache.swift
@@ -9,11 +9,6 @@
 import Foundation
 import AzureCore
 
-protocol DataEncryptor {
-    func encrypt(_ data: Data) -> Data
-    func decrypt(_ data: Data) -> Data
-}
-
 public class ResourceCache {
     
     static var isEnabled = true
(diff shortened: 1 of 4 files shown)


AzureData/Source/ResourceCache.swift
```diff
@@ -9,6 +9,11 @@
 import Foundation
 import AzureCore
 
+protocol DataEncryptor {
```
colbylwilliams: Let's rename `DataEncryptor` to `ResourceEncryptor` and break it out into a separate file.  It will also need to be made `public`.

AzureData/Source/ResourceCache.swift
```diff
@@ -34,15 +39,16 @@ public class ResourceCache {
 
     
     static var dispatchQueue: DispatchQueue { return DispatchQueue.global(qos: .default) }
-    
+
+    static var dataEncryptor: DataEncryptor? = nil
```
colbylwilliams: We'll need to add a pass-through property to `AzureData.swift` to allow the user to set this `dataEncryptor` (or `resourceEncryptor`).  Something along the lines of:

```swift
public var resourceEncryptor: ResourceEncryptor {
    get { return ResourceCache.resourceEncryptor }
    set { ResourceCache.resourceEncryptor = newValue }
}
```

PR #77, opened 2018-05-02
Configure AzureData from plists

ftchirou:
- Add the option to provide the Azure account name and master key using property lists.

### DIFF
--- a/AzureCore/Source/AzureKeys.swift
+++ b/AzureCore/Source/AzureKeys.swift
@@ -1,0 +1,95 @@
+//
+//  AzureKeys.swift
+//  AzureCore
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+public class AzureKeys: Codable {
+    enum CodingKeys: String, CodingKey {
+        case accountName = "AzureAccountName"
+        case masterKey = "AzureAccountMasterKey"
+    }
+
+    public let accountName: String
+    public let masterKey: String
+
+    public var hasValidAccountName: Bool {
+        return !accountName.trimmingCharacters(in: .whitespaces).isEmpty
+    }
+
+    public var hasValidMasterKey: Bool {
+        return !masterKey.trimmingCharacters(in: .whitespaces).isEmpty
+    }
+
+    private init?(from info: [String: Any]) {
+        guard let accountName = info[CodingKeys.accountName.rawValue] as? String else { return nil }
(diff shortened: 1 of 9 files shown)


AzureCore/Source/AzureKeys.swift
```diff
@@ -0,0 +1,95 @@
...
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+public class AzureKeys: Codable {
+    enum CodingKeys: String, CodingKey {
+        case accountName = "AzureAccountName"
```
colbylwilliams: These two keys are specific to Cosmos DB, rename the keys to:
```swift
cosmosAccountName // coding key = "AzureCosmosDbDatabaseAccountName"
cosmosMasterKey // coding key = "AzureCosmosDbDatabaseAccountMasterKey"
```

AzureCore/Source/AzureKeys.swift
```diff
@@ -0,0 +1,95 @@
...
+        case masterKey = "AzureAccountMasterKey"
+    }
+
+    public let accountName: String
+    public let masterKey: String
+
+    public var hasValidAccountName: Bool {
```
colbylwilliams: obviously you'll need to update these property names to correspond to the updated cosmos-specific names

AzureCore/Source/AzureKeys.swift
```diff
@@ -0,0 +1,95 @@
...
+        return !accountName.trimmingCharacters(in: .whitespaces).isEmpty
+    }
+
+    public var hasValidMasterKey: Bool {
```
colbylwilliams: We should be checking for default values like [here](https://github.com/colbylwilliams/CustomVision/blob/master/CustomVision/CustomVision/CustomVisionKeys.swift)

AzureData/README.md
```diff
@@ -22,6 +24,25 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 }
 ```
 
+## Using a .plist
+
```
colbylwilliams: This will need to be updated with the updated key names, and we should explicitly call out that adding master keys to a plist file **shouldn't be used in production** like [here](https://github.com/colbylwilliams/CustomVision#plist-file)

AzureCore/Source/AzureKeys.swift
colbylwilliams: Also, let's add a Azure.plist file with the default values.  Then we can add a custom `AzureTests.plist` file, use the `configure` override that takes the plist name, and ignore the `AzureTests.plist`

PR #80, opened 2018-05-21
Offline Write

### LINKED ISSUE #73: Implement offline write
- [x]  Add WRITE support when offline
- [x] Queue writes when offline
- [x] Automatically push any local changes as soon as device comes back online
- [x] Send global notification to let user know data changed

ftchirou:
Fixes #73.

- Queue write operations (create, replace and delete) while offline
- Perform write operations locally while offline
- Automatically push local changes to the remote database as soon as the network becomes reachable
- Notify observers when pending writes are performed online
- Add tests


SogoGolf:
whoa this is exciting stuff! can i ask, will offline write be exposed via the Azure.iOS api at some point (so we can plan ahead) ?

colbylwilliams:
Hi @SogoGolf, offline write is actually enabled by default in Azure.iOS ([AzureData](https://github.com/Azure/Azure.iOS/tree/master/AzureData)).  It should "just work" the same offline as it does online.  You can opt out of offline caching by setting `AzureData.offlineDataEnabled = false`.

PR #83, opened 2018-05-31
Add podspec file

bfaist:
Allows Azure.iOS to be installed from projects using Cocoapods


bfaist:
Closing pull request.  This allows cocoapods to install Azure.iOS but does not appear to build correctly. 
Module AzureCore and AzureData are not able to be imported for use in swift code.  Going to try to migrate to Carthage which works.

colbylwilliams:
Thanks @bfaist we're finishing up compatibility with objective-c now and I'll get the libs up on cocoapods right after that.  Probably not quick enough to help you now, but I'll ping you when they're up anyway.  In the meantime, the example app in this repo uses Carthage if you want to use it as reference.  I'll also create a new tag today so you'll get the latest.

PR #94, opened 2018-06-09
Create empty child directory for offline resources

### LINKED ISSUE #93: Create empty child directories for resources cached locally
Currently, we create a locally cached resource's child directories as needed.  For example, a collection resource cached at the following path:

`com.azure.data/dbs/xOptAA==/colls/xOplAM2GIwE=`

Subdirectories for the collection's children (`docs`, `sprocs`, `udfs`, `triggers`) will not be created until the corresponding directory is required to cache one of these children locally (for example getting a document from the server).

This poses an issue when the user tries to list one of this collection's child resources while offline as the path that we're querying (the directory below) doesn't exist.

For example (offline):
```swift
collection.get(documentsAs: Document.self) { r in
    // r.error == directory does not exist at: "com.azure.data/dbs/xOptAA==/colls/xOplAM2GIwE=/docs"
}
```

Not sure if the solution is to always create empty directories for a each resource's child types.  I'm open to discussing alternatives.  But this is currently throwing an error.

ftchirou:
Fixes #93.


AzureData/Source/ResourceType.swift
```diff
@@ -49,7 +49,22 @@ public enum ResourceType : String {
         case .collection, .document, .storedProcedure, .trigger, .udf, .attachment: return true
         }
     }
-    
+
+    var childrens: [ResourceType] {
```
colbylwilliams: Can we rename `childrens` to `children`

PR #103, opened 2018-10-11
Fix project reference in workspace

elektrojunge:
Hi,

I just checked out the project and found that the `.xcworkspace` contained an invalid  reference to the AzureData ObJC project file so I fixed it.


ftchirou:
Can you merge these changes? 🙏  @NateRickard

PR #124, opened 2019-05-02
Add support for partition keys when executing stored procedures

### LINKED ISSUE #121: partition key error calling stored proc (v0.3.0)
tried to call a test stored proc and saw the following error. is this because the stored proc API methods need a partition key param added?

```BadRequest | PartitionKey extracted from document doesn't match the one specified in the header```

**API call**
```
AzureData.execute(storedProcedureWithId: "purchaseTokens", usingParameters: [], inCollection: "collection", inDatabase: "db") {
	result in
	print(result)
}

```

**Stored Proc** (setup in portal)
```
function purchaseTokens() {

    var collection = getContext().getCollection();

    var isAccepted = collection.queryDocuments(
        collection.getSelfLink(),
        `SELECT f.id FROM f WHERE f.type = 'entity'`,

        function (err, data, options) {
            if (err) throw err;

            if (!data || !data.length) {
                var response = getContext().getResponse();
                response.setBody([{hmm: "nothin there"}]);
            }
            else {
                var response = getContext().getResponse();
                response.setBody({sku: "sku"});
            }
        });

    if (!isAccepted) throw new Error('The query was not accepted.');
}
```

ftchirou:
Fixes #121.

- Adds overloads of `execute(storedProcedure...)` with a new parameter for the partition key
- Updates the tests


NateRickard:
[Travis build](https://travis-ci.org/Azure/Azure.iOS) also passed for this PR, not sure why it's not syncing here correctly.  Merging.

PR #155, opened 2019-10-04
Logging policy

### LINKED ISSUE #149: AzureCore Logging Policy

- [x] Implement policy according to guidelines
- [x] Add logging util for use by client SDKs?
- [ ] Add test coverage

bsiegel:
Adds a logging framework and some pipeline logging policies.

Fixes #149

### DIFF
--- a/AzureCore/Source/Pipeline/PipelineBase.swift
+++ b/AzureCore/Source/Pipeline/PipelineBase.swift
@@ -8,15 +8,16 @@
 
 import Foundation
 
-public typealias HttpResultHandler<T> = (Result<T, Error>, HttpResponse) -> Void
-public typealias PipelineStageResultHandler = HttpResultHandler<PipelineResponse>
+public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HttpResponse) -> Void
+public typealias HttpResultHandler<T> = ResultHandler<T, Error>
+public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
 
 public protocol PipelineStageProtocol {
     var next: PipelineStageProtocol? { get set }
 
     func onRequest(_ request: inout PipelineRequest)
     func onResponse(_ response: inout PipelineResponse)
-    func onError(_ error: Error) -> Bool
+    func onError(_ error: PipelineError) -> Bool
 
     func process(request: inout PipelineRequest, completion: @escaping PipelineStageResultHandler)
 }
@@ -24,7 +25,7 @@ public protocol PipelineStageProtocol {
 extension PipelineStageProtocol {
     public func onRequest(_ request: inout PipelineRequest) {}
     public func onResponse(_ response: inout PipelineResponse) {}
-    public func onError(_ error: Error) -> Bool { return false }
+    public func onError(_ error: PipelineError) -> Bool { return false }
 
     public func process(request: inout PipelineRequest, completion: @escaping PipelineStageResultHandler) {
         self.onRequest(&request)
@@ -34,7 +35,7 @@ extension PipelineStageProtocol {
                 self.onResponse(&pipelineResponse)
                 completion(.success(pipelineResponse), httpResponse)
             case .failure(let error):
-                let handled = self.onError(error, fromResponse: httpResponse)
+                let handled = self.onError(error)
                 if !handled {
                     completion(.failure(error), httpResponse)
                 }
--- a/AzureCore/Source/Pipeline/PipelineBase.swift
+++ b/AzureCore/Source/Pipeline/PipelineBase.swift
@@ -16,7 +16,7 @@ public protocol PipelineStageProtocol {
 
     func onRequest(_ request: inout PipelineRequest)
     func onResponse(_ response: inout PipelineResponse)
-    func onError(request: PipelineRequest) -> Bool
+    func onError(_ error: Error) -> Bool
 
     func process(request: inout PipelineRequest, completion: @escaping PipelineStageResultHandler)
 }
@@ -24,7 +24,7 @@ public protocol PipelineStageProtocol {
 extension PipelineStageProtocol {
     public func onRequest(_ request: inout PipelineRequest) {}
     public func onResponse(_ response: inout PipelineResponse) {}
-    public func onError(request: PipelineRequest) -> Bool { return false }
+    public func onError(_ error: Error) -> Bool { return false }
 
     public func process(request: inout PipelineRequest, completion: @escaping PipelineStageResultHandler) {
         self.onRequest(&request)
@@ -34,7 +34,10 @@ extension PipelineStageProtocol {
                 self.onResponse(&pipelineResponse)
                 completion(.success(pipelineResponse), httpResponse)
             case .failure(let error):
-                completion(.failure(error), httpResponse)
+                let handled = self.onError(error, fromResponse: httpResponse)
+                if !handled {
+                    completion(.failure(error), httpResponse)
+                }
             }
         })
     }
--- a/AzureCore/Source/Pipeline/PipelineLogger.swift
+++ b/AzureCore/Source/Pipeline/PipelineLogger.swift
@@ -7,3 +7,99 @@
 //
 
 import Foundation
+import os.log
+
+public enum PipelineLogLevel: Int {
+    case error, warning, info, debug
+}
+
+public protocol PipelineLogger {
+    func debug(_: String)
+    func info(_: String)
+    func warning(_: String)
+    func error(_: String)
+
+    func log(_: String, atLevel: PipelineLogLevel)
+}
+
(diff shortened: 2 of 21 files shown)


AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -0,0 +1,133 @@
...
+    public func onRequest(_ request: inout PipelineRequest) {
+        let req = request.httpRequest
+        log.debug("Request URL: \(req.url)")
+        log.debug("Request method: \(req.httpMethod.rawValue)")
+        log.debug("Request headers:")
+        for (header, value) in req.headers {
+            log.debug("    \(header): \(value)")
+        }
+        log.debug("Request body:")
+        log.debug(loggableBodyFor(headers: req.headers, bodyText: req.text))
```
tjprescott: Gives some guidance on when and where prepositions should appear in argument labels:
https://swift.org/documentation/api-design-guidelines/#argument-labels

This seems like the appropriate use, most like the "move(toX:y)" example.

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    public func onRequest(_ request: inout PipelineRequest) {
+        let req = request.httpRequest
+        log.debug("Request URL: \(req.url)")
+        log.debug("Request method: \(req.httpMethod.rawValue)")
+        log.debug("Request headers:")
+        for (header, value) in req.headers {
+            log.debug("    \(header): \(value)")
```
tjprescott: We likely need to scrub certain headers, like "Authorization" for instance.

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -0,0 +1,133 @@
...
+        log.debug("Response headers:")
+        for (header, value) in res.headers {
+            log.debug("    \(header): \(value)")
+        }
+        log.debug("Response content:")
+        log.debug(loggableBodyFor(headers: res.headers, bodyText: res.text))
+    }
+
+    public func onError(_ error: Error) -> Bool {
+        log.error(error.localizedDescription)
```
tjprescott: This should be fine for transport errors. We should consider when we would want a service error to be detected and whether a deserialized error should be passed through the onError section of the logging policy. The way current pipelines are set up, I don't believe they would.

AzureCore/Source/Pipeline/Transport/HttpRequest.swift
```diff
@@ -42,6 +42,11 @@ public class HttpRequest {
         self.data = data
     }
 
+    public func text(encoding: String.Encoding = .utf8) -> String? {
```
tjprescott: Would it make sense to have HttpRequest and HttpResponse both implement some kind of HttpCommon protocol that has this method as an extension? In that way, we'd avoid potential double maintenance.

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -0,0 +1,133 @@
...
+        let captures = match.capturedValues(from: header)
+        if captures.count > 1 {
+            return captures[1]
+        } else {
+            return nil
+        }
+    }
+}
+
+public class CurlFormattedRequestLoggingPolicy: PipelineStageProtocol {
```
tjprescott: Is there a use case for this or was it a carry-over from Python?

colbylwilliams:
@bsiegel @tjprescott consider updating to use closures and update the api you've already written to use autoclosures.  You can see how Willow does something similar [here](https://github.com/Nike-Inc/Willow#single-line-closures).  See this snippet from Willow's readme for a brief explanation of why:

> Closures defer the execution of all the logic inside the closure until absolutely necessary, including the string evaluation itself. In cases where the Logger instance is disabled, log execution time was reduced by 97% over the traditional log message methods taking a String parameter. Additionally, the overhead for creating a closure was measured at 1% over the traditional method making it negligible.

So essentially if you don't have debug or info enabled, those messages won't slow your code down.

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -0,0 +1,133 @@
...
+    private lazy var attachmentRegex = NSRegularExpression("attachment; ?filename=([\"\\w.]+)",
+                                                           options: .caseInsensitive)
+
+    public init(logger: PipelineLogger? = nil) {
+        if let logger = logger {
+            self.log = logger
+        } else if #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) {
+            self.log = PipelineOSLogAdapter()
+        } else {
+            self.log = PipelineNSLogger()
```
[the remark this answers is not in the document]
bsiegel: I think the approach I'd like to go with here is to always have a single logger. However that logger can do various types of things:

1. Manually manage its configured verbosity and write log messages to a sink (eg NSLogger)
2. Act as a proxy to another logging framework (eg OSLogger)
3. Any combination of one or more of each #1 and #2 above

If for example a customer would like to log to the OSLogger and also ship log messages to a web service, they can create a new Logger that contains both an OSLogger and your custom logging logic, and uses both inside its `log` method.

Also if a customer would like to disable logging entirely, they can provide a null logger (null object pattern, not `nil` - which I think I was planning to provide out of the box).

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
bsiegel: It's brand new for this implementation. In my experience debugging mobile applications it's incredibly useful to be able to have this output during development, and I always seem to end up writing / porting one of these to every mobile application I've worked on so I figured it would be simple to provide one. It would generally be enabled by providing this policy in addition to, rather than instead of, the regular logging policy.

AzureCore/Source/Pipeline/PipelineLogger.swift
```diff
@@ -0,0 +1,105 @@
...
+//
+
+import Foundation
+import os.log
+
+public enum PipelineLogLevel: Int {
+    case error, warning, info, debug
+}
+
+public protocol PipelineLogger {
```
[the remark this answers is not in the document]
bsiegel: Yep, I'm refactoring to change how the logger is provided, and currently it's named "ClientLogger". Another option is simply "Logger" but I didn't want to conflict with other code that might exist in the customer's app or another 3rd party library.

AzureCore/Source/Pipeline/PipelineBase.swift
```diff
@@ -34,7 +34,10 @@ extension PipelineStageProtocol {
                 self.onResponse(&pipelineResponse)
                 completion(.success(pipelineResponse), httpResponse)
             case .failure(let error):
-                completion(.failure(error), httpResponse)
+                let shouldAbort = self.onError(error)
+                if !shouldAbort {
```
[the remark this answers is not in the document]
bsiegel: Fixed in 809cc65

AzureCore/Source/Pipeline/Transport/HttpRequest.swift
bsiegel: Fixed in e427da2

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
bsiegel: Fixed in f9cfac2

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
bsiegel: Fixed in f976ebb

bsiegel:
> @bsiegel @tjprescott consider updating to use closures and update the api you've already written to use autoclosures.

Fixed in b684027

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
tjprescott: Cool, interested in checking it out

AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
tjprescott: @bsiegel thoughts on this? Also happy to merge this now and defer this discussion.

bsiegel: I think it's outside the scope of logging, but yes we should have a chat about how to represent protocol errors vs. transport errors (and further, retryable errors vs. fatal errors). I'd want to understand how the other languages currently make this distinction so I should probably take some time to read thru their codebases / guidelines on this topic.

PR #159, opened 2019-11-18
[Storage] Multi-chunk stream downloader

tjprescott:
Implements a multi-chunk stream downloader for Azure Storage blobs.


AzureStorageBlob/Source/Models/BlobProperties.swift
```diff
@@ -151,6 +151,48 @@ public final class BlobProperties: XMLModelProtocol {
             "RemainingRetentionDays": XMLMetadata(jsonName: "remainingRetentionDays"),
         ])
     }
+
+    internal init(from headers: HttpHeaders) {
+        self.creationTime = headers["x-ms-creation-time"].asDate
```
bsiegel: Should we convert all of these hardcoded strings to cases of `HttpHeader`? We'd need to add some cases but I feel like that's a best practice. If there are `x-` headers custom to Storage, we could expose those in a `StorageHttpHeader` enum.

AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -256,47 +300,52 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
+    }
+
+    /**
+     Create a simple URL for a blob.
+     - Parameter blob: Name of the blob.
+     - Parameter container: Name of the container.
+     - Returns: The URL of the blob.
+     */
+    public func url(forBlob blob: String, inContainer container: String) -> URL? {
+        var urlString = baseUrl
```
bsiegel: Is there no URLBuilder or similar thing that lets us add path components? Just always hate seeing manual path manipulation.

AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -256,47 +300,52 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
+    /**
+     Download a blob from a specific container.
+     - Parameter url: A URL to a blob to download.
+     - Parameter options: A `DownloadBlobOptions` object to control the download operation.
+     - Parameter completion: An `HttpResultHandler` closure that returns a `StorageStreamDownloader` object on success.
+     */
+    public func download(url: URL, withOptions options: DownloadBlobOptions? = nil,
+                         then completion: @escaping HttpResultHandler<StorageStreamDownloader>) throws {
+        let (host, container, blob) = try parse(url: url)
+        // ensure that if the base URL is altered temporarily because of the URL that it is switched back.
```
bsiegel: Could maybe use more description in this comment. From what I have gleaned:

* `baseUrl` is a field (comes from the parent class)
* When calling `download(url:)` we temporarily change the `baseUrl` field and then change it back after the call is complete

If that's correct, I have two comments. One, and this is purely a style thing, but I was confused about where `baseUrl` came from until I realized it was probably a field on the parent class. How do you feel about prefixing fields with `self.`? Two, it seems like this means that each call to `download(url:)` on this client changes the internal state of the client. I feel like this would have to cause problems, for example if I have a client initialized for one storage account, then call `download(url:)` to download a file from a different account (host). While that operation is in flight, it seems like calls to other client methods will go to the wrong place?

AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -0,0 +1,435 @@
...
+                if let contentMD5 = headers["Content-MD5"] {
+                    let dataHash = data.md5
+                    guard contentMD5 == dataHash else {
+                        let error = AzureError.general("Block MD5 \(dataHash) did not match \(contentMD5).")
+                        completion(.failure(error), httpResponse)
+                        return
+                    }
+                }
+                if let contentCRC64 = headers["Content-CRC64"] {
+                    let dataHash = data.crc64
```
bsiegel: If we get here, this is always going to fail right? Because we don't have an implementation of crc64 yet. Should we comment this block out?

AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -0,0 +1,435 @@
...
+                    guard contentCRC64 == dataHash else {
+                        let error = AzureError.general("Block CRC64 \(dataHash) did not match \(contentCRC64).")
+                        completion(.failure(error), httpResponse)
+                        return
+                    }
+                }
+                let decryptedData = self.decrypt(data)
+                do {
+                    let handle = try self.openFileForWriting()
```
bsiegel: Should we move the call to `closeFile()` up into a `defer` block here?

bsiegel: Specifically, since opening the file itself is an operation that might fail, it would be something like
`defer { handle?.closeFile() }`

AzureStorageBlob/Source/Models/BlobProperties.swift
tjprescott: I think that's a good idea. Unfortunately, you can't extend enums in Swift to add cases. 

What we can do is have each library declare a Headers file with its unique headers, and give it the same extensions as the AzureCore one. Common headers would bubble up to AzureCore.

AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Agreed. I rewrote this using `URL.appendPathComponent`.

AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Swift best practices are that you don't use self unless it is necessary (i.e. in a closure or when a local variable has the same name and you want to target the property). Swiftlint will whine incessantly if you do this.

You raise a great point about messing with the baseUrl. Instead, I changed this to actually instantiate a new client in the event that the baseUrl doesn't match the URL. The scenario here is when you want to download a public blob URL. I added a TODO to reconsider this implementation and testing the scenario should be informative.

AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
tjprescott: Yes, good catch! In Swift the best time to put a defer statement is right after the one that makes it necessary in the first place. Fixed.

AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
tjprescott: I think the TODO is okay for now. Prior to releasing Preview 1, we would resolve all of the TODOs, either by providing the implementation or removing the capability. Once these PRs are merged, I'll create issues for each of these outstanding TODOs.

tjprescott:
I'm actually making the changes from this PR onto the MSAL PR, so I have made the changes based on the feedback and they will be reflected in that PR. For now, i'll merge this so I can continue to the MSAL PR.

PR #169, opened 2019-12-17
Barebones pipelines

mitchdenny:
This is the structural elements of the CI pipeline. Doesn't do anything yet but I wanted to get this merged in so that when I start doing more PRs I can start getting pass/fail in the PRs.


sdk/core/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+  paths:
+    include:
+      - sdk/core/
+
+stages:
+  - template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
+    parameters:
+      ServiceDirectory: core
+      Artifacts:
+        - name: AzureAppConfiguration
```
bsiegel: Should these be AzureCore instead?

mitchdenny: Whoops. Cut and paste error ;)

PR #178, opened 2020-01-09
Add ci configs for appconfig, cog services, storage and text analytics.

mitchdenny:
This PR adds the basic ```ci.yml``` stubs for the folders that existing in the ```sdk/``` folder. Once this is merged those pipelines will be automatically generated and I can start up building out the common template to use xcodebuild etc.


bsiegel:
LGTM, just noting here that in the future we will likely have multiple libraries under `cognitiveservices`, so we will need to add additional artifacts to the config.

PR #179, opened 2020-01-09
xcodebuild with basic args.

mitchdenny:
This PR is building out the basic CI plumbing for the iOS repository by using xcodebuild. It will include the necessary ```archetype-sdk-client.yml``` template changes to drive xcodebuild, as well as changes to the Xcode workspace files to support build for each pipeline.


mitchdenny:
This is ready to merge and when taken with [this PR](https://github.com/Azure/azure-sdk-for-ios/pull/178) will mean we have pipelines for each service directory which produces a bundles artifact that contains the ```.framework``` folder for the targetted libraries.

As part of this PR I added a scheme for each service directory (we should probably consider deleting the existing schemes. If we end up with multiple modules in each service directory we would add those targets to the scheme so that they would also be produced.

.gitignore
```diff
@@ -46,7 +46,7 @@ playground.xcworkspace
 # you should judge for yourself, the pros and cons are mentioned at:
 # https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
 #
-# Pods/
+Pods/
```
bsiegel: I think I am personally okay with this change, but I wanted to understand more the reasoning behind it. Also would like to hear others' (at least @tjprescott) opinions

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -18,6 +18,16 @@ jobs:
 
       - template: tools/daily-dev-build-variable/daily-dev-build-variable.yml@azure-sdk-tools
 
+      - script: |
+          pod install
+          xcodebuild -workspace AzureSDK.xcworkspace -scheme ${{parameters.ServiceDirectory}} -destination "platform=iOS Simulator,OS=13.3,name=iPhone 11 Pro Max" -configuration Release -derivedDataPath $(Build.ArtifactStagingDirectory)
```
bsiegel: I think you generally need to provide the command at the end of this invocation, e.g. "build" or "archive". I'm guessing that since this works, if you omit it there is an implicit one (probably "build") but we should be explicit.

mitchdenny: Added.

.gitignore
mitchdenny: I was torn on this one. On one hand committing the sources to the tree helps insulate us from unreliable dependencies that might get yanked and that might also get changed underneath us (if they are depending on a tag instead of a commit SHA).

However it is a lot of bloat to add into our repository and I'm not sure what the ramifications are from a licensing perspective either. The other thing that I want to avoid is the temptation for someone to go in and change anything in the ```Pods/``` folder since I think that could be a nightmare.

I don't think adding the ignore here is the end of the story. We really do need to consider the potential security ramifications for our engineering system.

tjprescott: https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control

tjprescott: Based on this, I am okay ignoring the Pods folder as well. We don't commit the dependencies in other SDKs.

PR #184, opened 2020-01-16
Core request policies

### LINKED ISSUE #143: AzureCore Telemetry (User-Agent) Policy

- [x] Update User-Agent format according to guidelines
- [x] Explore if possible to find out if client app is primarily Swift or ObjectiveC
- [x] Provide mechanism for client app to set their application ID (name and version) or find way to do this automatically
- [x] ~Provide mechanism for SDK author to specify dynamic telemetry information through "x-ms-azsdk-telemetry" header~
- [ ] Add test coverage

For the user agent, we want to report:

application ID / bundle ID
library name and version
os.version
user.region
user.language
min & target api level / os version
built for debugging
device model & manufacturer
language (swift vs objc)

Determine which of these metrics we can access, then propose an update to the guidelines with the format of the useragent string that we will use to report them. Implement the useragent string to match that guideline.

Also note: The telemetry on-off toggle was not intended to affect the value of the User-Agent header. We don't need to implement any logic inside this policy related to checking the value of that toggle.

### LINKED ISSUE #144: AzureCore Unique Request ID policy

This is still marked as a TODO in the guidelines, but principally there should be a standard way to set the "x-ms-client-request-id" and "x-ms-request-id" properties. The Python pipeline doesn't seem to do this (at least not in Azure core). While it is used in the distributed tracing policy, it isn't ever generated there, merely passed on.

We need to investigate (1) what is required for this policy and (2) where this should be set.

Additionally:
- [ ] Add test coverage

### LINKED ISSUE #163: AzureCore Date policy
Create a policy that adds a Date header, with implementation matching that of other SDK langauges.

bsiegel:
Fixes #163, Fixes #144, Fixes #143

Depends on: #187


sdk/core/AzureCore/Source/Providers/BundleInfoProvider.swift
```diff
@@ -0,0 +1,65 @@
...
+import Foundation
+
+public protocol BundleInfoProvider {
+    var identifier: String? { get }
+    var name: String? { get }
+    var version: String? { get }
+    var minDeploymentTarget: String? { get }
+}
+
+public struct DeviceBundleInfoProvider: BundleInfoProvider {
```
tjprescott: In reference to my other comment about Framework version, consider having a FrameworkBundleInfoProvider or simply a configurable BundleInfoProvider which, if no identifier is provided (to target a framework) simply targets the app itself.

sdk/core/AzureCore/Tests/AddDatePolicyTests.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// Test that the add date policy adds a non-empty date header
+    func test_RequestIdPolicy_AddsHeaderToRequest() {
+        let policy = AddDatePolicy()
+        let req = TestPipelineRequest()
+        policy.onRequest(req) { _ in }
+        let value = req.httpRequest.headers[.date]
+        XCTAssertTrue(value != nil && !value!.isEmpty)
+    }
+
+    /// Test that the add date policy adds a header whose value looks like an RFC1123 date
```
tjprescott: The DateUtil class already has methods for working with RFC1123 dates. We should consolidate that code.

sdk/core/AzureCore/Tests/UserAgentPolicyTests.swift
```diff
@@ -29,29 +29,242 @@ import XCTest
...
-    /// Test that supplying a baseUserAgent without overwrite still captures much of the standard data.
-    func testCustomUserAgentPolicyNoOverwrite() {
-        let policy = UserAgentPolicy(baseUserAgent: "MyAgent")
+    /// Test that the user agent policy creates the correct user agent with the custom useragent initializer
+    func test_UserAgentPolicy_WithCustomUserAgentInitializer() {
+        let customAgent = "MyAgent/5.0"
+        let policy = UserAgentPolicy(userAgent: customAgent)
         let userAgent = policy.userAgent
-        XCTAssertEqual(userAgent, "MyAgent ios/5.0 (iPhone) AzureCore/0.1.0")
+        XCTAssertEqual(userAgent, customAgent)
```
tjprescott: So a custom string hijacks the entire UA string? What if someone wanted to simply append to the standard string? I thought that was a supported use case in Python? Even if it's not, I remember getting queries from partner teams that wanted to do just that.

sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
```diff
@@ -33,6 +33,8 @@ public class AppConfigurationClient: PipelineClient {
         case latest = "2019-01-01"
     }
 
+    public static let version = "1.0"
```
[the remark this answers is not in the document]
bsiegel: Will investigate, definitely would be preferred to maintaining this manually.

sdk/core/AzureCore/Source/Pipeline/PipelineRequest.swift
```diff
@@ -26,7 +26,14 @@
 
 import Foundation
 
-public final class PipelineRequest: PipelineContextProtocol, Copyable {
+public protocol PipelineRequest: PipelineContextProtocol {
```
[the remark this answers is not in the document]
bsiegel: Yep, I have gone back and forth a little bit around how to test this, but especially for the policies, I need to be able to set up a policy, call its pipeline lifecycle methods, and then write assertions based on what the policy changed inside the request and/or response objects.

sdk/core/AzureCore/Source/Pipeline/PipelineRequest.swift
```diff
@@ -26,7 +26,14 @@
...
+    var httpRequest: HttpRequest { get }
+    var logger: ClientLogger { get }
+
+    init(copy: PipelineRequest)
+}
+
+public final class RealPipelineRequest: PipelineRequest {
```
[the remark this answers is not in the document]
bsiegel: Yep, it's a sucky name. I figured this would be in flux, will definitely take a pass on improving naming before I move this PR out of draft.

sdk/core/AzureCore/Source/Pipeline/PipelineResponse.swift
```diff
@@ -26,7 +26,15 @@
...
-public final class PipelineResponse: PipelineContextProtocol, Copyable {
+public protocol PipelineResponse: PipelineContextProtocol {
+    var httpRequest: HttpRequest { get }
+    var httpResponse: HttpResponse? { get }
+    var logger: ClientLogger { get }
+
+    init(copy: PipelineResponse)
+}
+
+public final class RealPipelineResponse: PipelineResponse {
```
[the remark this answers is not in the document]
bsiegel: 👍

sdk/core/AzureCore/Source/Pipeline/Policies/AddDatePolicy.swift
```diff
@@ -0,0 +1,46 @@
...
+public class AddDatePolicy: PipelineStageProtocol {
+    public var next: PipelineStageProtocol?
+
+    private static var httpDateTimeFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.timeZone = TimeZone(secondsFromGMT: 0)
+        formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss 'GMT'"
+        return formatter
+    }()
```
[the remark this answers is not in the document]
bsiegel: I also noticed that here locally :p It's corrected, but I hadn't seen that - I do plan/want to refactor the DateUtil class, maybe as part of this PR.

sdk/core/AzureCore/Tests/UserAgentPolicyTests.swift
```diff
@@ -29,29 +29,276 @@ import XCTest
 
 class UserAgentPolicyTests: XCTestCase {
 
-    /// Test that the default user agent policy is properly applied.
-    func testDefaultUserAgentPolicy() {
-        // This is an example of a functional test case.
-        // Use XCTAssert and related functions to verify your tests produce the correct results.
+    /// Test that the user agent policy creates the correct user agent with the empty initializer
+    func test_UserAgentPolicy_WithEmptyInitializer() {
```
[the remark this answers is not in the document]
bsiegel: Agreed, I will make sure to do so along with the rest of the guideline updates around the core policies.

sdk/core/AzureCore/Source/Providers/BundleInfoProvider.swift
bsiegel: Not sure exactly what is meant here, let's discuss in person.

sdk/core/AzureCore/Tests/AddDatePolicyTests.swift
bsiegel: Agreed.

sdk/core/AzureCore/Tests/UserAgentPolicyTests.swift
bsiegel: Yeah, so that's how it was previously, I just wrote a test case to cover it. Basically, to do the use case you suggest, first the SDK client would construct its initial pipeline with the useragent policy:

```swift
policies: [
    ...
    UserAgentPolicy(sdkName: "TheSDK",  sdkVersion: "1.0", ...)
    ...
]
```

Later, the consumer / extender / subclasser would insert their own user agent policy ahead of the default one:

```swift
client.insert(policy: UserAgentPolicy(userAgent: "MyServiceAgent"), at: .front)
```

Because the custom user agent policy is executed first, it will set the user agent to the explicitly provided string. Then, when the client's default user agent policy executes, it will see that there's an existing user agent and it will prepend the value with the conventional user agent string (see `test_UserAgentPolicy_WithCurrentNonDefaultUserAgent_PrependsHeaderValue`)

sdk/core/AzureCore/Source/Pipeline/PipelineRequest.swift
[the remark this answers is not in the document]
tjprescott: Total syncs with what I've read regarding testability on iOS. I just hope we can avoid naming gymnastics like "RealPipelineRequest"

sdk/core/AzureCore/Tests/UserAgentPolicyTests.swift
tjprescott: Sweet. Sounds reasonable.

sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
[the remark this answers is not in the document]
bsiegel: Resolved, please take a look at the new implementation - ended up rebasing things around a bunch, but the relevant changes are in BundleInfoProvider, DeviceProviders, and UserAgentPolicy

sdk/core/AzureCore/Source/Pipeline/PipelineRequest.swift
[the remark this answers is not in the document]
bsiegel: Removed this commit entirely, it was not needed

sdk/core/AzureCore/Source/Providers/BundleInfoProvider.swift
bsiegel: I figured out what you meant and implemented it, the relevant changes are in BundleInfoProvider and DeviceProviders

sdk/core/AzureCore/Tests/UserAgentPolicyTests.swift
bsiegel: Personally I could be convinced to remove the initializer that just blanket sets the user agent. In fact I think I'm going to do that. It offers no advantage over adding a HeadersPolicy that just sets the user agent:

```swift
let headers = HTTPHeaders()
headers[.userAgent] = "MyServiceAgent"
client.insert(policy: HeadersPolicy(addingHeaders: headers), at: .front)
```

bsiegel: Removed this initializer in 353a31e and fixed the tests in 94739a1

sdk/core/AzureCore/Tests/UtilTests.swift
```diff
@@ -69,7 +69,7 @@ class RegexUtilTests: XCTestCase {
     }
 
     func testStringUtil() {
-        let dateString = "Jan, 02 2020 07:12:34 GMT"
+        let dateString = "Thu, 02 Jan 2020 07:12:34 GMT"
```
tjprescott: This format was carried over from (presumably) the Python SDK. If what it called RFC1123 isn't actually RFC1123, that's fine, but if we have changed the format associated with these methods, they will no longer work with the Storage SDK formats that code was written for. I suspect it would return nil.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -109,24 +109,15 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
                 baseUrl: baseUrl,
                 transport: URLSessionTransport(),
                 policies: [
-                    // Python: QueueMessagePolicy(),
-                    HeadersPolicy(),
-                    // Python: config.proxy_policy,
-                    UserAgentPolicy(),
-                    // Python: StorageContentValidation(),
-                    // Python: StorageRequestHook(**kwargs),
+                    UserAgentPolicy(forLibraryWithBundleIdentifier: "com.azure.AzureStorageBlob"),
```
tjprescott: Is there a way to programmatically obtain the bundle identifier or access it with a compiler directive? I ask due to the recent PR where the bundle identifier for the demo app was changed from `.demo` to `.examples` and it broke the demo app. Same concern as when we had the static embedded version.

sdk/core/AzureCore/Source/ClientLogger.swift
```diff
@@ -44,7 +44,7 @@ public protocol ClientLogger {
     func warning(_: @autoclosure @escaping () -> String?)
     func error(_: @autoclosure @escaping () -> String?)
 
-    func log(_: @escaping () -> String?, atLevel: ClientLogLevel)
+    func log(_: () -> String?, atLevel: ClientLogLevel)
```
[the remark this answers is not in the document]
bsiegel: It was unneeded. We don't store the closure so its scope never escapes the log method, thus @enclosing was unneeded. I think I had added it initially because it was in Colby's logger.

sdk/core/AzureCore/Tests/LoggingPolicyTests.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+// --------------------------------------------------------------------------
+
+import XCTest
+@testable import AzureCore
+
+class LoggingPolicyTests: XCTestCase {
+
+    /// Test that the logging policy
+    func test_LoggingPolicy_() {
```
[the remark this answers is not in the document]
bsiegel: Oh, nope - I'm still working on the logging tests - I will have a commit pushed here shortly. That was just a placeholder.

sdk/core/AzureCore/Tests/PipelineTests.swift
```diff
@@ -35,7 +35,7 @@ class PipelineTests: XCTestCase {
             baseUrl: baseUrl,
             transport: URLSessionTransport(),
             policies: [
-                UserAgentPolicy(),
+                UserAgentPolicy(sdkName: "Test", sdkVersion: "1.0"),
```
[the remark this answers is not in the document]
bsiegel: There is a convenience initializer that requires only one parameter:

```swift
UserAgentPolicy(forLibraryWithBundleIdentifier: "com.azure.storage.AzureStorageBlob")
```

(It also accepts an optional `applicationId`)

sdk/core/AzureCore/Tests/UtilTests.swift
bsiegel: I noticed the tests were failing because that is not actually an RFC1123 string. The change shows a valid RFC1123 string. The original format is very odd because why would there be a comma after the short month name? So is Storage just using a very stupid custom date format based on someone misunderstanding RFC1123?

Going to research what the other SDKs do. Also if this ***is*** a custom format for storage, we should move these methods out of Core and into the Storage framework.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I would expect that the bundle ID should basically never change once we have our first release, so I don't see that as much of a problem. That said, we can get the main app bundle, or we can get a bundle by its bundle ID. There is no way to get the "current framework" bundle. The best we can do is call `Bundle(for: AnyClass)` and pass a class that was defined in the bundle. That incurs a reflection penalty though, whereas looking it up by bundle ID doesn't.

If we plumbed that through, I imagine it might look like:

```swift
// If you want to explicitly name the class
UserAgentPolicy(for: AppConfigurationClient.self)

// If you want to look it up from self
UserAgentPolicy(for: type(of: self))
```

I don't think there's a way to avoid providing that parameter at the call site. I don't know if I'm necessarily convinced this is better, what do you think?

bsiegel: I think I'm convinced that the first form is an improvement over specifying the bundle. I added some fixups to make the change.

sdk/core/AzureCore/Tests/UtilTests.swift
bsiegel: My change also matches the format specified in `StorageBlobClient.StorageJSONEncoder`.

bsiegel: JS seems like it expects the standard RFC1123 format: https://github.com/Azure/azure-sdk-for-js/blob/96de036571c46659938ace951a8eb78834b8dbc4/sdk/core/core-http/test/serializationTests.ts#L283

tjprescott: Ah, okay, then the problem must have been me! Thanks for checking!

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Implemented in f71eea8, you can see how this looks in 0fc5d5f

sdk/core/AzureCore/Tests/UtilTests.swift
bsiegel: I changed around how we implemented the various conversions that live in the Utils folder in 8be84e4. The extension properties felt weirdly non-swifty, so I replaced them with converting initializers. It feels a lot better to me now but very interested to hear your feedback.

tjprescott:
@bsiegel looks great. I'll let you merge when you're ready.

PR #189, opened 2020-01-20
Sets versions for packages (discuss?).

mitchdenny:
This PR sets the versions of the packages to ```1.0.0-beta.1``` across the board. We should discuss what the versions should actually be, but I needed to make sure the packages all had versions so that I could proceed with some scripting work.

### DIFF
--- a/examples/AzureSDKDemoSwift/AzureSDKDemoSwift.xcodeproj/project.pbxproj
+++ b/examples/AzureSDKDemoSwift/AzureSDKDemoSwift.xcodeproj/project.pbxproj
@@ -521,7 +521,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.azure.demo.AzureSDKDemoSwifty;
+				PRODUCT_BUNDLE_IDENTIFIER = com.azure.examples.AzureSDKDemoSwifty;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
@@ -542,7 +542,7 @@
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = com.azure.demo.AzureSDKDemoSwifty;
+				PRODUCT_BUNDLE_IDENTIFIER = com.azure.examples.AzureSDKDemoSwifty;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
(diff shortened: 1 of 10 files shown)


tjprescott:
Is "1.0.0-beta.X" what the other track 2 SDKs did, as opposed to something like "0.1.0"?

bsiegel:
This is good, we should also update / correct / standardize the bundle IDs. I'm not sure what the convention should be here. Each framework already has the word "Azure" and the service name (e.g. "AzureStorageBlob"). Should the bundle ID just be `com.azure.<FrameworkName>`? Or should we be putting the service name into the dotted part of the path too (e.g. `com.azure.storage.AzureStorageBlob`)?

bsiegel:
> Is "1.0.0-beta.X" what the other track 2 SDKs did, as opposed to something like "0.1.0"?

Generally yes, though most of them had pre-existing track 1 libraries so they generally bumped to the next major version for each service. For example, in .Net, I believe storage was 12.0.0-preview.1 and keyvault was 4.0.0-preview.1

I think we went with "preview" there because the ecosystem expects that terminology. Is that why you went with "beta" here @mitchdenny ?

mitchdenny:
> This is good, we should also update / correct / standardize the bundle IDs. I'm not sure what the convention should be here. Each framework already has the word "Azure" and the service name (e.g. "AzureStorageBlob"). Should the bundle ID just be `com.azure.<FrameworkName>`? Or should we be putting the service name into the dotted part of the path too (e.g. `com.azure.storage.AzureStorageBlob`)?

I would probably do ```com.azure.service.Library```. I think it gives you the room to break things into multiple libraries.

mitchdenny:
Regarding the version suffix. When we originally went out we tried a few things, but landed on ```-preview.1```. However we found in the Java ecosystem in particular that there were some assumptions around semantic sorting where they expect preview packages to have the ```-beta``` suffix. Which is why we swapped to ```-beta.1``` as a general approach. Python violates that though since they append the suffix ```b1``` to their version numbers.

So I think ```1.0.0-beta.1``` is the right way to go. There are no guarantees around preview versions.

bsiegel:
> I would probably do `com.azure.service.Library`. I think it gives you the room to break things into multiple libraries.

I think I tend to agree, would you want to add that cleanup to this PR or should we open a separate PR?

mitchdenny:
> I think I tend to agree, would you want to add that cleanup to this PR or should we open a separate PR?

I've updated the bundle identifiers to what they should be based on current file locations. One thing that I wanted to check on is the location of the AzureCSComputerVision library. Given we'll probably be breaking up the various cog-services related libraries into seperate ```sdk/[service]``` folders, should we be preemtively be putting this one in ```sdk/computervision``` and renaming the bundle to ```com.azure.computervision.AzureCSComputerVision```?

It doesn't have to be resolved now but I wanted to flag it for consideration.

bsiegel:
> Given we'll probably be breaking up the various cog-services related libraries into seperate `sdk/[service]` folders, should we be preemtively be putting this one in `sdk/computervision` and renaming the bundle to `com.azure.computervision.AzureCSComputerVision`?

Yeah, I believe it is this way for historical reasons but I'm not aware of a compelling reason to keep it this way. That said let's address that in a separate PR :)

mitchdenny:
> Yeah, I believe it is this way for historical reasons but I'm not aware of a compelling reason to keep it this way. That said let's address that in a separate PR :)

Created an issue to track it. Merging this PR now.

PR #195, opened 2020-01-27
Fix memory leaks

### LINKED ISSUE #172: XMLTreeNode has memory leak
Running the Leaks instrument on the Swift demo app revealed memory leaks in the XMLTreeNode data structure.

### LINKED ISSUE #193: ContentDecodeXMLParser should use weak delegate
Possibly related to #172. Update and remove FIXME in code.

tjprescott:
Fixes #193. Fixes #172. Also cleans up some warnings (which I put in a separate commit in an attempt to be a better person!).

### DIFF
--- a/sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
+++ b/sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
@@ -55,7 +55,7 @@ public class AppConfigurationClient: PipelineClient {
                     HeadersPolicy(),
                     UserAgentPolicy(),
                     authPolicy,
-                    ContentDecodePolicy(logger: options?.logger),
+                    ContentDecodePolicy(),
                     LoggingPolicy()
                 ],
                 withOptions: clientOptions)
--- a/sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
+++ b/sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
@@ -55,7 +55,7 @@ public class AppConfigurationClient: PipelineClient {
                     HeadersPolicy(),
                     UserAgentPolicy(),
                     authPolicy,
-                    ContentDecodePolicy(),
+                    ContentDecodePolicy(logger: options?.logger),
                     LoggingPolicy()
                 ],
                 withOptions: clientOptions)
--- a/sdk/core/AzureCore/Source/DataStructures/XMLModel.swift
+++ b/sdk/core/AzureCore/Source/DataStructures/XMLModel.swift
@@ -198,6 +198,7 @@ internal class XMLTreeNode {
     var dictionary: [String: Any]? {
         var propDict = [String: Any]()
         for (key, metadata) in properties {
+            assert(metadata.parent == nil, "Unexpectedly found a parent reference, which could leak to memory leaks.")
             switch metadata.type {
             case .property:
                 propDict[key] = metadata.value
--- a/sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
@@ -45,13 +45,6 @@ internal class ContentDecodeXMLParser: NSObject, XMLParserDelegate {
     internal var inferStructure = false
     internal var logger: ClientLogger?
 
-    // MARK: Initializers
-
-    public init(logger: ClientLogger?) {
-        super.init()
-        self.logger = logger
-    }
-
     // MARK: Methods
 
     public func parserDidStartDocument(_: XMLParser) {
@@ -178,8 +171,8 @@ public class ContentDecodePolicy: PipelineStage {
 
     // MARK: Initializers
 
-    public init(logger: ClientLogger?) {
-        self.xmlParser = ContentDecodeXMLParser(logger: logger)
+    public init() {
+        self.xmlParser = ContentDecodeXMLParser()
     }
 
     // MARK: Public Methods
@@ -194,6 +187,7 @@ public class ContentDecodePolicy: PipelineStage {
 
         // Store the logger so that the XML parser delegate functions can access it
         logger = response.logger
+        xmlParser?.logger = logger
 
         var contentType = (httpResponse.headers["Content-Type"]?.components(separatedBy: ";").first) ??
             "application/json"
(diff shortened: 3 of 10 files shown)


sdk/core/AzureCore/Source/DataStructures/XMLModel.swift
```diff
@@ -198,6 +198,7 @@ internal class XMLTreeNode {
     var dictionary: [String: Any]? {
         var propDict = [String: Any]()
         for (key, metadata) in properties {
+            assert(metadata.parent == nil, "Unexpectedly found a parent reference, which could leak to memory leaks.")
```
tjprescott: The Swift compiler will strip this out for release builds, but I consider it a good safeguard against regression, especially given that only running the demo in profiling mode caught this leak.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -102,17 +102,19 @@ internal class ContentDecodeXMLParser: NSObject, XMLParserDelegate {
             for (key, value) in attributeDict {
                 let defaultKey = "_\(key)"
                 let jsonKey = xmlMap?[defaultKey]?.jsonName ?? defaultKey
-                let attr = XMLTreeNode(name: jsonKey, type: .property, parent: newNode, value: value)
+                let attr = XMLTreeNode(name: jsonKey, type: .property, parent: nil, value: value)
```
tjprescott: This was the source of the XMLTree memory leak.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -169,15 +171,15 @@ public class ContentDecodePolicy: PipelineStage {
     // MARK: Properties
 
     public var next: PipelineStage?
-    public let jsonRegex = NSRegularExpression("^(application|text)/([0-9a-z+.]+)?json$")
     public var logger: ClientLogger?
 
-    internal var delegate: ContentDecodeXMLParser?
+    internal let jsonRegex = NSRegularExpression("^(application|text)/([0-9a-z+.]+)?json$")
+    internal var xmlParser: ContentDecodeXMLParser?
```
tjprescott: Swiftlint complained that, as a delegate, this should be a weak variable. However, for this to be a true delegate pattern, ContentDecodeXMLParser should be a protocol, not a class. Realizing that a protocol was not only unwanted but unnecessary, I changed this to not be called a delegate, as there's no opportunity here for a memory leak.

sdk/appconfiguration/AzureAppConfiguration/Source/AppConfigurationClient.swift
```diff
@@ -55,7 +55,7 @@ public class AppConfigurationClient: PipelineClient {
                     HeadersPolicy(),
                     UserAgentPolicy(),
                     authPolicy,
-                    ContentDecodePolicy(),
+                    ContentDecodePolicy(logger: options?.logger),
```
bsiegel: I wish we could avoid passing a logger here. All the other policies get it from the request context. Maybe if we delay initializing the xmlParser and only do it lazily in onRequest, that would solve the problem. I don't see a particular advantage to initializing it inside ContentDecodePolicy's init method.

tjprescott: I agree and will take a second look.

tjprescott: Okay, should be fixed now.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -194,6 +187,7 @@ public class ContentDecodePolicy: PipelineStage {
 
         // Store the logger so that the XML parser delegate functions can access it
         logger = response.logger
```
bsiegel: The logger is a per-request object, so we should probably avoid storing it in a property. I would probably remove the property and then change this to:
```suggestion
        let logger = response.logger
```

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -194,6 +187,7 @@ public class ContentDecodePolicy: PipelineStage {
...
+        xmlParser?.logger = logger
```
bsiegel: Also, since not every request will need an xmlParser, maybe it would be good to make the property a `lazy` property, and then only assign the logger from within the `parse` method (by which time we know we need to have an xmlParser available).

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
bsiegel: Is it worth adding a code comment about the fact that parent is set to nil explicitly to avoid a memory leak (and maybe explain why not doing so causes the leak)?

tjprescott: TL/DR: I don't think a comment would help here--the assert is the best defense.

There are many places where not handling the parent reference correctly could introduce a memory leak, not just here, which is why I put the assert statement in XMLModel. The test cases and the demo application will fail if a regression is introduced and from that point it should be easy to diagnose and correct the problem.

In fact, that's exactly what happened. As soon as I introduced the assert I saw that the underscored properties (which come from XML attributes) were holding onto a parent reference and creating a memory leak.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
tjprescott: Done!

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
bsiegel: Excellent explanation, thanks! And now it's at least archived on this PR for the future :)

PR #197, opened 2020-01-31
Update demo app.

tjprescott:
Modifies the demo app, per discussions:
- immediately moves to the video container (skips selecting a container)
- adds pull-to-refresh capability for the video list
- adds upload button, which opens an image picker

Also, adds a `.swiftlint.yml` config file to temporarily disable certain Swiftlint rules. Addressing these violations is tracked by https://github.com/Azure/azure-sdk-for-ios/issues/198.

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPRequest.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPRequest.swift
@@ -48,7 +48,10 @@ public class HTTPRequest: DataStringConvertible {
 
     public init(method: HTTPMethod, url: String, headers: HTTPHeaders, files: [String]? = nil, data: Data? = nil) {
         self.httpMethod = method
-        self.url = url
+        guard let encodedUrl = url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
+            fatalError("Cannot encode URL.")
+        }
+        self.url = encodedUrl
         self.headers = headers
         self.files = files
         self.data = data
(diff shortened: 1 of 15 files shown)


sdk/core/AzureCore/Source/Pipeline/Transport/HTTPRequest.swift
```diff
@@ -48,7 +48,10 @@ public class HTTPRequest: DataStringConvertible {
 
     public init(method: HTTPMethod, url: String, headers: HTTPHeaders, files: [String]? = nil, data: Data? = nil) {
         self.httpMethod = method
-        self.url = url
+        guard let encodedUrl = url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
```
bsiegel: I worry about doing the encoding here for a couple reasons. First, if I supply a URL to the initializer and then read the `url` property, they will be different. Also, if I set a new URL via that property, it won't be encoded.

Can we delay any sort of normalization like this until just before the request is actually dispatched (presumably in URLSessionTransport)?

In my experience, other workarounds like adding a setter on the `url` property that encodes the URL are brittle, because it's too easy to accidentally double-encode something.

This should also allow library authors to avoid having to fiddle with the encoding like you were doing below in StorageStreamDownloader.

tjprescott: Makes sense. I'll take a look. This was needed because otherwise URLs with space in them failed.

tjprescott: Fixed. The StorageStreamDownloader did really need URL encoding--it just needed to get the last part of a blob name that might have a "virtual folder" portion of the name (ex:  blob named "virtualdir/file.txt")

Issue #198, opened 2020-01-31
Address Swiftlint disabled rules


tjprescott:
- [ ] `line_length`: automatic fix?
- [ ] `identifier_name`: inline disables (`id` is fine as an identifier)
- [ ] `todo`: Resolve TODOs
- [ ] `function_body_length`: Break up functions
- [ ] `file_length`: Break up files

PR #203, opened 2020-02-07
Fix swiftlint issues and apply swiftformat

bsiegel:
Trying to stay as close to the defaults for these two tools as possible. 73e98d5 is just auto reformatting done by `swiftformat`. e014274 is my best attempt to fix or silence most `swiftlint` issues. I left the warnings that appear for TODO/FIXME comments, I think those are valuable to see so that we remember to fix them ASAP.


.swiftlint.yml
```diff
@@ -1,6 +1,2 @@
 disabled_rules: # rule identifiers to exclude from running
-  - line_length
-  - identifier_name
-  - todo
-  - function_body_length
-  - file_length
+  - large_tuple
```
tjprescott: Please update issue #198 that is opened to track these.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -24,6 +24,10 @@
 //
 // --------------------------------------------------------------------------
 
+// swiftlint:disable file_length
+// swiftlint:disable function_body_length
```
tjprescott: I would prefer to keep these in issue #198 as I feel they are legitimate and we could break these into smaller chunks. Ditto for elsewhere the file-level disable applies.

sdk/storage/AzureStorageBlob/AzureStorageBlob.xcodeproj/project.pbxproj
```diff
@@ -11,7 +11,6 @@
 		0A686493233E834D004B7E17 /* AzureStorageBlob.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 0A686489233E834D004B7E17 /* AzureStorageBlob.framework */; };
```
tjprescott: Are we running swiftformat automatically as part of the build step?  If not, we should. It could easily go into the `swiftlint.sh` script.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
bsiegel: Agree, I would have preferred to actually fix the issues in this PR. However it gets tricky because a lot of the function_length violations are when we set up a call, then make the request, and then have a callback for the completion. I went through a couple options but they would have made semantic changes rather than just cosmetic ones, and I didn't want to introduce semantic changes in this PR.

tjprescott: I agree, that's why I'm just saying don't mark them complete in the tracking issue :)

.swiftlint.yml
bsiegel: Done!

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
bsiegel: Ahh okay, I understand what you mean. I definitely want to keep all of the ones that have been disabled in-code in the tracking issue until we determine that we feel strongly about keeping a given disable rather than making the fix.

sdk/storage/AzureStorageBlob/AzureStorageBlob.xcodeproj/project.pbxproj
bsiegel: Not yet, but we definitely can. Right now there are three call sites where swiftformat replaces the parameter name with `_` because it's not used, but that's either because we have commented out the body of the function or (in one case) I think we have a legitimate bug.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -335,9 +333,7 @@ public class BlobStreamDownloader {
     /// Download the contents of this file to a stream.
     /// - Parameters:
     ///   - group: An optional `DispatchGroup` to wait for the download to complete.
-    ///   - completion: A completion handler with which to process the downloaded chunk.
-    public func next(inGroup group: DispatchGroup? = nil, then completion: HTTPResultHandler<Data>) {
```
bsiegel: @tjprescott Does this look like the correct fix for this? Swiftformat flagged that we weren't using the completion parameter. It seems like if we want to know about the progress / completion of a single chunk, then we should supply a callback in the `self.options.progressCallback` property, rather than passing in the callback as a parameter to the `next` method.

sdk/storage/AzureStorageBlob/AzureStorageBlob.xcodeproj/project.pbxproj
tjprescott: Well, swiftlint autocorrect does that too and while it's not wrong it can be annoying. Even so, it wouldn't be hard to replace the _ with something when (or if) we uncomment that code.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
tjprescott: It is correct that it wasn't being called. In my branch for Monday's demo, rather than passing callbacks around, I created a `BlobDownloaderDelegate` protocol for handling this. It is much cleaner that way.

bsiegel: I was thinking the same thing after seeing how large the callbacks were :)

PR #204, opened 2020-02-10
Demo app polishing.

tjprescott:
Updates to the Swift demo app. Adds activity spinner and customizable pre-play buffering. Adds BlobDownloaderDelegate, as a better alternative to callback closures.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/BlobTableViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobTableViewController.swift
@@ -160,7 +160,7 @@ class BlobTableViewController: UITableViewController, MSALInteractiveDelegate {
                 showAlert(error: "Unable to create URL!")
                 return
             }
-            self.showActivitySpinner()
+            showActivitySpinner()
             try blobClient.download(url: url, withOptions: options) { result, _ in
                 switch result {
                 case let .success(downloader):
@@ -248,7 +248,7 @@ extension BlobTableViewController: BlobDownloadDelegate {
         guard downloader === current else { return }
         print("Download: \(downloader.blobName) Progress: %\(progress.percentProgress)")
         // blobClient.options.logger.info("Progress: %\(percentDone)")
-        if progress.percentProgress > 25 && !downloaderBuffered {
+        if progress.percentProgress > 25, !downloaderBuffered {
             downloaderBuffered = true
             hideActivitySpinner()
             DispatchQueue.main.async {
@@ -263,7 +263,7 @@ extension BlobTableViewController: BlobDownloadDelegate {
         }
     }
 
-    func downloader(_ downloader: BlobStreamDownloader, didFinishWithProgress progress: BlobDownloadProgress) {
+    func downloader(_ downloader: BlobStreamDownloader, didFinishWithProgress _: BlobDownloadProgress) {
         guard let current = currentDownloader else { return }
         guard downloader === current else { return }
         // update playing with completed file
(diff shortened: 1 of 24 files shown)


examples/AzureSDKDemoSwift/Source/BlobTableViewController.swift
```diff
@@ -160,7 +160,7 @@ class BlobTableViewController: UITableViewController, MSALInteractiveDelegate {
                 showAlert(error: "Unable to create URL!")
                 return
             }
-            self.showActivitySpinner()
+            showActivitySpinner()
```
bsiegel: It seems like swiftformat is not picking up the config file, these changes shouldn't be showing up. If you just run `swiftformat sdk` from the root of the repo, does it revert these changes?

bsiegel: You can also explicitly pass the config file path with `--config <repo_root>/.swiftformat` but that shouldn't be necessary.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -318,22 +329,41 @@ public class BlobStreamDownloader {
...
+            }
         }
     }
 
     /// Download the contents of this file to a stream.
     /// - Parameters:
     ///   - group: An optional `DispatchGroup` to wait for the download to complete.
-    public func next(inGroup group: DispatchGroup? = nil) {
+    ///   - completion: A completion handler with which to process the downloaded chunk.
+    public func next(inGroup group: DispatchGroup? = nil, then completion: @escaping (Result<Data, Error>, HTTPResponse) -> Void) {
```
bsiegel: The type for `completion` should be `HTTPResultHandler<Data>` (the type alias just makes things simpler).

examples/AzureSDKDemoSwift/Source/BlobTableViewController.swift
tjprescott: I think swiftformat is working, because I ran a build and it made a bunch of changes I had to commit. This was likely swiftformat undoing parts that may have been reintroduced by a rebase.

sdk/appconfiguration/AzureAppConfiguration/Source/Models/ConfigurationSetting.swift
```diff
@@ -54,7 +54,7 @@ public struct ConfigurationSetting: Codable {
         self.label = label
         self.tags = tags
         self.contentType = contentType
-        self.etag = eTag
+        etag = eTag
```
bsiegel: Here's an example where swiftlint should not have made this change. Very odd. I will check out your branch and try running it locally.

bsiegel: Just tried this and swiftformat undid a bunch of changes including this one.

tjprescott: Hmm...yeah, I see what you mean. I just pushed that simple change you recommended about the alias and locally ran `swiftformat`. It made a bunch of changes. So it seems that its inclusion in the `swiftlint.sh` script isn't always working as intended.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
tjprescott: Fixed.

PR #205, opened 2020-02-11
Specify path of config files in swiftformat

bsiegel:
This fixes an issue where running the script from Xcode could result in the tools not being able to locate their config files correctly and would apply the default ruleset instead.


bsiegel:
To test:

On `dev`, perform a build out of Xcode of AzureCore. Notice that swiftformat has reformatted many files. Reset the working directory, check out this PR, and do the same build. No files should be modified this time.

PR #228, opened 2020-03-10
Transfer Manager - Blob Download

tjprescott:
Implements BlobDownload using Transfer Manager.
Add the "AzureSDKTransferManager" demo app.

Notes:
- MultiBlobTransfer is only "sketched out" and is not intended to be fully functioning
- There are about 20 or so occurrence of `fatalError` in the code, which will have to be resolved. Each will require individual consideration of how to handle in production code, so they will be addressed in a future PR (after TM upload is complete)


sdk/core/AzureCore/Source/DataStructures/ResumableOperationQueue.swift
```diff
@@ -0,0 +1,148 @@
...
+    public func pause(_ operation: ResumableOperation) {
+        operation.pause()
+        removeFromDataStore(operation)
+    }
+
+    public func remove(_ operation: ResumableOperation) {
+        removeFromDataStore(operation)
+
+        // Cancel the operation.
+        if !(operation as Operation).isFinished {
```
bsiegel: Odd that this cast is needed here since `ResumableOperation` is a subclass of `Operation`.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -73,13 +77,11 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
-     - Parameter accountUrl: Base URL for the storage account.
-     - Parameter credential: A credential object used to retrieve authentication tokens.
-     - Parameter withOptions: A `StorageBlobClientOptions` object to control the download.
-     - Returns: A `StorageBlobClient` object.
-     */
+    /// Create a Storage blob data client.
+    /// - Parameters:
+    ///   - accountUrl: Base URL for the storage account.
+    ///   - credential: A credential object used to retrieve authentication tokens.
+    ///   - options: A `StorageBlobClientOptions` object to control the download.
```
bsiegel: For docstrings, shouldn't we be naming the parameters by their external name (here, "withOptions")? Not sure what the guidance is on that, I can see arguments either way.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -344,41 +339,111 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
-     - Parameter options: A `DownloadBlobOptions` object to control the download operation.
-     - Parameter completion: An `HTTPResultHandler` closure that returns a `BlobStreamDownloader` object on success.
-     */
-    public func download(
+    /// Download a blob from a given URL.
+    /// - Parameters:
+    ///   - url: A URL to a blob to download.
+    ///   - options: A `DownloadBlobOptions` object to control the download operation.
+    ///   - completion: An `HTTPResultHandler` closure that returns a `BlobStreamDownloader` object on success.
+    public func rawDownload(
```
bsiegel: We'll want to think about the names of these method pairs, and also make sure we make the same naming choices in Android.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -344,41 +339,111 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
+    ///   - container: The name of the container.
     public func url(forBlob blob: String, inContainer container: String) -> URL? {
         var url = URL(string: baseUrl)
         url?.appendPathComponent(container)
         url?.appendPathComponent(blob)
         return url
     }
+
+    public func localUrl(
+        blob: String,
```
bsiegel: Should these have the same names as the parameters to `url()`? E.g. "forBlob", "inContainer"

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -344,41 +339,111 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
...
+            client: self,
+            delegate: nil,
+            name: blob,
+            container: container,
+            options: options
+        )
+        return downloader.downloadDestination
+    }
+
+    public func localUrl(remoteUrl url: URL, withOptions options: DownloadBlobOptions? = nil) throws -> URL? {
```
bsiegel: Similarly consider the parameter names here.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+    public typealias TransferManagerType = URLSessionTransferManager
+
+    // MARK: Properties
+
+    public weak var delegate: TransferManagerDelegate?
+
+    public var logger: ClientLogger
+
+    internal lazy var session: URLSession = {
+        let config = URLSessionConfiguration.background(withIdentifier: "com.azuresdk.transfermanager")
```
bsiegel: If this doesn't have to be a triple, we should probably use one that matches or includes the bundle identifier.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    public var persistentContainer: NSPersistentContainer? = {
+        guard let bundle = Bundle(identifier: "com.azure.storage.AzureStorageBlob") else { return nil }
+        guard let url = bundle.url(forResource: "AzureStorage", withExtension: "momd") else { return nil }
+        guard let model = NSManagedObjectModel(contentsOf: url) else { return nil }
+        let container = NSPersistentContainer(name: "AzureSDKTransferManager", managedObjectModel: model)
+        container.loadPersistentStores(completionHandler: { _, error in
+            if let error = error as NSError? {
+                // Replace this implementation with code to handle the error appropriately.
```
bsiegel: What should we actually do here? Is this a TODO? if so we should mark with a //TODO. It's written like they're instructions for the end user.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+            transfer.totalBlocks = Int64(transfer.transfers.count)
+        }
+        queueOperationsFor(blobTransfer: transfer)
+    }
+
+    // MARK: Cancel Operations
+
+    public func cancel(transfer: Transfer) {
+        transfer.state = .canceled
+        assert(transfer.operation != nil, "Transfer operation unexpectedly nil.")
```
bsiegel: Seems like maybe this is not a fatal situation, should we remove the assert? Or do those only occur in debug builds?

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+        let allowed: [TransferState] = [.pending, .inProgress]
+        guard allowed.contains(transfer.state) else { return }
+
+        switch transfer {
+        case let transfer as BlockTransfer:
+            pause(transfer: transfer)
+        case let transfer as BlobTransfer:
+            pause(transfer: transfer)
+        default:
+            fatalError("Unrecognized transfer type: \(transfer.self)")
```
bsiegel: Could this be a MultiBlobTransfer? And if so would it hit this fatalError case?

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+            operationQueue.add(BlockOperation(withTransfer: transfer))
+        case let transfer as BlobTransfer:
+            for blockTransfer in transfer.transfers {
+                if allowed.contains(blockTransfer.state) {
+                    blockTransfer.state = .pending
+                }
+            }
+            queueOperationsFor(blobTransfer: transfer)
+        default:
+            fatalError("Unrecognized transfer type: \(transfer.self)")
```
bsiegel: Likewise here.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+        }
+        operation(transfer.operation, didChangeState: transfer.state)
+    }
+
+    // MARK: Core Data Operations
+
+    public func loadContext() {
+        // Hydrate operationQueue from CoreData
+        guard let context = persistentContainer?.viewContext else {
+            fatalError("Unable to load persistent container.")
```
bsiegel: When would this condition fail? And is there something better to do than a fatalError?

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    public func saveContext() {
+        DispatchQueue.main.async { [weak self] in
+            guard let context = self?.persistentContainer?.viewContext else { return }
+            if context.hasChanges {
+                do {
+                    try context.save()
+                } catch {
+                    // Replace this implementation with code to handle the error appropriately.
```
bsiegel: Same as above - what should we actually do here? Is this a TODO? if so we should mark with a //TODO. It's written like they're instructions for the end user.

sdk/core/AzureCore/Source/DataStructures/ResumableOperationQueue.swift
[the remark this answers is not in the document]
tjprescott: Indeed. I removed the cast and it works fine.

sdk/core/AzureCore/Tests/ReachabilityManagerTests.swift
```diff
@@ -0,0 +1,253 @@
...
+        // Given, When
+        let manager = ReachabilityManager()
+
+        // Then
+        XCTAssertNotNil(manager)
+    }
+
+    func test_HostManager_IsReachableOnWiFi() {
+        // Given, When
+        let manager = ReachabilityManager(host: "localhost")
```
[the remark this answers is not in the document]
tjprescott: Honestly I didn't put much (translation: any) thought into this. The Reachability Manager files were all lifted directly from the old iOS code (which I think was lifted from Alamofire if the license header is any indication). I simply updated the test names to match our naming schema and verified that they all passed.

sdk/storage/AzureStorageBlob/Source/DataStructures/BlobOperation.swift
```diff
@@ -0,0 +1,115 @@
...
+                    group.leave()
+                case .failure:
+                    self.transfer?.state = .failed
+                    // TODO: The failure needs to propagate to the entire operation...
+                    self.delegate?.operation(self, didChangeState: .failed)
+                    group.leave()
+                }
+            }
+        } else {
+            let delay = UInt32.random(in: 5 ... 10)
```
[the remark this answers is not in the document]
tjprescott: Yep, this is the "dummy workload" I use in the AzureSDKTransferManager example. Will add a TODO.

sdk/storage/AzureStorageBlob/Source/DataStructures/MultiBlobOperation.swift
```diff
@@ -24,27 +24,35 @@
...
+    }
 
-    // MARK: Required Methods
+    // MARK: Public Methods
 
-    /// Converts `self` to its corresponding
-    /// `ObjectiveCType`.
-    func bridgeToObjectiveC() -> ObjectiveCType
+    public override func main() {
+        guard let transfer = self.transfer as? MultiBlobTransfer else { return }
```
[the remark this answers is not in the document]
tjprescott: Yes, totally missing logic. I was sketching out the way this class should look based on the other two, but the multi-blob scenario isn't intended to be complete. I added a TODO and comment to indicate as much

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamDownloader.swift
```diff
@@ -435,7 +450,8 @@ public class BlobStreamDownloader {
                     // if the requestedSize was not specfied, the block list will need to be recomputed
                     // based on the actual file size.
                     var recomputeBlockList = self.requestedSize == nil
-                    self.requestedSize = (self.requestedSize ?? fileSize) - (self.options.range?.offset ?? 0)
+                    self.requestedSize = (self.requestedSize ?? fileSize) - (self.options.range?.offset ?? 0
```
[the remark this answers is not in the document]
tjprescott: Nope, though I'm surprised neither of the format checkers fixed this. Just a stray newline :) Fixed

sdk/storage/AzureStorageBlob/Source/DataStructures/TransferManager.swift
```diff
@@ -0,0 +1,103 @@
...
+    // MARK: Properties
+
+    var reachability: ReachabilityManager? { get }
+    var persistentContainer: NSPersistentContainer? { get }
+    static var shared: Self { get }
+    var logger: ClientLogger { get set }
+
+    // MARK: Storage Methods
+
+    // func upload(_ url: URL) -> Transferable
```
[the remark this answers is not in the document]
tjprescott: Added a TODO and comment. Essentially, once the signatures that exist in the TM implementation are reviewed and approved, we can uplevel them into the protocol definition. They definitely won't be these signatures though, since "Transferable" is no longer a thing.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: I would agree with you... but when you use the "Add Document" functionality in XCode, it exposes the internal names so I went with that.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Agreed. This was just a temporary rename because I didn't want to throw away a working implementation.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: OR the others should match this scheme. :)

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: This is boilerplate XCode gives you when you generate a CoreData project. Realistically, we will need to go through and determine the best way to resolve all instances of "fatalError" in the code (which there are currently many)

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: Yes, the assert will only fail in debug.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: Yes, there are two issues here:
1) MultiBlobTransfer is not fully implemented (and I don't plan to for the UX study)
2) it shouldn't fatalError regardless.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: Yep, same.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: This would fail if the persistent container were nil, because the bundle identifier or something else was wrong. Arguably it could proceed as if there was no saved CoreData

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: I removed both of these boilerplate comments. I will address removing all of the 20 or so fatalError occurrences once I have upload working with TM since there's no good reason to have fatalError in production code.

sdk/core/AzureCore/Tests/ReachabilityManagerTests.swift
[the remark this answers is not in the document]
bsiegel: Got it, I see that we've kept the Alamofire copyright in the headers for this file and ReachabilityManager. However the ReachabilityManagerType.swift source file has our copyright instead - should we fix that?

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Either way, we should use a preposition before the noun e.g. "forBlob"

sdk/core/AzureCore/Tests/ReachabilityManagerTests.swift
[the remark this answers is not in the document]
tjprescott: Done.

PR #230, opened 2020-03-16
Blob Upload (Transfer Manager)

tjprescott:
Adds:
- `upload` using Transfer Manager.
- `rawUpload` using the raw client
- updates the `AzureSDKDemoSwift` app with Transfer Manager-enabled upload capability.


sdk/core/AzureCore/Source/DataStructures/XMLMap.swift
```diff
@@ -0,0 +1,180 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+// MARK: XML Metadata
+
+public enum ElementToJsonStrategy {
```
tjprescott: No changes to this file. It was simply split out from a very large file.

sdk/core/AzureCore/Source/DataStructures/XMLModel.swift
```diff
@@ -34,231 +34,12 @@ public protocol XMLModel {
     // MARK: Required Methods
 
     static func xmlMap() -> XMLMap
-}
-
-// MARK: XML Metadata
-
-public enum ElementToJsonStrategy {
```
tjprescott: Moved into `XMLMap.swift`

sdk/core/AzureCore/Source/DataStructures/XMLTree.swift
```diff
@@ -0,0 +1,101 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+// MARK: XML Tree
+
+internal class XMLTree {
```
tjprescott: No changes to this code. It was simply moved in here from a larger file for maintainability.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -133,9 +133,12 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
             return try self.init(accountUrl: blobEndpoint, credential: sasCredential, withOptions: options)
         }
 
-    // MARK: Private Methods
+    public static func url(forHost host: String, container: String, blob: String) -> URL? {
+        let urlString = "\(host)/\(container)/\(blob)".addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)
```
bsiegel: The URLSessionTransport will perform the percent-encoding before dispatching the request, do we also need to do it here?

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+        ]
+        let url = client.url(forTemplate: urlTemplate, withKwargs: pathParams)
+        let chunkSize = endRange - startRange
+        var buffer = Data(capacity: chunkSize)
+        do {
+            guard let fileHandle = try? FileHandle(forReadingFrom: uploadSource) else {
+                throw AzureError.general("Unable to open for reading: \(uploadSource)")
+            }
+            if #available(iOS 13.0, *) {
+                try fileHandle.seek(toOffset: UInt64(startRange))
```
bsiegel: Is there not an equivalent of `try?` for methods that may throw but return void? Odd that we have to encase the thing in a `do... catch` block just because the method doesn't produce a return value.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+
+        if let transactionalContentMd5 = transactionalContentMd5 {
+            headers[.contentMD5] = String(data: transactionalContentMd5, encoding: .utf8)
+        }
+
+        if let transactionalContentCrc64 = transactionalContentCrc64 {
+            headers[.contentCRC64] = String(data: transactionalContentCrc64, encoding: .utf8)
+        }
+
+        if let requestId = requestId { headers["x-ms-client-request-id"] = requestId }
```
bsiegel: We should probably use the HTTPHeader and StorageHTTPHeader values here (and below) instead of raw strings.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+        delegate: BlobUploadDelegate? = nil,
+        source: URL,
+        name: String,
+        container: String,
+        properties: BlobProperties? = nil,
+        options: UploadBlobOptions? = nil
+    ) throws {
+        let manager = FileManager.default
+        guard manager.fileExists(atPath: source.path) else {
+            throw AzureError.general("File not found: \(source.path)")
```
bsiegel: It occurs to me we should probably make some Storage-specific error subclasses.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+        let handle = try FileHandle(forReadingFrom: uploadSource)
+        defer { handle.closeFile() }
+        return handle.readDataToEndOfFile()
+    }
+
+    /// Uploads the entire blob in a parallel fashion.
+    /// - Parameters:
+    ///   - group: An optional `DispatchGroup` to wait for the download to complete.
+    ///   - completion: A completion handler called when the download completes.
+    public func complete(inGroup group: DispatchGroup? = nil, then completion: @escaping () -> Void) throws {
```
bsiegel: If we never store the value of the `completion` parameter, does it need to be marked `@escaping`? I admit I don't know if I fully understand the do's and don'ts of `@escaping`, but I -think- that it's only needed if the lifetime of the parameter outlives that of the method execution.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+
+        // Construct parameters
+        var queryParams = [
+            ("comp", "blocklist")
+        ]
+        if let timeout = options.timeout { queryParams.append(("timeout", String(timeout))) }
+
+        // Construct headers
+        var headers = HTTPHeaders([
+            .contentType: "application/xml; charset=utf-8",
```
bsiegel: Maybe we should pull the content-types (this one & "octet-stream") up into constants as well. Seems like they're well-defined values needed for the Storage APIs.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+
+        if let transactionalContentMd5 = transactionalContentMd5 {
+            headers[.contentMD5] = String(data: transactionalContentMd5, encoding: .utf8)
+        }
+
+        if let transactionalContentCrc64 = transactionalContentCrc64 {
+            headers[.contentCRC64] = String(data: transactionalContentCrc64, encoding: .utf8)
+        }
+
+        if let requestId = requestId { headers["x-ms-client-request-id"] = requestId }
```
bsiegel: Same comment here - we should probably use the HTTPHeader and StorageHTTPHeader values here (and below) instead of raw strings.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
bsiegel: Also applies to other uses of `@escaping` below

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -0,0 +1,540 @@
...
+        }
+        return blockList
+    }
+
+    /// Parses the blob length from the content range header: bytes 1-3/65537
+    private func parseLength(fromContentRange contentRange: String?) throws -> Int {
+        let error = AzureError.general("Unable to parse content range: \(contentRange ?? "nil")")
+        guard let contentRange = contentRange else { throw error }
+        // First, split in space and take the second half: "1-3/65537"
+        guard let byteString = contentRange.split(separator: " ", maxSplits: 1).last else { throw error }
```
bsiegel: Arguably this step isn't needed, but I don't feel strongly about removing it.

sdk/core/AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -189,7 +189,8 @@ public class LoggingPolicy: PipelineStage {
     }
 
     private func redact(url: String) -> String? {
-        guard var urlComps = URLComponents(string: url) else { return nil }
+        guard let encodedUrl = url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else { return nil }
```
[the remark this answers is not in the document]
tjprescott: Originally the step was moved as close to the Transport network request as possible. The problem, so I'm thinking I should remove it from there as well.

sdk/storage/AzureStorageBlob/Source/DataStructures/BlobDownloadFinalOperation.swift
```diff
@@ -28,7 +28,7 @@ import AzureCore
 import CoreData
 import Foundation
 
-internal class BlobOperation: ResumableTransfer {
+internal class BlobDownloadFinalOperation: ResumableTransfer {
```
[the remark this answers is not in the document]
tjprescott: There isn't an initial upload operation actually! This was the reason things got more complex. Essentially, this is how it goes:

==Blob Download==
1. Initial Operation must happen to get the actual blob size. Once this happens, the true number of blocks necessary can be calculated.
2. All of the remaining blocks can be downloaded in parallel.
3. Once all of the blocks are finished, the "Final" download operation executes, which marks the entire Blob operation as finished.

==Blob Upload==
1. The file size, and thus all of the blocks, are known. They can all be uploaded in parallel.
2. Once all of the blocks are staged, the block list can be committed, which finalizes the uploaded blob. This is the upload's "Final" operation.

Download has a key step in the beginning whereas upload has a key step at the end. I can definitely chop these into smaller files.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Yes. The URL initializer will fail if the string is not percent encoded. I suppose I could convert the URL objects to strings and see what repercussions that would have.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: Yeah, let me take a look :)

sdk/core/AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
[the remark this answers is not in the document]
tjprescott: Done

sdk/storage/AzureStorageBlob/Source/DataStructures/BlobDownloadFinalOperation.swift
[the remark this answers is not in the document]
tjprescott: Split the Upload and Download operations into separate files.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: The logic is all plumbed through with URL including the Core Data models, which does require the string to be percent encoded.

Refactoring from URL to String introduces a bunch of extra work. If it would truly be valuable to make this change, I would recommend create a stand-alone issue to be addressed separately.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: I took a look at this section and it's wrapped in a do/catch to catch various file-related errors that could occur. I updated the logic and removed the TODO. It now calls a `completion(.failure(error))` to ensure that any file errors are surfaced to the developer instead of being swallowed.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: Done.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: I added a fileSystem error type under AzureError. I did a dedicated pass to replace `AzureError.general` with more appropriate categories. We should discuss this as part of our public API review.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: Short answer is yes. If you remove it, you get the following:
`Escaping closure captures non-escaping parameter 'completion'`

The issue is that the completion closure needs to survive the lifetime of the block that called it. As soon as the async method starts, the original block exits, and thus the completion closure must be marked `@escaping` or it will be released when the calling block finishes.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: So, I could create an enum with common ContentType values, in which case the call looks like this:

```Swift
var headers = HTTPHeaders([
  .accept: ContentType.xml.rawValue, // new
  .accept: "application/xml"         // old
])
```

I'm not sure the the enum adds value here due to the awkward calling mechanism and need to use rawValue. Also, I don't think we'd be able to come up with sensible names for each type, so we'd have to leave the raw string approach open.

Plain constants would work, but there's still the question of having names for each possible value, and they would be much less discoverable than an enum.

My preference here would be to keep the raw string.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: Yep, done.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: Good point! Fixed.

sdk/storage/AzureStorageBlob/AzureStorageBlob.xcodeproj/project.pbxproj
```diff
@@ -76,6 +78,8 @@
 		0A2868B423FF38AA0084CF15 /* MultiBlobTransfer+CoreDataProperties.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = "MultiBlobTransfer+CoreDataProperties.swift"; path = "/Users/travisprescott/repos/azure-sdk-for-ios/sdk/storage/AzureStorageBlob/Source/Data/MultiBlobTransfer+CoreDataProperties.swift"; sourceTree = "<absolute>"; };
 		0A2868B523FF38AA0084CF15 /* BlobTransfer+CoreDataProperties.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = "BlobTransfer+CoreDataProperties.swift"; path = "/Users/travisprescott/repos/azure-sdk-for-ios/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift"; sourceTree = "<absolute>"; };
 		0A2868B623FF38AA0084CF15 /* BlockTransfer+CoreDataProperties.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; name = "BlockTransfer+CoreDataProperties.swift"; path = "/Users/travisprescott/repos/azure-sdk-for-ios/sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataProperties.swift"; sourceTree = "<absolute>"; };
```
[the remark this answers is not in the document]
tjprescott: Fixed. Since it was just these three files that had this issue, it looks like it happens when you use the "Create NSManagedObject Subclass..." wizard in the IDE, and there are no obvious options to affect this. Removing the file references and re-adding them fixed the problem.

tjprescott:
@bsiegel addressed your comments and pushed changes.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I don't think it's worth changing from URL to string, not least because it reduces fidelity of the parameter's type. In my experience, encoding becomes a pretty rich source of bugs because it's very difficult to tell if you're dealing with an encoded URL or not, and you can easily accidentally wind up double-encoding things.

Based on what you mentioned (the initializer will fail if not encoded) - can we assume that any instance of URL *must* be encoded already? If that's the case, that helps resolve this complexity by a ton.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
bsiegel: Ah, okay I think I understand that. And having a compile warning when it's needed also very much helps resolve that confusion too :)

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
bsiegel: I don't feel strongly about it.

bsiegel:
@tjprescott Only one remaining question, otherwise good to go. Thanks!

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Yes, that should be the case. Actually, by the same logic it might make sense to make the URL parameter in `HTTPRequest` a URL object instead of a string. I will explore that possibility. Then we are consistently using URLs all around.

PR #232, opened 2020-03-23
Plumb errors and fix pause/resume.

tjprescott:
Fixes pause/resume issues, reachability and crash recovery for download e2e.

### DIFF
--- a/sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
+++ b/sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
@@ -50,6 +50,24 @@ public enum TransferState: Int16 {
             return "Deleted"
         }
     }
+
+    public var pauseable: Bool {
+        switch self {
+        case .unknown, .pending, .inProgress, .failed:
+            return true
+        case .paused, .complete, .canceled, .deleted:
+            return false
+        }
+    }
+
+    public var resumable: Bool {
+        switch self {
+        case .paused, .failed:
+            return true
+        case .unknown, .pending, .inProgress, .complete, .canceled, .deleted:
+            return false
+        }
+    }
 }
 
 public enum TransferType: Int16 {
--- a/sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
+++ b/sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
@@ -53,9 +53,9 @@ public enum TransferState: Int16 {
 
     public var pauseable: Bool {
         switch self {
-        case .unknown, .pending, .inProgress, .failed:
+        case .unknown, .pending, .inProgress:
             return true
-        case .paused, .complete, .canceled, .deleted:
+        case .paused, .complete, .canceled, .deleted, .failed:
             return false
         }
     }
--- a/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
@@ -26,10 +26,10 @@
 
 import Foundation
 
-public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse) -> Void
+public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse?) -> Void
 public typealias HTTPResultHandler<T> = ResultHandler<T, Error>
 public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
-public typealias OnRequestCompletionHandler = (PipelineRequest) -> Void
+public typealias OnRequestCompletionHandler = (PipelineRequest, Error?) -> Void
 public typealias OnResponseCompletionHandler = (PipelineResponse) -> Void
 public typealias OnErrorCompletionHandler = (PipelineError, Bool) -> Void
 
@@ -70,7 +70,7 @@ public protocol PipelineStage {
 /// Default implementations for `PipelineStage`.
 extension PipelineStage {
     public func on(request: PipelineRequest, then completion: @escaping OnRequestCompletionHandler) {
-        completion(request)
+        completion(request, nil)
     }
 
     public func on(response: PipelineResponse, then completion: @escaping OnResponseCompletionHandler) {
@@ -85,7 +85,19 @@ extension PipelineStage {
         request pipelineRequest: PipelineRequest,
         then completion: @escaping PipelineStageResultHandler
     ) {
-        on(request: pipelineRequest) { request in
+        on(request: pipelineRequest) { request, error in
+            // if error occurs during the onRequest phase, back out and
+            // propagate immediately
+            if let error = error {
+                let pipelineResponse = PipelineResponse(
+                    request: request.httpRequest,
+                    response: nil,
+                    logger: request.logger,
+                    context: request.context
+                )
+                let pipelineError = PipelineError(fromError: error, pipelineResponse: pipelineResponse)
+                completion(.failure(pipelineError), nil)
+            }
             self.next!.process(request: request) { result, httpResponse in
                 switch result {
                 case let .success(pipelineResponse):
--- a/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
@@ -96,7 +96,12 @@ extension PipelineStage {
                     context: request.context
                 )
                 let pipelineError = PipelineError(fromError: error, pipelineResponse: pipelineResponse)
-                completion(.failure(pipelineError), nil)
+                self.on(error: pipelineError) { _, handled in
+                    if !handled {
+                        completion(.failure(pipelineError), nil)
+                        return
+                    }
+                }
             }
             self.next!.process(request: request) { result, httpResponse in
                 switch result {
@@ -108,6 +113,7 @@ extension PipelineStage {
                     self.on(error: pipelineError) { error, handled in
                         if !handled {
                             completion(.failure(error), httpResponse)
+                            return
                         }
                     }
                 }
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/BlobDownloadOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/BlobDownloadOperations.swift
@@ -31,7 +31,7 @@ import Foundation
 internal class BlobDownloadInitialOperation: ResumableTransfer {
     // MARK: Initializers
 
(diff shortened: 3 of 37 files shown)


sdk/storage/AzureStorageBlob/Source/DataStructures/BlockOperations.swift
```diff
@@ -69,18 +61,33 @@ public class BlockOperation: ResumableTransfer {
...
+                    let contentLength = blobProperties.contentLength ?? 0
+                    downloader.progress += contentLength
+                    transfer.state = .complete
+                    self.notifyDelegate(withTransfer: transfer)
+                case let .failure(error):
+                    var err = error
+                    if let pipelineError = error as? PipelineError {
+                        err = pipelineError.innerError
+                    }
+                    transfer.state = .failed
```
bsiegel: Should `state` be a property that, when updated, contains the logic to update the parent as appropriate as well? I feel like it might be a good way to make sure that all the "state" properties always get updated correctly.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
```diff
@@ -121,7 +121,8 @@ internal class ChunkUploader {
             let tempData = fileHandle.readData(ofLength: chunkSize)
             buffer.append(tempData)
         } catch {
-            completion(.failure(error), HTTPResponse(request: nil, statusCode: nil))
+            let request = try? HTTPRequest(method: .put, url: url, headers: HTTPHeaders())
```
bsiegel: If we're synthesizing a fake response here just to make the completion handler happy, maybe we should rethink the design of the handler. Do we need to supply a response at all?

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
```diff
@@ -35,19 +35,36 @@ extension URLSessionTransferManager: ResumableOperationQueueDelegate {
...
+                // Block transfers should propagate up to their BlobTransfer and only notify that the
+                // entire Blob transfer failed.
+                if let transferError = (transfer as? BlobTransfer)?.error as? NSError {
+                    if [-1009, -1005].contains(transferError.code) {
+                        pause(transfer: transfer)
+                    } else {
+                        delegate?.transferManager(self, didFailTransfer: transfer, withError: transferError)
+                    }
+                } else {
+                    // ignore BlockTransfer failures
```
bsiegel: I know it's not implemented yet but we should revisit this when MultiBlob transfers are in.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -110,6 +105,28 @@ public final class URLSessionTransferManager: NSObject, TransferManager, URLSess
         return transfers[index]
     }
 
+    public func transfer(forName name: String, type: TransferType? = nil) -> Transfer? {
```
bsiegel: Unsure if we want to ultimately keep this. I was working on adding unique IDs for each transfer, which the customer can use to retrieve a Transfer object by its ID. Retrieving by file name or part thereof seems like it could become a source of bugs.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -52,10 +52,6 @@ public class StorageBlobClient: PipelineClient, PagedCollectionDelegate {
         }
     }
 
-    private lazy var manager: URLSessionTransferManager = {
```
bsiegel: Ultimately I want to revisit this, from a user's perspective I'm working to avoid exposing the transfer manager at all, and just run those operations through the public API surface of StorageBlobClient. I went the other way in my branch, promoting the TransferDelegate to a (public) property of the Client and hiding the manager as private. I understand why this was moved to the Options, but ultimately I think what we are going to have to do is:

1. Serialize the transfer-specific Options (e.g. DownloadBlobOptions) associated with a transfer to CoreData so that we can access them if the app is restarted, and
2. Also store the subset of StorageBlobClient properties / Options needed to determine whether any given Client is "suitable" to be used for a transfer. That might allow us to get rid of the "downloader/uploader" property on the Transfer, and instead just keep a record of all Clients created, and if that list changes, determine if there is a "suitable" client for any transfers without an owning client, and assign & resume them.

There might be other approaches that would work better, but for now this is the way I'm thinking about avoiding the need for us to try to re-create a Client for a transfer that was not completed.

sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
```diff
@@ -50,6 +50,24 @@ public enum TransferState: Int16 {
             return "Deleted"
         }
     }
+
+    public var pauseable: Bool {
+        switch self {
+        case .unknown, .pending, .inProgress, .failed:
+            return true
```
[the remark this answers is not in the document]
bsiegel: We talked about this after standup today, failed is considered resumable because that actually has "retry" semantics rather than resume. I'm still not clear whether it's correct that a failed transfer is also "pausable" though.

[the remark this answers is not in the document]
tjprescott: Good point.  The intent was to intercept any "lost connection" failures to immediately pause and pre-empt the failed state. However, that isn't terribly reliable, so I had to make the failed state resumable. I agree--it shouldn't be both. I will make it simply resumable.

sdk/storage/AzureStorageBlob/Source/DataStructures/BlobDownloadOperations.swift
```diff
@@ -42,11 +42,19 @@ internal class BlobDownloadInitialOperation: ResumableTransfer {
...
 
-        var operations = [BlockOperation]()
-        guard let downloader = transfer.downloader else { return }
+        // The intialDownloadOperation marks the transfer complete, so
+        // if blocks remain, we must reset the state to inProgress.
+        transfer.state = .inProgress
+
+        var operations = [ResumableOperation]()
+        let finalOperation = BlobDownloadFinalOperation(withTransfer: transfer, queue: opQueue)
+        operations.append(finalOperation)
```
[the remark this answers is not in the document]
tjprescott: No, because the final operation is dependent on all the other operations completing. The operations don't get added to the operations queue until all of those dependencies are defined.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -26,10 +26,10 @@
 
 import Foundation
 
-public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse) -> Void
+public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse?) -> Void
 public typealias HTTPResultHandler<T> = ResultHandler<T, Error>
 public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
-public typealias OnRequestCompletionHandler = (PipelineRequest) -> Void
+public typealias OnRequestCompletionHandler = (PipelineRequest, Error?) -> Void
```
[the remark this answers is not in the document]
tjprescott: Hmm... we should chat about this one. It's a short section of code. Essentially, I'm thinking that OnResponseCompletionHandler should also have an optional Error (which is probably the opposite of what you'd expect).

sdk/storage/AzureStorageBlob/Source/DataStructures/BlockOperations.swift
tjprescott: It's certainly possible, but I'm not sure it would be a simple value add. There are states for which a change in the child shouldn't propagate up to the parent. It's reasonable to say that if any block failed, the whole blob transfer is a failure, but for example, when a block is completed, the blob download itself is very likely not.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
tjprescott: It's a tradeoff between SDK code and client code. If I did make it `HTTPResponse?`, then the client code is going to have to add guard statements or otherwise unwrap that optional every time they want to use it. I could make it `HTTPResponse!` but (assuming XCode or a linter don't howl in anger) the app would crash if the user tried to use the response that was nil.

Thus I settled on preferring to always return some kind of Response object, even if it is junk. Ugly code inside the SDK, but I figure I'd rather have it there.

That was my rationale, but I'm happy to examine other options!

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
tjprescott: I'm actually thinking it should notify the delegate that the block transfer failed. I was previously notifying just the parent when blocks completed, but had to switch notifying both. The idea being that the app developer can decide which perspective they care about. So I am open to bubbling this failure up. Is that what you were thinking?

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: Yep, I agree. :)

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Now that we have some implementation, we should use our time Wednesday to revisit this problem with the Android team as well. When we discussed it before it was all hypothetical.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageStreamUploader.swift
bsiegel: Ahh yes, I see - the problem here is encoded in the type. The assumption is that regardless of whether the operation succeeds or fails, there will always be a HTTPResponse available. I don't think that's a valid assumption, so I would be in favor of changing the type signature for ResultHandler to make the response optional. It makes perfect sense that there will sometimes be errors for which no response (or even no request) object is available. For comparison, OkHttp has a `Callback` interface with an `onResponse(Request, Response)` and an `onFailure(Request, Exception)` so they seem to at least make the assumption that there will always be a request available.

sdk/core/AzureCore/Source/DataStructures/ResumableOperation.swift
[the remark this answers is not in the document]
tjprescott: This has been fixed.

PR #238, opened 2020-03-31
Upload page demo

tjprescott:
Removes the TransferManager demo. Use the review filters to ignore deleted files :)

Adds the upload page to the SwiftDemo. Still working through some transfer coordination.


examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -93,10 +87,14 @@ class BlobTableViewController: UITableViewController, MSALInteractiveDelegate {
         fetchData(self)
     }
 
-    override func viewWillDisappear(_: Bool) {
-        transferManager.reachability?.stopListening()
-        transferManager.pauseAll()
-    }
+//    override func viewDidAppear(_: Bool) {
```
tjprescott: I'm still looking at this... it is problematic in this tabbed view situation.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -140,22 +138,71 @@ class BlobTableViewController: UITableViewController, MSALInteractiveDelegate {
     private func reloadTableView() {
         DispatchQueue.main.async { [weak self] in
             self?.tableView.reloadData()
-            self?.refreshControl?.endRefreshing()
         }
     }
 
+    private func reloadTableView(forTransfer transfer: BlobTransfer) {
```
tjprescott: Supposedly better for performance, but it causes some animation glitches that are very annoying.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
```diff
@@ -191,11 +199,12 @@ public final class URLSessionTransferManager: NSObject, TransferManager, URLSess
             // Unknown transfers are invalid
             remove(transfer: transfer)
         }
+        print("QUEUEING \(operations.count) OPERATIONS FOR BLOB \(transfer.hash)")
```
tjprescott: All of these print statements will be removed when the PR goes live.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+        }
+        cell.backgroundColor = .white
+
+        // Match any blobs to existing transfers.
+        // Update download map and progress.
+        if let transfer = blobClient.transfer(
+            forBlob: blobName,
+            inContainer: AppConstants.videoContainer,
+            ofType: .download
+        ) {
```
tjprescott, line 178: Using `transfer(withId)` doesn't work here because the whole point is to see if there is a transfer in progress for a blob on the service. The transfer IDs are generated randomly by transfer manager, so there's nothing to tie them to something on the service.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    func transfer(
+        _ transfer: Transfer,
+        didUpdateWithState _: TransferState,
+        andProgress progress: Float?
+    ) {
+        if let blobTransfer = transfer as? BlobTransfer {
+            print("BLOB DOWNLOAD \(blobTransfer.hash) DID UPDATE WITH PROGRESS %\((progress ?? -0.01) * 100)")
+            reloadTableView()
```
tjprescott, line 259: Trying to selectively reload cells caused all kinds of visual and performance glitches. Then again, maybe I was just doing it wrong.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
```diff
@@ -29,23 +29,23 @@ import CoreData
...
-    @nonobjc internal class func fetchRequest() -> NSFetchRequest<BlobTransfer> {
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<BlobTransfer> {
         return NSFetchRequest<BlobTransfer>(entityName: "BlobTransfer")
     }
 
-    @NSManaged public internal(set) var destination: URL?
-    @NSManaged internal var endRange: Int64
-    @NSManaged public internal(set) var error: Error?
-    /// The unique identifier for this transfer operation.
-    @NSManaged public internal(set) var id: UUID
```
tjprescott, line 40: @bsiegel how did you get these `internal(set)` annotations?  Did you just manually edit?

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -39,9 +39,6 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     /// The maximum size of a single chunk in a blob upload or download.
     public let maxChunkSize: Int
 
-    // TransferManager configuration
-    internal var transferManager: TransferManager?
```
tjprescott, line 43: This was not used anywhere

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -466,6 +457,41 @@ public final class StorageBlobClient: PipelineClient {
...
+    ///   - blob: The name of the blob.
+    ///   - container: The name of the container.
+    ///   - options: A `DownloadBlobOptions` object to control the download operation.
+    /// - Throws: `AzureError.fileSystem` when cache directory cannot be found.
+    /// - Returns: Destination `URL` for the blob download.
+    public func downloadUrl(
+        forBlob blob: String,
+        inContainer container: String,
+        withOptions options: DownloadBlobOptions? = nil
+    ) throws -> URL {
```
tjprescott, line 472: This is definitely worthwhile to keep in the client. This is the logic that I was previous created a download specifically for, and it doesn't make sense to make the client copy this snippet.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -489,7 +515,7 @@ extension StorageBlobClient: TransferDelegate {
     public func transfer(
         _ transfer: Transfer,
         didUpdateWithState state: TransferState,
-        andProgress progress: TransferProgress?
+        andProgress progress: Float?
```
tjprescott, line 518: `TransferProgress` still exists for raw operations. The TransferManager ones use plain ol' `Float`.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -505,13 +531,13 @@ extension StorageBlobClient: TransferDelegate {
     }
 
     /// :nodoc:
-    public func uploader(for transfer: BlobTransfer) -> BlobUploader? {
-        transferDelegate?.uploader(for: transfer)
+    public func client(forRestorationId restorationId: String) -> PipelineClient? {
+        transferDelegate?.client(forRestorationId: restorationId)
```
tjprescott, line 535: Pretty unhappy with these at the moment. Not sure they are in any way better than `uploader/downloader(for transfer)`

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -607,14 +632,60 @@ extension StorageBlobClient {
...
+    /// - Parameters:
+    ///   - blob: The name of the blob.
+    ///   - container: The name of the container.
+    ///   - type: The `TransferType` to retrieve. Leave blank to retrieve either uploads or downloads.
+    /// - Returns: The matching `BlobTransfer` if found, or nil.
+    public func transfer(
+        forBlob blob: String,
+        inContainer container: String,
+        ofType type: TransferType? = nil
+    ) -> BlobTransfer? {
```
tjprescott, line 654: Definitely need some way to match a transfer to something that is a property of the blob on the service. The UUID used for transfer ID is completely random and doesn't serve this purpose. The closest thing to a key for a blob that I'm aware of is its full URL.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -607,14 +632,60 @@ extension StorageBlobClient {
...
+                }
+            }
+        }
+        return nil
+    }
+
+    /// Parse a blob URL into its component parts.
+    /// - Parameter url: The complete blob URL to parse.
+    /// - Returns: A triple containing the parsed results.
+    public func parse(url: URL) -> (baseURL: String, blobName: String, containerName: String)? {
```
tjprescott, line 680: This again is very helpful to have, especially give we tossed the upload/download signatures that take a URL directly and that the transfer objects store URLs exclusively.

bsiegel:
Adding notes from our arch discussion - after rebase, we still want to validate that this PR makes the following changes:

- [ ] Don't send blocktransfer updates & keep blocktransfer private
- [ ] Don't send updates if the state and progress are the same
- [ ] Remove unnecessary use of downloader / uploader in transfer manager (I think this is done, it's still used but I believe it's necessary now)
- [ ] Use implicitly-unwrapped optionals for enums that are serialized to/from CoreData

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
tjprescott: I removed this.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager.swift
tjprescott: Removed.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
tjprescott: Updated to use your transfers extensions.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Moved to PathHelpers.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Updated to use transfer collection extensions.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: No longer needed.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: I removed `options(forRestorationId:...)` so you only get the client. I store the options and the properties as part of the CoreData object, and I'm happy with that for now.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -180,8 +180,8 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
         guard let blobName = cell.keyLabel.text else { return }
         guard let containerName = AppConstants.videoContainer else { return }
         guard let blobClient = blobClient else { return }
-        let destination = StorageBlobClient.PathHelper
-            .localUrl(inDirectory: StorageBlobClient.PathHelper.cacheDir, forBlob: blobName, inContainer: containerName)
+        // FIXME: Major bug--cannot rehydrate transfers due to hard coding of local paths
```
tjprescott, line 183: Uploads and downloads cannot be rehydrated due to changes made to use hardcoded URLs. You can pause and resume but if you try and close the app and reload, both operations will fail. It will need to be fixed in a separate PR.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -221,28 +221,22 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
 }
 
 extension BlobDownloadViewController: TransferDelegate {
-    func client(forRestorationId restorationId: String) -> PipelineClient? {
-        guard restorationId == "download" else { return nil }
+    func client(forRestorationId _: String) -> PipelineClient? {
         return blobClient
     }
 
-    func options(forRestorationId restorationId: String) -> AzureOptions? {
```
tjprescott: Got rid of this. I serialize the options to a JSON string and work with that. I can't simply make them Transformable because they would need to work with `@objc`. It's a lot lower bar to make something `Codable` vice `@objc`, so I just handle this in the SDK. We can revisit later.

examples/AzureSDKDemoSwift/Source/Common.swift
```diff
@@ -210,6 +210,8 @@ extension TransferState {
             return .systemRed
         case .deleted:
             return .systemRed
+        @unknown default:
```
tjprescott, line 213: Turns out iOS has this solution, making it unnecessary to have either a `.unknown` or use the implicitly unwrapped optional.

sdk/core/AzureCore/Source/Pipeline/Pipeline.swift
```diff
@@ -40,10 +40,13 @@ internal class Pipeline {
             }
             prevPolicy = policy
         }
-        // FIXME: Handle case where there are no policies
```
tjprescott, line 43: Figured I would knock this fixme out.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
```diff
@@ -83,6 +83,58 @@ public class BlobTransfer: NSManagedObject, TransferImpl {
         }
     }
 
+    // FIXME: The fact that BlobProperties is a struct causes serious problems here. It will always return nil if
+    // allowed to be optional.
+    internal var properties: BlobProperties {
```
tjprescott, line 88: This can be revisited later, if necessary. Since `BlobProperties` is a struct, making this optional always returns a nil value. This was the quickest solution. Another solution (not convinced it is better) would be to convert this to a class.

sdk/storage/AzureStorageBlob/Source/DataStructures/Enumerations.swift
```diff
@@ -136,9 +136,10 @@ public enum TransferState: Int16 {
     /// Indicates whether the transfer is currently in a state that can be resumed.
     public var resumable: Bool {
         switch self {
-        case .paused, .failed:
+        // TODO: Switch .pending and .inProgress back to false eventually... but not now.
+        case .paused, .failed, .pending, .inProgress:
```
tjprescott, line 140: Unless we can guarantee that the OS will have an opportunity to switch in progress or pending transfer to paused when the app crashes, transfers that should resume are not resuming. I haven't found a foolproof way to guarantee that.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -261,7 +261,7 @@ public struct DownloadBlobOptions: AzureOptions {
 }
 
 /// User-configurable options for the blob upload operations.
-public struct UploadBlobOptions: AzureOptions {
+public class UploadBlobOptions: AzureOptions, Codable {
```
tjprescott, line 264: Again, simplest fix. Would also be appropriate to remain a struct but change `BlobTransfer.downloadOptions/.uploadOptions` to not be optional.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -27,6 +27,20 @@
 import AzureCore
 import Foundation
 
+internal class StorageJSONDecoder: JSONDecoder {
```
tjprescott, line 30: Needed to use these in BlobTransfer so I moved them out, though they remain internal.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -62,9 +62,10 @@ public final class StorageBlobClient: PipelineClient {
 
     fileprivate var managing = false
     fileprivate lazy var manager: TransferManager = {
-        let instance = URLSessionTransferManager(delegate: self, logger: self.logger)
-        instance.loadContext()
-        return instance
+        var manager = URLSessionTransferManager.shared
```
tjprescott, line 65: Re-verified that creating multiple Transfer Managers (really creating multiple ManagedObjectContexts) causes unworkable problems with CoreData. You will reliably, but unpredictably get CoreData failures because "blocks was invalid" due to it trying to save in different contexts.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
bsiegel: Yep, I did. Unfortunately I don't know any way to hint to the codegen how you want the access defined for CoreData properties.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I was trying to think a little more about this. Is there a reason we don't want the customer to get a TransferProgress passed to their delegate methods? I could see places where they want to report on the bytes transferred / total bytes. Sure they can cast to BlobTransfer and read from there, but what if it's a MultiBlobTransfer?

Also I was concerned about comparison of Floats. I don't know about Swift, but in some languages comparing floats for equality can lead to bugs because of precision issues, and determining if they differ by some epsilon value is preferred in these cases:
```
if (a == b) // BAD
if (b - a <= Float.EPSILON) // GOOD
```

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
tjprescott: No problem. I thought there was some "magic" in the code gen that necessitated you needing to regenerate the code through XCode to make properties accessible. Seems that's not the case, so I think it is fine to manually update these.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Cool, I was thinking about maybe in the future, having the restoration ID be part of the client constructor instead, and keeping a static registry of (restoration ID -> client). Any time a new client is registered, we could see if there are unfulfilled transfers and potentially resume them if their restoration ID exists in the registry. Even if we don't do that last part, we could avoid needing to have the `client(forRestorationId:)` delegate method.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
bsiegel: Definitely will want to chat and understand why this is the case. I'm okay addressing in a followup PR.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
bsiegel: I think this is a good approach!

examples/AzureSDKDemoSwift/Source/Common.swift
bsiegel: Interesting, will read up on this!

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
bsiegel: I don't quite follow what's going on here. UploadBlobOptions is also a struct, doesn't `uploadOption` exhibit the same behavior?

sdk/storage/AzureStorageBlob/Source/DataStructures/Enumerations.swift
bsiegel: Makes sense to me. All transfers should be resumable unless explicitly block-listed.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Sounds good, but maybe we should put them in a separate file then.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I guess I'm not surprised. Using a static shared instance shouldn't be too much of a problem, but there's a bug here that now each newly-initialized Client will overwrite the `delegate` and `logger` properties of the shared TransferManager instance. We'll need to have some way of setting those properties on each new Transfer instead of on the TransferManager itself.

bsiegel: I guess I'm not surprised. Using a static shared instance shouldn't be too much of a problem, but there's a bug here that now each newly-initialized Client will overwrite the delegate and logger properties of the shared TransferManager instance. We'll need to have some way of setting those properties on each new Transfer instead of on the TransferManager itself.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Sounds good, but maybe we should put them in a separate file then.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
bsiegel: I don't understand why this would be the case, or why it would matter between a struct and a class. Is the issue related to some semantics of how/when the getter for a computed property that returns a struct is called? Some sort of weird optimization for structs?

Podfile.lock
```diff
@@ -13,6 +13,6 @@ SPEC REPOS:
 SPEC CHECKSUMS:
   MSAL: 74bf430ec0f46dd4c79dde2596e2b553610ccab1
 
-PODFILE CHECKSUM: a5fd21670ef5dd405bcb0d5adf5591f8d31ee53f
+PODFILE CHECKSUM: e70aefde8398e9dc7b42edaa007f940c8a787366
 
-COCOAPODS: 1.9.1
```
bsiegel, line 18: Might want to just update your Cocoapods rather than committing this.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
```diff
@@ -83,6 +83,58 @@ public class BlobTransfer: NSManagedObject, TransferImpl {
...
+            }
+        }
+    }
+
+    internal var uploadOptions: UploadBlobOptions? {
+        get {
+            guard transferType == .upload,
+                let options = rawOptions,
+                let jsonData = options.data(using: .utf8) else { return nil }
+            return try? StorageJSONDecoder().decode(UploadBlobOptions.self, from: jsonData)
```
bsiegel, line 111: Do you think we should memoize this, so that multiple `get`s on this property don't need to invoke the JSON parser more than once? Likewise we could optimize the `set` the same way so that if the new value is equal to the existing one, we don't invoke the JSON serializer.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
```diff
@@ -29,7 +29,7 @@ import CoreData
 import Foundation
 
 extension BlobTransfer {
-    @nonobjc internal class func fetchRequest() -> NSFetchRequest<BlobTransfer> {
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<BlobTransfer> {
```
bsiegel, line 32: Do these (this one, and the two at the end of the file below) need to be made public? Since `BlobTransfer` is part of the public API, these methods will also become part of the public API.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataProperties.swift
```diff
@@ -29,13 +29,13 @@ import CoreData
 import Foundation
 
 extension BlockTransfer {
-    @nonobjc internal class func fetchRequest() -> NSFetchRequest<BlockTransfer> {
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<BlockTransfer> {
```
bsiegel, line 32: Same question about promoting these to public. It doesn't matter as much here because BlockTransfer is internal. I'm guessing these changes just happened automatically from the code generator and weren't intentional.

sdk/storage/AzureStorageBlob/Source/Data/MultiBlobTransfer+CoreDataProperties.swift
```diff
@@ -29,28 +29,28 @@ import CoreData
 import Foundation
 
 extension MultiBlobTransfer {
-    @nonobjc internal class func fetchRequest() -> NSFetchRequest<MultiBlobTransfer> {
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<MultiBlobTransfer> {
```
bsiegel, line 32: Ditto

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -455,7 +457,7 @@ public final class StorageBlobClient: PipelineClient {
         file sourceUrl: URL,
         toContainer container: String,
         asBlob blob: String,
-        properties: BlobProperties? = nil,
+        properties: BlobProperties,
```
bsiegel, line 460: We should find a way to let this continue to be optional. I don't believe that there's any `BlobProperties` member that is -required- to be set by the customer.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
```diff
@@ -483,8 +483,7 @@ internal class BlobStreamDownloader: BlobDownloader {
                 var blockEnd = index + chunkLength
                 if totalSize < 0 {
                     blockEnd = chunkLength
-                }
-                if blockEnd > totalSize {
+                } else if blockEnd > totalSize {
```
bsiegel, line 486: Good catch, that was definitely a bug I introduced.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -372,32 +358,37 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
 
-    public func resumeAll() {
+    func resumeAll() {
         for transfer in transfers {
             resume(transfer: transfer)
         }
     }
 
-    public func resume(transfer: Transfer) {
+    func ensureUploader(for transfer: BlobTransfer) {
```
bsiegel, line 367: This appears to be unused.

sdk/storage/AzureStorageBlob/Source/StorageBlobClientOptions.swift
```diff
@@ -0,0 +1,283 @@
+// --------------------------------------------------------------------------
```
bsiegel, line 1: It looks like this got duplicated? The existing copy is at `Source/DataStructures/StorageBlobClientOptions.swift`.

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
```diff
@@ -0,0 +1,86 @@
+// --------------------------------------------------------------------------
```
bsiegel, line 1: Duplicated from `Source/Transport/URLSessionTransferManager+ResumableOperationQueueDelegate.swift`

sdk/storage/AzureStorageBlob/Source/DataStructures/Transfer.swift
```diff
@@ -0,0 +1,96 @@
+// --------------------------------------------------------------------------
```
bsiegel, line 1: Duplicated from `Source/Data/Transfer.swift`

examples/AzureSDKDemoSwift/Source/Common.swift
tjprescott: https://www.avanderlee.com/swift/unknown-default-enums-in-swift/

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
tjprescott: I don't rightly understand why this would be the case either, but it was definitely the culprit. Building the struct and examining its value right before returning would show the correct values and then stepping to the next line, immediately after the return, would inexplicably be nil. Making it non-optional worked (no idea why), and making it a class (as I did with the Options) worked. It may be related to the fact that structs are value types instead of reference types.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
tjprescott: I don't think so. I think this may have just been the result of regenerating the NSManagedObject class.

sdk/storage/AzureStorageBlob/Source/DataStructures/Transfer.swift
tjprescott: Probably a merge issue. Wasn't in the XCode project so it didn't cause warnings. I've removed the extra file. Thanks for the catch!

sdk/storage/AzureStorageBlob/Source/DataStructures/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
tjprescott: Removed.

sdk/storage/AzureStorageBlob/Source/StorageBlobClientOptions.swift
tjprescott: Removed.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: Removed.

sdk/storage/AzureStorageBlob/Source/Data/MultiBlobTransfer+CoreDataProperties.swift
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataProperties.swift
tjprescott: Fixed (yes, code gen)

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: I will do this in a follow-up PR and thus left the TODOs in place. I don't want to block Mania's work while I investigate the best path forward, and converting this to a class caused all kinds of errors in XCode, so this was the most straightforward solution.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
tjprescott: Sounds smart to me!

Podfile.lock
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
tjprescott: Done.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Right, ideally I'd like to just have the PersistentContainer be shared (which ultimately yields the ManagedObjectContext), but I think a separate PR would be appropriate, since that would be a non-trivial change.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Done. Placed in `Util/StorageJSONCoding.swift`

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: I agree. Ideally, I'd like the simplest case to not require any registration (or perhaps automatic registration) and not require this delegate method. For now I'm just focused on having something that works as a baseline for improvement.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: I think it's a valid point. At the time, I got rid of it because we were trying to get rid of relying on the uploader/downloader classes, and creating a `TransferProgress` object had such a dependency. Also, I found I was creating the progress objects for the TransferDelegate calls by passing in completed blocks and total blocks for bytes and total bytes as an imprecise proxy anyways. Since we now recreate the uploader and downloader, there's no reason we couldn't update this to use TransferProgress again. I would want to do it as a separate PR though.

To your point about Float comparisons, the `TransferProgress` struct has the same issues. You can return the progress `asFloat` or `asPercent` (which is just `asFloat * 100`). Float was also chosen because that is the native type for the UIProgressView component.

I would not be surprised if Swift is subject to the same floating point precision errors as other languages, but in my mind this falls squarely on the developer's shoulder, who should know that performing equality comparisons on floating point values is generally unsafe. Most typically, they would be taking action base on inequality operators (i.e. do X when progress > 0.5). What alternative would we have? Even with access to the cumulative and total bytes, they'd inevitably be doing that calculation to a float point value themselves.

tjprescott:
@bsiegel pushed changes addressing most comments. Others I agree with but would like to tackle with small, targeted follow-up PRs, as Mania needs the bulk of the functionality in this PR to work on the UX study.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
tjprescott: Well.. to implement the memoization of properties and upload/download options, the most straightforward solution was to make the options classes into structs anyways and then make them non-optional. Doing the memoization required the objects to conform to Equatable. For structs, that can be automatically synthesized, but not for classes. Again, I'll reconsider all of this in a future PR, but for now, the side effect is that if you called `.downloadOptions` on an upload transfer, instead of getting `nil` you'll get a new `DownloadBlobOptions` struct. Certainly non-ideal.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Right, I'm not sure how important it is, but if for some reason a developer wrote `if progress == 1.0` they might not realize that the comparison may not work like they intend. I do agree it's ultimately the developer's responsibility to know about this, but with a TransferProgress they could check if `bytes == totalBytes`.

The use case for this is probably small though, because really they should be implementing whatever logic in `transterDidComplete` instead.

bsiegel: I'm okay with considering reverting this back to `TransferProgress` in a future PR.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
bsiegel: Would it make sense to do the same for BlobProperties? I'm not clear why you resolved the weird issue above by changing the *BlobOptions structs into classes, but for BlobProperties you instead made the property non-optional and return an empty value.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: That's fine, but we have a known bug here so let's at least mark it as a TODO.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
bsiegel: Sounds good, I think we should definitely get to the bottom of what's going on here, but we can address that in a future PR.

examples/AzureSDKDemoSwift/Source/Common.swift
bsiegel: I understand how this works from the consuming side, but I think I'll have to play around with it to understand how it impacts our code which creates an enum from its stored rawValue.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: Agree, we can iterate on this in the future. For now I think it's adding a relatively minimal complexity to make the feature work.

PR #242, opened 2020-04-08
Keep a registry of restoration ID -> Client inside the TransferManager

bsiegel:
This way, a customer only has to provide the restoration ID when
initializing a client. The client is later looked up from the registry
if it's found to be missing from the transfer. If no client exists in
the registry when the transfer is resumed, it will be marked as failed.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
@@ -33,7 +33,11 @@ extension BlobTransfer {
         return NSFetchRequest<BlobTransfer>(entityName: "BlobTransfer")
     }
 
-    @NSManaged internal var clientRestorationId: String
+    /// An identifier used to associate this transfer with the client that created it. When a transfer is reloaded from
+    /// disk (e.g. after an application crash), it can only be resumed once a client with the same `restorationId` has
+    /// been initialized. Attempting to resume it without previously initializing such a client will cause the transfer
+    /// to transition to the `failed` state.
+    @NSManaged public var clientRestorationId: String
     /// The destionation of the transfer. For uploads, this is the blob URL where the file is being uploaded. For
     /// downloads, this is the local path on the device to which the blob is being downloaded.
     @NSManaged public internal(set) var destination: URL?
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
@@ -64,16 +64,16 @@ extension BlobTransfer {
 
 // MARK: Generated accessors for blocks
 
-extension BlobTransfer {
+internal extension BlobTransfer {
     @objc(addBlocksObject:)
-    @NSManaged internal func addToBlocks(_ value: BlockTransfer)
+    @NSManaged func addToBlocks(_ value: BlockTransfer)
 
     @objc(removeBlocksObject:)
-    @NSManaged internal func removeFromBlocks(_ value: BlockTransfer)
+    @NSManaged func removeFromBlocks(_ value: BlockTransfer)
 
     @objc(addBlocks:)
-    @NSManaged public func addToBlocks(_ values: NSSet)
+    @NSManaged func addToBlocks(_ values: NSSet)
 
     @objc(removeBlocks:)
-    @NSManaged public func removeFromBlocks(_ values: NSSet)
+    @NSManaged func removeFromBlocks(_ values: NSSet)
 }
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
@@ -29,19 +29,19 @@ import CoreData
 import Foundation
 
 internal class BlockTransfer: NSManagedObject, TransferImpl {
-    public var operation: ResumableOperation?
+    internal var operation: ResumableOperation?
 
     public var debugString: String {
         return "\t\tTransfer \(type(of: self)) \(hash): Status \(state.label)"
     }
 
-    public var clientRestorationId: String {
+    internal var clientRestorationId: String {
         return parent.clientRestorationId
     }
 }
 
-extension BlockTransfer {
-    internal static func with(
+internal extension BlockTransfer {
+    static func with(
         context: NSManagedObjectContext,
         id: UUID? = nil,
         startRange: Int64,
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
@@ -34,6 +34,10 @@ internal class BlockTransfer: NSManagedObject, TransferImpl {
     public var debugString: String {
         return "\t\tTransfer \(type(of: self)) \(hash): Status \(state.label)"
     }
+
+    public var clientRestorationId: String {
+        return parent.clientRestorationId
+    }
 }
 
 extension BlockTransfer {
@@ -42,7 +46,7 @@ extension BlockTransfer {
         id: UUID? = nil,
         startRange: Int64,
         endRange: Int64,
-        parent: BlobTransfer? = nil
+        parent: BlobTransfer
     ) -> BlockTransfer {
         guard let transfer = NSEntityDescription.insertNewObject(
             forEntityName: "BlockTransfer",
--- a/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
@@ -61,6 +61,7 @@ public protocol Transfer: AnyObject {
 }
 
 internal protocol TransferImpl: Transfer {
+    var clientRestorationId: String { get }
     var operation: ResumableOperation? { get set }
     var state: TransferState { get set }
     var rawState: Int16 { get set }
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -540,7 +540,10 @@ public final class StorageBlobClient: PipelineClient {
         ///   - name: The name of the blob.
         ///   - container: The name of the container
         /// - Returns: A tuple of (`dirName`, `fileName`)
-        public static func pathComponents(forBlob name: String, inContainer container: String) -> (String, String) {
+        public static func pathComponents(
+            forBlob name: String,
+            inContainer container: String
+        ) -> (dirName: String, fileName: String) {
             var defaultUrlComps = "\(container)/\(name)".split(separator: "/").compactMap { String($0) }
             let baseName = defaultUrlComps.popLast()!
             let dirName = defaultUrlComps.joined(separator: "/")
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -39,20 +39,16 @@ public final class StorageBlobClient: PipelineClient {
     /// Options provided to configure this `StorageBlobClient`.
     public let options: StorageBlobClientOptions
 
-    /// The `TransferDelegate` to inform about transfer events.
+    /// The `TransferDelegate` to inform about events from transfers created by this `StorageBlobClient`.
     public weak var transferDelegate: TransferDelegate?
 
     private static let defaultScopes = [
         "https://storage.azure.com/.default"
     ]
 
-    fileprivate var managing = false
-    fileprivate lazy var manager: TransferManager = {
-        var manager = URLSessionTransferManager.shared
-        manager.delegate = self
-        manager.logger = self.logger
-        return manager
-    }()
+    private var managing = false
+    private let manager: TransferManager = URLSessionTransferManager.shared
+    private let restorationId: String
 
     // MARK: Initializers
 
@@ -60,9 +56,21 @@ public final class StorageBlobClient: PipelineClient {
     /// - Parameters:
     ///   - baseUrl: Base URL for the storage account.
     ///   - authPolicy: An `Authenticating` policy to use for authenticating client requests.
+    ///   - restorationId: An identifier used to associate this client with transfers it creates. When a transfer is
+    ///     reloaded from disk (e.g. after an application crash), it can only be resumed once a client with the same
+    ///     `restorationId` has been initialized. If your application only uses a single `StorageBlobClient`, it is
+    ///     recommended to use a value unique to your application (e.g. "MyApplication"). If your application uses
+    ///     multiple clients with different configurations, use a value unique to both your application and the
+    ///     configuration (e.g. "MyApplication.userClient").
     ///   - options: Options used to configure the client.
-    private init(baseUrl: URL, authPolicy: Authenticating, withOptions options: StorageBlobClientOptions? = nil) {
+    private init(
+        baseUrl: URL,
+        authPolicy: Authenticating,
+        withRestorationId restorationId: String,
+        withOptions options: StorageBlobClientOptions? = nil
+    ) {
         self.options = options ?? StorageBlobClientOptions(apiVersion: ApiVersion.latest.rawValue)
+        self.restorationId = restorationId
         super.init(
             baseUrl: baseUrl,
             transport: URLSessionTransport(),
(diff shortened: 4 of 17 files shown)


bsiegel:
I had a thought - instead of marking it as failed when we attempt to resume a transfer and we don't get a client to use, we could just leave the transfer in whatever state it is in immediately after rehydration (I'm guessing it would be `pending` or `paused`?). Though I'm not sure if it matters exactly, because failed transfers can be "resumed" (aka retried) in the same way as pending or paused transfers can be.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -107,6 +102,14 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         return transfers[index]
     }
 
+    func register(client: StorageBlobClient?, forRestorationId restorationId: String) {
+        clients.setObject(client, forKey: restorationId as NSString)
```
tjprescott, line 106: Might be worth adding the extension shown here (https://nshipster.com/nshashtable-and-nsmaptable/) for adding regular subscripting support to NSMapTable.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -85,15 +83,12 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
         self.transfers = [Transfer]()
-        self.logger = ClientLoggers.none
 
         super.init()
         operationQueue.delegate = self
         operationQueue.maxConcurrentOperationCount = 4
     }
 
-    // TODO: This will interfere with trying to use multiple BlobClients simultaneously. Find an alternate
-    // solution such that the minimal set (the NSPersistentContainer) is shared.
```
tjprescott, line 96: Does this PR actually resolve this?

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
```diff
@@ -30,6 +30,7 @@ extension URLSessionTransferManager: ResumableOperationQueueDelegate {
     internal func operation(_ operation: ResumableOperation?, didChangeState state: TransferState) {
         saveContext()
         guard let transfer = operation?.transfer else { return }
+        let delegate = client(forRestorationId: transfer.clientRestorationId)
```
tjprescott, line 33: Ultimately, the delegate needs to be a UIViewController that implements TransferDelegate (or something that has a reference to such a controller) so that the UI can be updated. Does the AzureSDKSwiftDemo app work with this change?

tjprescott:
> I had a thought - instead of marking it as failed when we attempt to resume a transfer and we don't get a client to use, we could just leave the transfer in whatever state it is in immediately after rehydration (I'm guessing it would be pending or paused?). Though I'm not sure if it matters exactly, because failed transfers can be "resumed" (aka retried) in the same way as pending or paused transfers can be.

The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
```diff
@@ -50,6 +51,7 @@ extension BlockTransfer {
         ) as? BlockTransfer else {
             fatalError("Unable to create BlockTransfer object.")
         }
+        transfer.clientRestorationId = clientRestorationId
```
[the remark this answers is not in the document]
bsiegel: I kind of went back and forth on this. Block blob could easily read it from its parent. I'm not sure about multiblob. Maybe those don't need it at all because they delegate operations to their children?

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
```diff
@@ -56,6 +56,11 @@ public protocol Transfer: AnyObject {
     var state: TransferState { get }
     /// A debug representation of the transfer.
     var debugString: String { get }
+    /// An identifier used to associate this transfer with the client that created it. When a transfer is reloaded from
+    /// disk (e.g. after an application crash), it can only be resumed once a client with the same `restorationId` has
+    /// been initialized. Attempting to resume it without previously initializing such a client will cause the transfer
+    /// to transition to the `failed` state.
+    var clientRestorationId: String { get set }
```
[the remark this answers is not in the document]
bsiegel: I thought there could be a use case here for customers to look at a transfer and know what client started it. For example, maybe they would enumerate through all transfers, record the unique restoration IDs, and then only create those clients, before calling `resumeAll()`. It doesn't specifically need to be here though.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -39,30 +39,38 @@ public final class StorageBlobClient: PipelineClient {
...
+    ///     `restorationId` has been initialized. If your application only uses a single `StorageBlobClient`, it is
+    ///     recommended to use a value unique to your application (e.g. "MyApplication"). If your application uses
+    ///     multiple clients with different configurations, use a value unique to both your application and the
+    ///     configuration (e.g. "MyApplication.userClient").
     ///   - options: Options used to configure the client.
-    private init(baseUrl: URL, authPolicy: Authenticating, withOptions options: StorageBlobClientOptions? = nil) {
+    private init(
+        baseUrl: URL,
+        authPolicy: Authenticating,
+        withRestorationId restorationId: String,
```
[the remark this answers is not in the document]
bsiegel: I had a default value initially, but I decided to remove it. I like having a default for ergonomic reasons, but then I thought - what if the customer created a blob storage client using the default ID, but the customer happened to also use another third-party library and that library -also- created a storage client with the default ID. It would be easy for this to happen by accident, and it would cause major problems. So I went down the path of making the ID mandatory, but with API documentation that tries to explain what it's for and how to choose a good one.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
bsiegel: I considered that, but since I also have to always cast the key to NSString, I decided to encapsulate that into a function, `client(forRestorationId:)` instead. It seemed to keep the callsites pretty clean.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
bsiegel: I believe so, but I will test. Multiple blob clients will always use the single static TransferManager instance, with a single `NSPersistentContainer`. I believe this comment was originally referring to the plan of having multiple TransferManagers but letting them share the `NSPersistentContainer` so that we don't encounter CoreData concurrency problems. Now with a single TransferManager instance and a single `NSPersistentContainer` that should no longer be a problem.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+ResumableOperationQueueDelegate.swift
bsiegel: It does - this is because `StorageBlobClient` itself implements `TransferDelegate`, and it has a `delegate` property which is the actual delegate the customer supplies. The TransferManager calls the delegate method on the client, and the client in turn calls the delegate method on whatever the value of its `delegate` property is, if present. I missed committing the fixes to the demo app, but it's just a matter of updating the call sites that use the restorationID in the public API.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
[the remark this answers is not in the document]
tjprescott: Right, that's what I'm thinking. The multiblob operation will really just be firing off a bunch of blob operations, with the "MultiBlob" grouping being notified of changes.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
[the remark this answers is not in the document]
tjprescott: That makes sense. Other classes like OperationQueue and Logger require the same, so not too onerous.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: It could... depends on where and how things stateful things like the delegate get moved out of the Transfer Manager.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
```diff
@@ -33,7 +33,11 @@ extension BlobTransfer {
         return NSFetchRequest<BlobTransfer>(entityName: "BlobTransfer")
     }
 
-    @NSManaged internal var clientRestorationId: String
+    /// An identifier used to associate this transfer with the client that created it. When a transfer is reloaded from
+    /// disk (e.g. after an application crash), it can only be resumed once a client with the same `restorationId` has
+    /// been initialized. Attempting to resume it without previously initializing such a client will cause the transfer
+    /// to transition to the `failed` state.
+    @NSManaged public var clientRestorationId: String
```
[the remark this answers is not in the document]
bsiegel: I reverted the coredata changes, now only BlobTransfer has a real clientRestorationId. I still lean towards keeping it exposed in the public API - if the customer has cast / filtered to a BlobTransfer, I think it's a reasonable property to expose. However it no longer -needs- to be exposed, so if we want to keep it internal and consider promoting it later, that's a trivial change now.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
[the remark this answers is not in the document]
bsiegel: Handled this by computing the property from the parent in BlockTransfer, and from the first child in MultiBlobTransfer. There is an edge case, of a MultiBlob transfer that has no children. In that case, I am just returning a bogus value that should never match an existing client, which will just cause the MultiBlobTransfer to transition to "failed". I don't love this approach, but it works. Can you think of a better alternative?

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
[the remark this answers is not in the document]
bsiegel: Removed it from the protocol, but left public in BlobTransfer - as mentioned above, could be convinced to hide it.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
bsiegel: > The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.

I don't think I would call it that. As a naive developer, I might just do the following:

onApplicationStart:
- create Client A
- call resumeAll()

Then later in some other part of the application:
- create Client B
- call resumeAll()

Right now, resumeAll() resumes -all- transfers. Maybe we should make the client's *all methods only operate on transfers that match the client's restoration ID? That would make logical sense. But there would no longer be a handle to unconditionally pause/resume each and every transfer regardless of client. I'm going to put a little thought into this, because I can see use cases for both, and for the latter, I don't think it makes sense for those APIs to live on the client, since everything else the client does is scoped to that client.

bsiegel:
> The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.

I don't think I would call it that. As a naive developer, I might just do the following:

onApplicationStart:
- create Client A
- call resumeAll()

Then later in some other part of the application:
- create Client B
- call resumeAll()

Right now, resumeAll() resumes -all- transfers. Maybe we should make the client's *all methods only operate on transfers that match the client's restoration ID? That would make logical sense. But there would no longer be a handle to unconditionally pause/resume each and every transfer regardless of client. I'm going to put a little thought into this, because I can see use cases for both, and for the latter, I don't think it makes sense for those APIs to live on the client, since everything else the client does is scoped to that client.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: If you did the subscript wrapper could you not cast the String to NSString there?

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataProperties.swift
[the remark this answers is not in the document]
tjprescott: I'm okay leaving it public.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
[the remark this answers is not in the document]
tjprescott: Not really, but since MultiBlobTransfer is not yet implemented, it's probably fine. Realistically, it's nonsensical for a MultiBlobTransfer to contain no children. It must have at least one child (conceivably at least 2 or it's not really "multi").

tjprescott:
> > The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.
> 
> I don't think I would call it that. As a naive developer, I might just do the following:
> 
> onApplicationStart:
> 
> * create Client A
> * call resumeAll()
> 
> Then later in some other part of the application:
> 
> * create Client B
> * call resumeAll()
> 
> Right now, resumeAll() resumes -all- transfers. Maybe we should make the client's *all methods only operate on transfers that match the client's restoration ID? That would make logical sense. But there would no longer be a handle to unconditionally pause/resume each and every transfer regardless of client. I'm going to put a little thought into this, because I can see use cases for both, and for the latter, I don't think it makes sense for those APIs to live on the client, since everything else the client does is scoped to that client.

Actually, I would prefer what you suggest--that resumeAll resumes all transfers that match the restorationId of the client. You could make an optional parameter on those calls that would ignore restoration Id and defaults to false.

bsiegel:
> Actually, I would prefer what you suggest--that resumeAll resumes all transfers that match the restorationId of the client. You could make an optional parameter on those calls that would ignore restoration Id and defaults to false.

I had also talked myself into that, I'm working on it now. One thing came up that I'm not sure about: what should we do if the customer calls `client.pause(transfer: myTransfer)` but `myTransfer` has a different `restorationId`? I was thinking that maybe the correct thing to do in this situation is to throw an exception, but technically there is no reason why that call wouldn't work.

tjprescott:
> I had also talked myself into that, I'm working on it now. One thing came up that I'm not sure about: what should we do if the customer calls `client.pause(transfer: myTransfer)` but `myTransfer` has a different `restorationId`? I was thinking that maybe the correct thing to do in this situation is to throw an exception, but technically there is no reason why that call wouldn't work.

I'd be curious if it came up since most operations begin with resumeAll. I'd say keep it simple and let it succeed.

bsiegel:
> I'd be curious if it came up since most operations begin with resumeAll. I'd say keep it simple and let it succeed.

I went with this approach but I added an assertion. I feel like catching this in debug builds is reasonable, what do you think?

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
bsiegel: > The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.

I added an assert for this condition as suggested. I think that's the most pragmatic approach.

sdk/storage/AzureStorageBlob/Source/Data/BlockTransfer+CoreDataClass.swift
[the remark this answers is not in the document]
bsiegel: Agreed. Currently the `with` initializer doesn't require specifying any children. We can clean that up when we implement multiblob.

bsiegel:
> The question for me would be, is this a programming error? If so, perhaps an `assertionFailed` call so that way it crashes during debug development, but then mark the transfer failed and NOT crash during a production build.

I added an assert for this condition as suggested. I think that's the most pragmatic approach.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
bsiegel: I'm just going to poke at the sample app and see if I can reproduce a coredata issue. If you have a set of reliable repro steps, please do share :)

tjprescott: Pretty much, you start a bunch of parallel downloads and then jump to the upload page. It's almost like even if you try to always commit to the same context, if two threads try at the same time, rather than synchronize the operation, CoreData creates a new context and commits the change to that. With the parent and its child separated, CoreData validation fails when you try to save.

tjprescott:
@bsiegel this looks good to me. Feel free to merge when you are ready.

PR #249, opened 2020-04-08
Fix issues with local filesystem URLs.

### LINKED ISSUE #241: Download and Upload cannot rehydrate due to storing absolute URLs to local files
Trying to rehydrate downloads results in a message "You do not have permission to write to "blah" in the cache directory."  Trying to rehydrate an upload results in a "file not found error". This occurs because we are storing the absolute URL to the local file in CoreData instead of reconstructing the URL. The URL for the cache directory changes each time the app is run.

tjprescott:
Fixes issue with rehydrating transfers. Draft as I'm still investigating some CoreData issues that shouldn't be popping up.

Fixes #241.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
@@ -28,6 +28,7 @@ import AzureCore
 import AzureStorageBlob
 import MSAL
 import os.log
+import Photos
 
 import AVFoundation
 import AVKit
@@ -56,6 +57,7 @@ class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        PHPhotoLibrary.authorizationStatus()
         fetchData(self)
         blobClient?.startManaging()
     }
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -535,9 +535,11 @@ public final class StorageBlobClient: PipelineClient {
             } else if urlString.starts(with: tempDirPrefix) {
                 urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
                 return URL(string: urlString)!
-            } else {
-                return url
+            } else if urlString.starts(with: documentsDirPrefix) {
+                urlString = urlString.replacing(prefix: documentsDirPrefix, with: realDocumentsDir.absoluteString)
+                return URL(string: urlString)!
             }
+            return url
         }
 
         /// Retrieve a URL for a location on the local device in which to store a blob downloaded from a container.
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -512,10 +512,33 @@ public final class StorageBlobClient: PipelineClient {
     /// Helper containing properties and values to aid in constructing local paths for working with blobs.
     public struct PathHelper {
         /// The application's temporary directory.
-        public static let tempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+        public static let tempDir = URL(string: tempDirPrefix)!
+        private static let tempDirPrefix = "tempDir:/"
+        private static let realTempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
 
         /// The application's cache directory.
-        public static let cacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
+        public static let cacheDir = URL(string: cacheDirPrefix)!
+        private static let cacheDirPrefix = "cacheDir:/"
+        private static let realCacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
+
+        /// The application's documents directory.
+        public static let documentsDir = URL(string: documentsDirPrefix)!
+        private static let documentsDirPrefix = "documentsDir:/"
+        private static let realDocumentsDir = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+            .first!
+
+        internal static func absoluteUrl(forStorageRelativeUrl url: URL) -> URL {
+            var urlString = url.absoluteString
+            if urlString.starts(with: cacheDirPrefix) {
+                urlString = urlString.replacing(prefix: cacheDirPrefix, with: realCacheDir.absoluteString)
+                return URL(string: urlString)!
+            } else if urlString.starts(with: tempDirPrefix) {
+                urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
+                return URL(string: urlString)!
+            } else {
+                return url
+            }
+        }
 
         /// Retrieve a URL for a location on the local device in which to store a blob downloaded from a container.
         /// - Parameters:
--- a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
@@ -274,9 +274,6 @@ internal class BlobStreamUploader: BlobUploader {
         options: UploadBlobOptions? = nil
     ) throws {
         let manager = FileManager.default
-        guard manager.fileExists(atPath: source.path) else {
-            throw AzureError.fileSystem("File not found: \(source.path)")
-        }
         let attributes = try manager.attributesOfItem(atPath: source.path)
         guard let fileSize = attributes[FileAttributeKey.size] as? Int else {
             throw AzureError.fileSystem("Unable to determine file size: \(source.path)")
--- a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
@@ -283,7 +283,7 @@ internal class BlobStreamUploader: BlobUploader {
         self.client = client
         self.delegate = delegate
         self.options = options ?? UploadBlobOptions()
-        self.uploadSource = source
+        self.uploadSource = StorageBlobClient.PathHelper.absoluteUrl(forStorageRelativeUrl: source)
         self.uploadDestination = destination
         self.blobProperties = properties
         self.blockList = computeBlockList()
--- a/examples/AzureSDKDemoSwift/Source/BlobUploadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobUploadViewController.swift
@@ -139,6 +139,7 @@ class BlobUploadViewController: UIViewController, MSALInteractiveDelegate {
             }
         }
         group.notify(queue: .main) {
+            PHPhotoLibrary.authorizationStatus()
             self.blobClient?.startManaging()
             self.reloadCollectionView()
         }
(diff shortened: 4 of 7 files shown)


bsiegel:
This approach with the path schemes as placeholders makes sense to me. However I don't think we necessarily have to deal with changing CoreData to represent them as strings, and marshaling between the placeholder form and the real form in the source and destination properties. Even if they're not "real" URI schemes, they still form a valid URI so they can be parsed and stored just the same.

I prototyped an alternative where we just leave the placeholders there until we actually need to read/write to/from the URL, and made sure to resolve the real URL at a reasonably late stage right before the actual operation occurs - when constructing the BlobStreamUploader/Downloader. What do you think about the following?

```patch
diff --git a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
index 91d7a64..958e9f0 100644
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -512,10 +512,14 @@ public final class StorageBlobClient: PipelineClient {
     /// Helper containing properties and values to aid in constructing local paths for working with blobs.
     public struct PathHelper {
         /// The application's temporary directory.
-        public static let tempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+        public static let tempDir = URL(string: tempDirPrefix)!
+        private static let tempDirPrefix = "tempDir:/"
+        private static let realTempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
 
         /// The application's cache directory.
-        public static let cacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
+        public static let cacheDir = URL(string: cacheDirPrefix)!
+        private static let cacheDirPrefix = "cacheDir:/"
+        private static let realCacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
 
         /// Retrieve a URL for a location on the local device in which to store a blob downloaded from a container.
         /// - Parameters:
@@ -548,6 +552,19 @@ public final class StorageBlobClient: PipelineClient {
             let dirName = defaultUrlComps.joined(separator: "/")
             return (dirName, baseName)
         }
+
+        internal static func absoluteUrl(forStorageRelativeUrl url: URL) -> URL {
+            var urlString = url.absoluteString
+            if urlString.starts(with: cacheDirPrefix) {
+                urlString = urlString.replacing(prefix: cacheDirPrefix, with: realCacheDir.absoluteString)
+                return URL(string: urlString)!
+            } else if urlString.starts(with: tempDirPrefix) {
+                urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
+                return URL(string: urlString)!
+            } else {
+                return url
+            }
+        }
     }
 }
 
diff --git a/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift b/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
index a00fe59..3d7fd40 100644
--- a/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
@@ -308,7 +308,7 @@ internal class BlobStreamDownloader: BlobDownloader {
         self.delegate = delegate
         self.options = options ?? DownloadBlobOptions()
         self.downloadSource = source
-        self.downloadDestination = destination
+        self.downloadDestination = StorageBlobClient.PathHelper.absoluteUrl(forStorageRelativeUrl: destination)
         self.requestedSize = self.options.range?.length
         self.blobProperties = nil
         self.totalSize = -1
diff --git a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
index aa650bf..5bfb19e 100644
--- a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
@@ -286,7 +286,7 @@ internal class BlobStreamUploader: BlobUploader {
         self.client = client
         self.delegate = delegate
         self.options = options ?? UploadBlobOptions()
-        self.uploadSource = source
+        self.uploadSource = StorageBlobClient.PathHelper.absoluteUrl(forStorageRelativeUrl: source)
         self.uploadDestination = destination
         self.blobProperties = properties
         self.blockList = computeBlockList()

```

tjprescott:
> This approach with the path schemes as placeholders makes sense to me. However I don't think we necessarily have to deal with changing CoreData to represent them as strings, and marshaling between the placeholder form and the real form in the source and destination properties. Even if they're not "real" URI schemes, they still form a valid URI so they can be parsed and stored just the same.
> 
> I prototyped an alternative where we just leave the placeholders there until we actually need to read/write to/from the URL, and made sure to resolve the real URL at a reasonably late stage right before the actual operation occurs - when constructing the BlobStreamUploader/Downloader. What do you think about the following?
> 
> ```diff
> diff --git a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
> index 91d7a64..958e9f0 100644
> --- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
> +++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
> @@ -512,10 +512,14 @@ public final class StorageBlobClient: PipelineClient {
>      /// Helper containing properties and values to aid in constructing local paths for working with blobs.
>      public struct PathHelper {
>          /// The application's temporary directory.
> -        public static let tempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
> +        public static let tempDir = URL(string: tempDirPrefix)!
> +        private static let tempDirPrefix = "tempDir:/"
> +        private static let realTempDir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
>  
>          /// The application's cache directory.
> -        public static let cacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
> +        public static let cacheDir = URL(string: cacheDirPrefix)!
> +        private static let cacheDirPrefix = "cacheDir:/"
> +        private static let realCacheDir = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first!
>  
>          /// Retrieve a URL for a location on the local device in which to store a blob downloaded from a container.
>          /// - Parameters:
> @@ -548,6 +552,19 @@ public final class StorageBlobClient: PipelineClient {
>              let dirName = defaultUrlComps.joined(separator: "/")
>              return (dirName, baseName)
>          }
> +
> +        internal static func absoluteUrl(forStorageRelativeUrl url: URL) -> URL {
> +            var urlString = url.absoluteString
> +            if urlString.starts(with: cacheDirPrefix) {
> +                urlString = urlString.replacing(prefix: cacheDirPrefix, with: realCacheDir.absoluteString)
> +                return URL(string: urlString)!
> +            } else if urlString.starts(with: tempDirPrefix) {
> +                urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
> +                return URL(string: urlString)!
> +            } else {
> +                return url
> +            }
> +        }
>      }
>  }
>  
> diff --git a/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift b/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
> index a00fe59..3d7fd40 100644
> --- a/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
> +++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
> @@ -308,7 +308,7 @@ internal class BlobStreamDownloader: BlobDownloader {
>          self.delegate = delegate
>          self.options = options ?? DownloadBlobOptions()
>          self.downloadSource = source
> -        self.downloadDestination = destination
> +        self.downloadDestination = StorageBlobClient.PathHelper.absoluteUrl(forStorageRelativeUrl: destination)
>          self.requestedSize = self.options.range?.length
>          self.blobProperties = nil
>          self.totalSize = -1
> diff --git a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
> index aa650bf..5bfb19e 100644
> --- a/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
> +++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
> @@ -286,7 +286,7 @@ internal class BlobStreamUploader: BlobUploader {
>          self.client = client
>          self.delegate = delegate
>          self.options = options ?? UploadBlobOptions()
> -        self.uploadSource = source
> +        self.uploadSource = StorageBlobClient.PathHelper.absoluteUrl(forStorageRelativeUrl: source)
>          self.uploadDestination = destination
>          self.blobProperties = properties
>          self.blockList = computeBlockList()
> ```

Yes, I do like that approach!

tjprescott:
@bsiegel I pushed the changes you suggested and will track the CoreData issues as separate work. I also added the "documentsDirectory" as a PathHelper source.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -56,6 +57,7 @@ class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        PHPhotoLibrary.authorizationStatus()
```
tjprescott, line 60: This is sufficient to fix the permissions issues (assuming permission was previously granted). This was an issue with the client app, not the SDK.

sdk/storage/AzureStorageBlob/Source/StorageStreamUploader.swift
```diff
@@ -274,9 +274,6 @@ internal class BlobStreamUploader: BlobUploader {
         options: UploadBlobOptions? = nil
     ) throws {
         let manager = FileManager.default
-        guard manager.fileExists(atPath: source.path) else {
-            throw AzureError.fileSystem("File not found: \(source.path)")
-        }
```
tjprescott, line 279: Remove the code to swallow this error, as it masked the real problem (permissions).

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
bsiegel: Excellente~

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -512,10 +512,33 @@ public final class StorageBlobClient: PipelineClient {
...
+
+        internal static func absoluteUrl(forStorageRelativeUrl url: URL) -> URL {
+            var urlString = url.absoluteString
+            if urlString.starts(with: cacheDirPrefix) {
+                urlString = urlString.replacing(prefix: cacheDirPrefix, with: realCacheDir.absoluteString)
+                return URL(string: urlString)!
+            } else if urlString.starts(with: tempDirPrefix) {
+                urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
+                return URL(string: urlString)!
+            } else {
```
bsiegel, line 538: Need to add a case for `documentsDir` here :)

tjprescott: Fixed.

tjprescott:
@bsiegel should be good to go.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -535,9 +535,10 @@ public final class StorageBlobClient: PipelineClient {
             } else if urlString.starts(with: tempDirPrefix) {
                 urlString = urlString.replacing(prefix: tempDirPrefix, with: realTempDir.absoluteString)
                 return URL(string: urlString)!
-            } else {
-                return url
+            } else if urlString.starts(with: documentsDirPrefix) {
+                urlString = urlString.replacing(prefix: documentsDirPrefix, with: realDocumentsDir.absoluteString)
             }
+            return URL(string: urlString)!
```
bsiegel, line 541: Just for consistency: we should either return this on L540 (like we do in the other cases) and continue to just return `url` here, or refactor so that we have a single return here.

I ever so slightly prefer the first option, just so that we're not doing URL -> String -> URL again if the URL is unchanged.

PR #250, opened 2020-04-09
Remove cocoapods/carthage install instructions, and remove unused projects

bsiegel:
Some cleanup now that things are in master.


bsiegel:
Note to go through and remove CI configurations for the removed services.

bsiegel:
> As long as the removed services are retained on the dev branch, I think this is fine.

They're also retained in the commit history here :)

PR #256, opened 2020-04-09
Set max concurrency of StorageBlobClient.

### LINKED ISSUE #253: Allow configured max concurrency of Transfer Manager through StorageBlobClient.
Currently the max concurrency of the operation queue that backs the TransferManager is hard coded for 4 threads. This should be configurable through the BlobClient and is necessary for the UX study hypotheses.

tjprescott:
Allows setting client max concurrency through StorageBlobClientOptions. If not provided, will let the OS use the default value. Closes #253.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/Common.swift
+++ b/examples/AzureSDKDemoSwift/Source/Common.swift
@@ -99,7 +99,8 @@ struct AppState {
             )
             let options = StorageBlobClientOptions(
                 apiVersion: StorageBlobClient.ApiVersion.latest.rawValue,
-                logger: ClientLoggers.none
+                logger: ClientLoggers.none,
+                maxConcurrency: 2
             )
             client = StorageBlobClient(
                 accountUrl: AppConstants.storageAccountUrl,
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -39,6 +39,9 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     /// The maximum size of a single chunk in a blob upload or download.
     public let maxChunkSize: Int
 
+    /// The maximum number of threads to use for uploading and downloading. Omit to let the OS decide.
+    public let maxConcurrency: Int?
+
     /// Initialize a `StorageBlobClientOptions` structure.
     /// - Parameters:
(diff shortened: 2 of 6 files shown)


examples/AzureSDKDemoSwift/Source/Common.swift
```diff
@@ -99,7 +99,8 @@ struct AppState {
             )
             let options = StorageBlobClientOptions(
                 apiVersion: StorageBlobClient.ApiVersion.latest.rawValue,
-                logger: ClientLoggers.none
+                logger: ClientLoggers.none,
+                maxConcurrency: 2
```
tjprescott, line 103: The lower this is, the less we experience concurrency issues. Once we have those fixed, my preference would be for the default to simply be OS determined.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -199,9 +204,6 @@ public struct DownloadBlobOptions: AzureOptions, Codable, Equatable {
     /// a secure connection must be established to transfer the key.
     public let customerProvidedEncryptionKey: CustomerProvidedEncryptionKey?
 
-    /// The number of parallel connections with which to download.
-    public let maxConcurrency: Int?
```
tjprescott, line 203: Removed this option since we are not even attempting to honor it.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -288,9 +287,6 @@ public struct UploadBlobOptions: AzureOptions, Codable, Equatable {
     /// this value implies use of the default account encryption scope.
     public let customerProvidedEncryptionScope: String?
 
-    /// The number of parallel connections with which to upload.
-    public let maxConcurrency: Int?
```
tjprescott, line 292: Same here. Removed this option since we are not even attempting to honor it.

bsiegel:
I like the spirit of this PR, but we still have the problem that one client can change the concurrency level on the transfer manager, but all clients share a single manager, so this means that one client can change the behavior for all other clients.

One other issue (not problem per se) is that since this property is being added to the StorageBlobClientOptions structure, the customer can only ever set it once (per client), when initializing the client. They can't change it once it's been set.

tjprescott:
> In the long run, I think we want to do option 1, but option 2 (implemented in #255) might be better in the short run. I don't have a concern about making breaking API changes when we do get around to implementing option 1.

I'm good with that.

PR #258, opened 2020-04-10
Local urls proposal

bsiegel:
This was something I prototyped that allows the customer to use any well-known directory as the base for a local URL. Wanted to know what you thought.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -528,33 +528,46 @@ public final class StorageBlobClient: PipelineClient {
 
     /// Helper containing properties and values to aid in constructing local paths for working with blobs.
     public struct PathHelper {
-        /// The application's temporary directory.
-        public static let tempDir = URL(string: tempDirPrefix)!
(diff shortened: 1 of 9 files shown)


sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -549,7 +562,7 @@ public final class StorageBlobClient: PipelineClient {
         ///   - name: The name of the blob.
         ///   - container: The name of the container.
         public static func localUrl(
-            inDirectory directoryUrl: URL = cacheDir,
+            inDirectory directoryUrl: URL = url(for: .cachesDirectory),
```
tjprescott, line 565: Instead of us arbitrarily choosing a default, what if you configured the default? `BlobStorageClient.PathHelper.defaultDirectory = .foo`. That way if I want something other than cachesDirectory I don't have to manually include it with every call?

bsiegel: I started working on this, and ended up with a fairly different idea which I think improves ergonomics and correctness quite a bit. Please check out the most recent commit, and especially the full commit message.

PR #262, opened 2020-04-13
UX Study Fixes.

### LINKED ISSUE #261: [BUG] OperationQueue hangs when operations are paused
Set a low number for max concurrent operations. Then start a download, wait for the initial operation to complete and pause it by going to airplane mode. When you come out of airplane mode, rather than resuming, the transfer will appear as pending but will never continue. Resetting the app will permit the operations to continue.

This is occurring because there is a bug where canceled operations are not being marked as complete by the queue, and thus they are never returned to the pool.

### LINKED ISSUE #266: [BUG] Vague error when using `showAlert(error:)` method for a decoding error
**Describe the bug**
When attempting to use the Storage SDK with an invalid credential, the AuthetincationErrorDetail error is swallowed and ignored by the XMLParser, resulting in a JSON error that manifests to the customer as "The data couldn't be read because it isn't in the correct format."

***Exception or Stack Trace***
```
2020-04-14 01:20:42.344963+0430 AzureSDKDemoSwift[78885:9675246] [StorageClient] --> [7A8E4330-ADFF-4CE0-87C1-9AD12A87EED4]
2020-04-14 01:20:42.345232+0430 AzureSDKDemoSwift[78885:9675246] [StorageClient] GET https://azureiostorage.blob.core.windows.net/videoss?comp=list&resType=container&maxresults=20&sv=2019-02-02&ss=bfqt&srt=sco&sp=rwdlacup&se=2020-04-12T06:17:09Z&st=2020-04-11T22:17:09Z&spr=https&sig=REDACTED
2020-04-14 01:20:42.345426+0430 AzureSDKDemoSwift[78885:9675246] [StorageClient] --> [END 7A8E4330-ADFF-4CE0-87C1-9AD12A87EED4]
2020-04-14 01:20:43.869130+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] <-- [7A8E4330-ADFF-4CE0-87C1-9AD12A87EED4] (1523.43ms)
2020-04-14 01:20:43.869334+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] 403 Forbidden
2020-04-14 01:20:43.869540+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] <-- [END 7A8E4330-ADFF-4CE0-87C1-9AD12A87EED4]
2020-04-14 01:20:43.870196+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] No XML metadata found for Error. Ignoring.
2020-04-14 01:20:43.870494+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] No XML metadata found for Code. Ignoring.
2020-04-14 01:20:43.870822+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] No XML metadata found for Message. Ignoring.
2020-04-14 01:20:43.871094+0430 AzureSDKDemoSwift[78885:9683234] [StorageClient] No XML metadata found for AuthenticationErrorDetail. Ignoring.
```

**To Reproduce**

1. Create a blob client with a `StorageSASCredential` and an invalid SAS connection string.
2. Attempt to use the `listBlobs` operation to trigger a 403 forbidden error.

**Expected behavior**
The error message should be properly deserialized and surfaced to the user. 

ccc @maaniaa55

tjprescott:
- Fixes #261. 
- Improves the `showAlert(error:)` method to print the debugDescription, if available (fixes #266).
- Fixes an issue with the `AccessTier` enum that prevented it from being properly deserialized.

There were some holes in the Operation logic. I took the opportunity to simplify ResumableOperation and ResumableOperationQueue. I would like to further shrink these, but right now, it isn't necessary.

I believe the key problem was in the BlockOperation logic. After entering a dispatch group, there were return statements that did not have a matching `group.leave()` call. As a result, these threads were lost to the ether and never able to complete. So, when the the operation was resumed, all incomplete blocks were readded to the queue but could never proceed because the original threads were lost and could never complete their lifecycle.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
@@ -68,15 +68,12 @@ internal class BlockOperation: ResumableOperation {
                 options: downloader.options
             )
             chunkDownloader.download { result, httpResponse in
-                if !transfer.isActive {
-                    group.leave()
-                    return
-                }
+                defer { group.leave() }
+                if !transfer.isActive { return }
                 switch result {
                 case .success:
                     guard let responseHeaders = httpResponse?.headers else {
                         assertionFailure("Response headers not found.")
-                        group.leave()
                         return
                     }
                     let blobProperties = BlobProperties(from: responseHeaders)
@@ -94,7 +91,6 @@ internal class BlockOperation: ResumableOperation {
                     parent.error = err
                     self.notifyDelegate(withTransfer: transfer)
                 }
-                group.leave()
             }
         case .upload:
             guard let uploader = parent.uploader else {
@@ -112,10 +108,8 @@ internal class BlockOperation: ResumableOperation {
                 options: uploader.options
             )
             chunkUploader.upload { result, _ in
-                if !transfer.isActive {
-                    group.leave()
-                    return
-                }
+                defer { group.leave() }
+                if !transfer.isActive { return }
                 switch result {
                 case .success:
                     // Add block ID to the completed list and lookup where its final
@@ -136,7 +130,6 @@ internal class BlockOperation: ResumableOperation {
                     parent.error = err
                     self.notifyDelegate(withTransfer: transfer)
                 }
-                group.leave()
             }
         }
         group.wait()
--- a/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
@@ -32,7 +32,7 @@ internal class BlockOperation: ResumableOperation {
     // MARK: Initializers
 
     public convenience init(withTransfer transfer: BlockTransfer) {
-        self.init(state: transfer.state)
+        self.init(transfer: transfer, state: transfer.state)
         self.transfer = transfer
         transfer.operation = self
     }
@@ -45,17 +45,18 @@ internal class BlockOperation: ResumableOperation {
             return
         }
         let parent = transfer.parent
-        if isCancelled || isPaused { return }
+        if !transfer.isActive { return }
         transfer.state = .inProgress
         parent.state = .inProgress
         notifyDelegate(withTransfer: transfer)
-        if isCancelled || isPaused { return }
+        if !transfer.isActive { return }
         let group = DispatchGroup()
         group.enter()
         switch parent.transferType {
         case .download:
             guard let downloader = parent.downloader else {
                 assertionFailure("Downloader not found for BlobOperation.")
+                group.leave()
                 return
             }
             let chunkDownloader = ChunkDownloader(
@@ -67,11 +68,15 @@ internal class BlockOperation: ResumableOperation {
                 options: downloader.options
             )
             chunkDownloader.download { result, httpResponse in
-                if self.isCancelled || self.isPaused { return }
+                if !transfer.isActive {
+                    group.leave()
+                    return
+                }
                 switch result {
                 case .success:
                     guard let responseHeaders = httpResponse?.headers else {
                         assertionFailure("Response headers not found.")
+                        group.leave()
                         return
                     }
                     let blobProperties = BlobProperties(from: responseHeaders)
@@ -94,6 +99,7 @@ internal class BlockOperation: ResumableOperation {
         case .upload:
             guard let uploader = parent.uploader else {
                 assertionFailure("Uploader not found for BlobOperation.")
+                group.leave()
                 return
             }
             let chunkUploader = ChunkUploader(
@@ -106,7 +112,10 @@ internal class BlockOperation: ResumableOperation {
                 options: uploader.options
             )
             chunkUploader.upload { result, _ in
-                if self.isCancelled || self.isPaused { return }
+                if !transfer.isActive {
+                    group.leave()
+                    return
+                }
                 switch result {
                 case .success:
                     // Add block ID to the completed list and lookup where its final
--- a/sdk/storage/AzureStorageBlob/Source/Operation/MultiBlobOperation.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/MultiBlobOperation.swift
@@ -35,6 +35,7 @@ internal class MultiBlobOperation: ResumableOperation {
     // MARK: Initializers
 
     public init(withTransfer transfer: MultiBlobTransfer) {
+        super.init(transfer: transfer)
         self.parent = transfer
     }
 
@@ -46,13 +47,5 @@ internal class MultiBlobOperation: ResumableOperation {
         transfer.state = .complete
         delegate?.operation(self, didChangeState: transfer.state)
         super.main()
-    }
-
-    public override func cancel() {
-        super.cancel()
-    }
-
-    public override func pause() {
-        super.pause()
     }
 }
--- a/sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperation.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperation.swift
@@ -39,7 +39,7 @@ internal class ResumableOperation: Operation {
         return internalState
     }
 
-    public weak var transfer: TransferImpl?
+    public var transfer: TransferImpl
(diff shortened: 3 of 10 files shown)


sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
```diff
@@ -67,11 +68,15 @@ internal class BlockOperation: ResumableOperation {
                 options: downloader.options
             )
             chunkDownloader.download { result, httpResponse in
-                if self.isCancelled || self.isPaused { return }
+                if !transfer.isActive {
+                    group.leave()
```
tjprescott, line 72: I believe this was the root cause of the issue. All code paths after a `group.enter()` call must have a matching `group.leave()` or the lifecycle for the operation hangs, starving the queue of threads.

sdk/storage/AzureStorageBlob/Source/Operation/MultiBlobOperation.swift
```diff
@@ -47,12 +48,4 @@ internal class MultiBlobOperation: ResumableOperation {
         delegate?.operation(self, didChangeState: transfer.state)
         super.main()
     }
-
-    public override func cancel() {
```
tjprescott, line 51: Removed intentionally.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperation.swift
```diff
@@ -39,7 +39,7 @@ internal class ResumableOperation: Operation {
         return internalState
     }
 
-    public weak var transfer: TransferImpl?
```
tjprescott, line 42: A `ResumableOperation` (which I'm considering renaming TransferOperation) must have a reference to the transfer that spawned it, meaning it should not be a weak reference.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperation.swift
```diff
@@ -53,8 +53,9 @@ internal class ResumableOperation: Operation {
 
     // MARK: Initializers
 
-    public init(state: TransferState = .pending, delegate: ResumableOperationDelegate? = nil) {
+    public init(transfer: TransferImpl, state: TransferState = .pending, delegate: ResumableOperationDelegate? = nil) {
```
tjprescott, line 56: Necessary update to initializer since `transfer` is no longer optional.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperation.swift
```diff
@@ -63,6 +64,7 @@ internal class ResumableOperation: Operation {
     open override func main() {
         if isCancelled || isPaused { return }
         internalState = .complete
+        super.main()
```
tjprescott, line 67: Not sure if this was contributing to the problem, but best practice, this should call its superclass like the other methods do.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
```diff
@@ -34,13 +34,11 @@ internal protocol ResumableOperationQueueDelegate: ResumableOperationDelegate {
 internal class ResumableOperationQueue {
     // MARK: Properties
 
-    internal lazy var operations: [ResumableOperation] = []
```
tjprescott, line 37: Removes the "data store" aspect of the queue, making it much closer to just a regular operation queue. I originally thought this was needed to make operations that could pause/resume, but now the necessary info to facilitate that lives in the BlobTransfer and BlockTransfer objects, making this added complexity a source of confusion and bugs.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
```diff
@@ -34,13 +34,11 @@ internal protocol ResumableOperationQueueDelegate: ResumableOperationDelegate {
...
+    public lazy var operationQueue: OperationQueue = OperationQueue()
 
-    internal lazy var operationQueue: OperationQueue = OperationQueue()
+    weak var delegate: ResumableOperationQueueDelegate?
```
tjprescott, line 39: FYI, I want to get rid of ResumableOperationQueueDelegate entirely and let the operations directly call `transfer(didUpdate:...)` instead of an `operation(didUpdate:...)` that then signals the transfer. However, I thought that would be too much in this PR and wasn't necessary to fix the bug.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
```diff
@@ -105,44 +99,21 @@ internal class ResumableOperationQueue {
...
-        operations.remove(at: index)
-    }
-
-    public func pause(_ operation: ResumableOperation) {
+    func pause(_ operation: ResumableOperation) {
         operation.pause()
-        removeFromDataStore(operation)
-    }
-
-    public func remove(_ operation: ResumableOperation) {
```
tjprescott, line 134: With the removal of the "data source" aspect, all this did was call cancel, so I just removed it and updated references of `remove(...)` to `cancel(...)`

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
```diff
@@ -105,44 +99,21 @@ internal class ResumableOperationQueue {
...
-        removeFromDataStore(operation)
-
-        // Cancel the operation.
-        if !operation.isFinished {
-            operation.cancel()
-        }
-    }
-
-    public func replace(old oldOperation: ResumableOperation, withNew newOperation: ResumableOperation) {
```
tjprescott, line 143: No longer necessary. Pausing simply cancels the relevant operations. Resuming the paused transfer results in the necessary operations being requeued. The necessary state is tracked in the Transfer object.

tjprescott:
@bsiegel ping on this PR.

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
```diff
@@ -45,17 +45,18 @@ internal class BlockOperation: ResumableOperation {
...
         notifyDelegate(withTransfer: transfer)
-        if isCancelled || isPaused { return }
+        if !transfer.isActive { return }
         let group = DispatchGroup()
         group.enter()
         switch parent.transferType {
         case .download:
             guard let downloader = parent.downloader else {
                 assertionFailure("Downloader not found for BlobOperation.")
+                group.leave()
```
bsiegel, line 59: Would it be a good idea to instead follow every call to `group.enter()` with a `defer { group.leave() }` ? That way we don't have to make sure we call `leave()` at every return point in the method, and it will also catch if the method is terminated by throwing an exception rather than returning.

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
bsiegel: One thing I would like to clean up eventually for the public API is that transfers also have a `remove()` method which I would have a hard time explaining to a customer when they should be calling it. It feels like an implementation detail, maybe we can get rid of that for transfers in a future PR.

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
tjprescott: In theory it would be elegant and wonderful, but in practice, this results in an EXC_BAD_INSTRUCTION. It effectively does this:

```Swift
  group.enter()
  defer { group.leave() }
  {
    // Works great for early returns
  }
  group.wait()
  super.main()
  // fails here
}
```

tjprescott: You have to call `group.leave()` in the normal scenario, resulting in a double call to `leave()` or the leave won't logically get called until after the `wait()` and thus the application will hang.

tjprescott: I pushed a change that works. The key is to place the defer statement in the beginning of the callback closures. You can't just place one right after the intial enter call.

PR #264, opened 2020-04-13
Refactor ResumableOperation and ResumableOperationQueue

tjprescott:
Builds off of PR #262, which should be merged first.

- Removes `transfer(_:didUpdateWithState:)` since this is covered by the existing `transfer(_:didUpdateWithState:andProgress)` call.
- Adds a `transfersDidUpdate(_)` method for `TransferDelegate` which is analogous to the `operations(_:didUpdateWithState)` method of the now removed `ResumableOperationQueueDelegate`. Useful when a bunch of transfers change at once and the UI should respond.
- Renamed `ResumableOperation` to `TransferOperation` and shrunk the implementation.
- Renamed `ResumableOperationQueue` to `TransferOperationQueue` and greatly reduced the implementation, retaining only a couple helper methods.

The goal here is to reduce the layers of indirection in trying to maintain a pure, standalone ResumableOperationQueue to make diagnosing the CoreData issues easier.


sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
```diff
@@ -74,6 +74,7 @@ internal class BlockOperation: ResumableOperation {
                 case .success:
                     guard let responseHeaders = httpResponse?.headers else {
                         assertionFailure("Response headers not found.")
+                        group.leave()
```
bsiegel, line 77: Probably not needed now?

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -61,7 +61,7 @@ public final class StorageBlobClient: PipelineClient {
         "https://storage.azure.com/.default"
     ]
 
-    private static let manager: TransferManager = URLSessionTransferManager.shared
+    private static var manager: TransferManager = URLSessionTransferManager.shared
```
bsiegel, line 64: Any particular reason for this change?

sdk/storage/AzureStorageBlob/Source/Operation/ResumableOperationQueue.swift
```diff
@@ -26,62 +26,17 @@
...
-    var maxConcurrentOperationCount: Int {
-        get {
-            return operationQueue.maxConcurrentOperationCount
-        }
-
-        set {
-            operationQueue.maxConcurrentOperationCount = newValue
-        }
-    }
+internal class TransferOperationQueue: OperationQueue {
```
bsiegel, line 29: We should change this file name to match the type.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
```diff
@@ -50,17 +50,43 @@ extension URLSessionTransferManager: ResumableOperationQueueDelegate {
...
                 }
             } else {
                 delegate?.transfer(transfer, didUpdateWithState: state)
             }
         }
+        saveContext()
+    }
+
+    func transfersDidUpdate(_ transfers: [Transfer]) {
+        guard let restorationId = (transfers.first as? TransferImpl)?.clientRestorationId else { return }
```
bsiegel, line 69: There's an assumption here that all transfers in `transfers` come from the same client. Is that always true? If not we could refactor this to group transfers by their restoration ID, and then notify each client with the appropriate subset of `transfers`.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
```diff
@@ -50,17 +50,43 @@ extension URLSessionTransferManager: ResumableOperationQueueDelegate {
...
 
-    internal func operations(_: [ResumableOperation]?, didChangeState _: TransferState) {
+    func transfer(_ transfer: Transfer, didFailWithError error: Error) {
+        guard let restorationId = (transfer as? TransferImpl)?.clientRestorationId else { return }
+        guard let delegate = client(forRestorationId: restorationId) as? TransferDelegate else { return }
+        delegate.transfer(transfer, didFailWithError: error)
         saveContext()
     }
+
+    func client(forRestorationId restorationId: String) -> PipelineClient? {
```
bsiegel, line 89: This should probably move to `URLSessionTransferManager.swift` since its use is not exclusive to this protocol extension.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -166,51 +165,58 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
         let resumableOperations: [TransferState] = [.pending, .inProgress]
         guard !disallowed.contains(transfer.state) else { return }
-        var operations = [ResumableOperation]()
-
+        var operations = [TransferOperation]()
+        var pendingTransfers: [Transfer]
         switch transfer.transferType {
         case .download:
-            let pendingTransfers = transfer.transfers.filter { resumableOperations.contains($0.state) }
+            pendingTransfers = transfer.transfers.filter { resumableOperations.contains($0.state) }
```
bsiegel, line 181: It would be simpler to just filter by `$0.state.active`

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -166,51 +165,58 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
                 operations.append(initialOperation)
             }
         case .upload:
-            let finalOperation = BlobUploadFinalOperation(withTransfer: transfer, queue: operationQueue)
+            let finalOperation = BlobUploadFinalOperation(withTransfer: transfer, queue: operationQueue, delegate: self)
             operations.append(finalOperation)
-            let pendingTransfers = transfer.transfers.filter { resumableOperations.contains($0.state) }
-            for blockTransfer in pendingTransfers {
-                let blockOperation = BlockOperation(withTransfer: blockTransfer)
+            pendingTransfers = transfer.transfers.filter { resumableOperations.contains($0.state) }
```
bsiegel, line 210: Same here

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -456,7 +461,7 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         }
 
         // attempt to attach one
-        guard let client = client(forRestorationId: transfer.clientRestorationId) else {
+        guard let client = client(forRestorationId: transfer.clientRestorationId) as? StorageBlobClient else {
```
bsiegel, line 464: We can call `blobClient(forRestorationId:)` here instead.

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
tjprescott: Slipped through the rebase I see.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: It was to allow you to set the concurrent operations on the TM. I modified it so this isn't needed.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
tjprescott: Good point.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
tjprescott: I kept it here because it's part of the `TransferDelegate` protocol.

tjprescott:
@bsiegel comments addressed

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -39,9 +39,13 @@ public final class StorageBlobClient: PipelineClient {
     /// The global maximum number of managed transfers that will be executed concurrently by all `StorageBlobClient`
     /// instances. The default value is `maxConcurrentTransfersDefaultValue`. To allow this value to be determined
     /// dynamically based on current system conditions, set it to `maxConcurrentTransfersDynamicValue`.
-    public static var maxConcurrentTransfers: Int {
-        get { return manager.maxConcurrency }
-        set { manager.maxConcurrency = newValue }
+    public static var maxConcurrentTransfers: Int? {
```
bsiegel, line 42: We should never have a situation where this returns nil - we could just remove the type annotation and let `manager` be of type `URLSessionTransferManager` on L68.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
bsiegel: Oh, I missed that >< Good call.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I pushed 39d22b8 to address this. If you agree with the change, then feel free to merge :)

tjprescott: Yup. I dig it!

PR #271, opened 2020-04-17
Fix MD5 hashing and restore to true

### LINKED ISSUE #248: MD5 hashing enabled yields EXC_BAD_ACCESS
Workaround was to turn this feature off temporarily.

tjprescott:
Fixes #248.  I think the actual `EXC_BAD_ACCESS` was fixed by an earlier PR. This simply tweaks previously merged changes that caused turning the option on to yield a 400 error.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -47,7 +47,7 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     public init(
         apiVersion: String = StorageBlobClient.ApiVersion.latest.rawValue,
         logger: ClientLogger = ClientLoggers.default(tag: "StorageBlobClient"),
-        maxChunkSize: Int = 4 * 1024 * 1024
+        maxChunkSize: Int = 4 * 1024 * 1024 - 1
     ) {
         self.apiVersion = apiVersion
         self.logger = logger
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -44,6 +44,7 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     ///   - apiVersion: The API version of the Azure Storage Blob service to invoke.
     ///   - logger: The `ClientLogger` to be used by this `StorageBlobClient`.
     ///   - maxChunkSize: The maximum size of a single chunk in a blob upload or download.
+    ///     Must be less than 4MB if enabling MD5 or CRC64 hashing.
     public init(
         apiVersion: String = StorageBlobClient.ApiVersion.latest.rawValue,
         logger: ClientLogger = ClientLoggers.default(tag: "StorageBlobClient"),
--- a/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
@@ -160,9 +160,7 @@ internal class ChunkDownloader {
         let modifiedAccessConditions = options.modifiedAccessConditions
         let cpk = options.customerProvidedEncryptionKey
 
-        let rangeString = "bytes=\(startRange)-\(endRange)"
-        headers[.range] = rangeString
-        headers[.xmsRange] = rangeString
+        headers[.range] = "bytes=\(startRange)-\(endRange)"
         if let rangeGetContentMD5 = options.range?.calculateMD5 {
             headers[.rangeGetContentMD5] = String(rangeGetContentMD5)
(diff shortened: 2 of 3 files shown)


sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -47,7 +47,7 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     public init(
         apiVersion: String = StorageBlobClient.ApiVersion.latest.rawValue,
         logger: ClientLogger = ClientLoggers.default(tag: "StorageBlobClient"),
-        maxChunkSize: Int = 4 * 1024 * 1024
+        maxChunkSize: Int = 4 * 1024 * 1024 - 1
```
tjprescott, line 50: Above this value, you will get a 400 error.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
```diff
@@ -160,12 +160,16 @@ internal class ChunkDownloader {
...
         if let rangeGetContentMD5 = options.range?.calculateMD5 {
             headers[.rangeGetContentMD5] = String(rangeGetContentMD5)
+            headers[.range] = "bytes=\(startRange)-\(endRange)"
         }
         if let rangeGetContentCRC64 = options.range?.calculateCRC64 {
             headers[.rangeGetContentCRC64] = String(rangeGetContentCRC64)
+            headers[.range] = "bytes=\(startRange)-\(endRange)"
+        }
+        if headers[.range] == nil {
+            headers[.xmsRange] = "bytes=\(startRange)-\(endRange)"
```
tjprescott, line 172: To use the hashing functions you must use `.range`, not `.xmsRange` or you'll get a 400 error.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
bsiegel: Oh, we should definitely document that in the parameter comment. We should also validate this in the StorageBlobClient initializer and throw an exception (since those initializers are already marked as throwing).

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
bsiegel: Is it safe to just use `range` all the time? The API docs state:
> If the x-ms-range header is specified on a request, then the service uses the range specified by x-ms-range; otherwise, the range specified by the Range header is used.

tjprescott: Since you changed it to x-ms-range in your PR, I assumed there was some case for it so I kept that. But when requesting hashes, you _must_ use Range.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
tjprescott: I added the parameter comment, but adding the validation in the SDK was always an anti-pattern in CLI. If the Storage service were updated to allow hashes greater than this limit, we would be artificially restricting the user. The requirements to use hashing are clearly spelled out in the REST API docs:
https://docs.microsoft.com/en-us/rest/api/storageservices/get-blob#request-headers

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
bsiegel: Oh, no I just realized that we had `HTTPHeaders.range = "Range"` and `StorageHTTPHeaders.range = "x-ms-range"`, I simply renamed the enum value for the second to `.xmsRange` so that places where it was used didn't need to qualify the value.

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
bsiegel: OH, I see it's only a limitation for requests that use hashing. Yep, what you wrote makes perfect sense.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
tjprescott: I didn't see any issues when I set both headers to the same value, so I went with that. Let me know what you think.

bsiegel: It's a small thing so I don't think it matters, but I guess for simplicity's sake if we can just use Range, I'd prefer that. Probably just so that coming back to this code later, someone (maybe me :p) doesn't look at it and ask "okay why are we setting both headers to the same value??".

tjprescott: Updated.

PR #272, opened 2020-04-20
Update managed transfer progress reporting to use TransferProgress

### LINKED ISSUE #244: Update TransferManager progress to use TransferProgress rather than Float
`rawUpload` and `rawDownload` both use the TransferProgress struct to determine and report progress, but the TM operations simply return a Float based on the number of total blocks compared to the blocks completed. We can and should update the TM operations to use TransferProgress based on actual bytes transferred, not blocks as a proxy.

tjprescott:
Fix #244.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
@@ -47,8 +47,7 @@ public class BlobTransfer: NSManagedObject, TransferImpl {
         return Int64(transfers.filter { $0.state != .complete }.count)
     }
 
-    /// The current progress of the transfer, calculated as the number of bytes transferred divided by the total number
-    /// of bytes that comprise the blob transfer.
+    /// The current progress of the transfer, based on bytes transferred.
     public var progress: TransferProgress {
         return TransferProgress(bytes: Int(bytesTransferred), totalBytes: Int(totalBytesToTransfer))
     }
--- a/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
@@ -47,11 +47,10 @@ public class BlobTransfer: NSManagedObject, TransferImpl {
         return Int64(transfers.filter { $0.state != .complete }.count)
     }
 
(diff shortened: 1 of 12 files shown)


sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
```diff
@@ -47,11 +47,10 @@ public class BlobTransfer: NSManagedObject, TransferImpl {
         return Int64(transfers.filter { $0.state != .complete }.count)
     }
 
-    // TODO: Convert this to return a TransferProgress object
-    /// The current progress of the transfer, calculated as the number of completed blocks divided by the total number
-    /// of blocks that comprise the blob transfer.
-    public var progress: Float {
-        return Float(Int64(transfers.count) - incompleteBlocks) / Float(transfers.count)
+    /// The current progress of the transfer, calculated as the number of bytes transferred divided by the total number
```
bsiegel, line 50: Probably need to update this comment since the value isn't computed anymore - TransferProgress contains both values.

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
```diff
@@ -42,6 +42,9 @@ public struct TransferProgress {
 
     /// Percentage of the transfer that is complete, as a Float between 0 and 1.
     public var asFloat: Float {
+        // Returning NaN sometimes results in progress appearing as 100% when it shouldn't.
+        // This ensures progress will report 0%.
+        if totalBytes <= 0 { return 0 }
```
bsiegel, line 47: Should we also handle the situation where bytes > totalBytes (IDK what happened here, maybe just return 1.0?) and bytes == totalBytes (explicitly return 1.0 here to avoid potential float division weirdness?)

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
tjprescott: Tested for the 1.0 case and included an assertion for the abnormal case so it will crash during development but not production. In production it would simply return 1.x.

PR #273, opened 2020-04-21
Serialize logging

### LINKED ISSUE #268: LoggingPolicy messages are interleaved
Logging messages are interleaved when there are multiple threads. Sometimes the messages even get interrupted in the middle of logging a log line.

We probably need to run the logger on its own thread?

bsiegel:
Fixes #268

### DIFF
--- a/sdk/core/AzureCore/Source/ClientLogger.swift
+++ b/sdk/core/AzureCore/Source/ClientLogger.swift
@@ -73,11 +73,11 @@ public struct ClientLoggers {
 
     // MARK: Static Methods
 
-    public static func `default`(tag: String) -> ClientLogger {
+    public static func `default`(tag: String, level: ClientLogLevel = .info) -> ClientLogger {
         if #available(macOS 10.12, iOS 10.0, watchOS 3.0, tvOS 10.0, *) {
-            return OSLogger(category: tag)
+            return OSLogger(category: tag, level: level)
         } else {
-            return NSLogger(tag: tag)
+            return NSLogger(tag: tag, level: level)
         }
     }
 }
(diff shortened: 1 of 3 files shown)


sdk/core/AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -72,9 +76,24 @@ public class LoggingPolicy: PipelineStage {
     // MARK: PipelineStage Methods
 
     public func on(request: PipelineRequest, then completion: @escaping OnRequestCompletionHandler) {
-        var returnRequest = request.copy()
```
bsiegel, line 75: In testing with the demo app, it didn't seem like this was necessary - just updating the context on the provided `request` worked, the response logger was able to get the request start time back out of the context just fine.

sdk/core/AzureCore/Source/ClientLogger.swift
```diff
@@ -156,15 +156,15 @@ public class OSLogger: ClientLogger {
...
         self.osLogger = osLogger
     }
 
     public init(
-        subsystem: String = "com.azure",
-        category: String,
-        level: ClientLogLevel = .info
+        forSubsystem subsystem: String = "com.azure",
+        withCategory category: String,
+        atLevel level: ClientLogLevel = .info
```
tjprescott, line 167: Any specific reason for this change?

bsiegel: No, just that this was the first swift code I wrote, and I probably didn't understand the whole internal vs external name thing. I think this is what I would have called it if I'd had the benefit of hindsight that I do now.

tjprescott: It just seems that the old style was more consistent with the internal implementation classes:
```Swift
OSLog(subsystem: subsystem, category: category)
NSLogger(tag: "MyAppClient", level: .debug)
```

bsiegel: I don't have a strong opinion, I guess the underlying logger APIs don't really feel particularly "swifty" either. But maybe you're right, better to be consistent with the existing Foundation classes even if they feel less idiomatic. I'll remove that change.

PR #277, opened 2020-04-23
Fix progress bar regressions

tjprescott:
Fixes issues introduced by #272:
- Completed downloads only showing a progress bar to 90-ish % complete due to the initial block not being counted. 
- off-by-one errors in calculating total bytes
- issue where reconnected uploaders/downloaders would have incorrect progress information

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
@@ -44,9 +44,7 @@ public struct TransferProgress {
     public var asFloat: Float {
         // Returning NaN sometimes results in progress appearing as 100% when it shouldn't.
         // This ensures progress will report 0%.
-        assert(bytes <= totalBytes, "Transferred bytes unexpectedly > than total bytes.")
         if totalBytes <= 0 { return 0 }
-        if bytes == totalBytes { return 1.0 }
         return Float(bytes) / Float(totalBytes)
     }
 }
--- a/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
@@ -44,7 +44,10 @@ public struct TransferProgress {
     public var asFloat: Float {
         // Returning NaN sometimes results in progress appearing as 100% when it shouldn't.
         // This ensures progress will report 0%.
+        // TODO: Debug concurrency instabilitites triggering this
+        // assert(bytes <= totalBytes, "Transferred bytes unexpectedly > than total bytes.")
         if totalBytes <= 0 { return 0 }
+        if bytes == totalBytes { return 1.0 }
         return Float(bytes) / Float(totalBytes)
     }
 }
--- a/sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
@@ -96,10 +96,16 @@ internal class BlobDownloadInitialOperation: TransferOperation {
 
         let group = DispatchGroup()
         group.enter()
-        downloader.initialRequest { result, _ in
+        downloader.initialRequest { result, httpResponse in
             switch result {
             case .success:
+                guard let responseHeaders = httpResponse?.headers else {
+                    assertionFailure("Response headers not found.")
+                    return
+                }
+                let blobProperties = BlobProperties(from: responseHeaders)
                 parent.totalBytesToTransfer = Int64(downloader.totalSize)
+                parent.bytesTransferred += Int64(blobProperties.contentLength ?? 0)
                 transfer.state = .complete
                 self.queueRemainingBlocks(forTransfer: parent)
                 self.notifyDelegate(withTransfer: parent)
--- a/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
@@ -77,7 +77,7 @@ internal class BlockOperation: TransferOperation {
                         return
                     }
                     let blobProperties = BlobProperties(from: responseHeaders)
-                    let bytesTransferred = blobProperties.contentLength ?? 0
+                    let bytesTransferred = (blobProperties.contentLength ?? 1) - 1
                     downloader.progress += bytesTransferred
                     parent.bytesTransferred += Int64(bytesTransferred)
                     transfer.state = .complete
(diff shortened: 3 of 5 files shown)


sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
```diff
@@ -44,7 +44,8 @@ public struct TransferProgress {
     public var asFloat: Float {
         // Returning NaN sometimes results in progress appearing as 100% when it shouldn't.
         // This ensures progress will report 0%.
-        assert(bytes <= totalBytes, "Transferred bytes unexpectedly > than total bytes.")
+        // TODO: Debug concurrency instabilitites triggering this
+        // assert(bytes <= totalBytes, "Transferred bytes unexpectedly > than total bytes.")
```
tjprescott, line 48: During sunny day scenarios (both with and without restart) the final bytes will equal the total expected bytes. However, core data instabilities can result in this assert being triggered, which would result in the app continuously crashing.

sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
```diff
@@ -96,10 +96,16 @@ internal class BlobDownloadInitialOperation: TransferOperation {
...
+        downloader.initialRequest { result, httpResponse in
             switch result {
             case .success:
+                guard let responseHeaders = httpResponse?.headers else {
+                    assertionFailure("Response headers not found.")
+                    return
+                }
+                let blobProperties = BlobProperties(from: responseHeaders)
                 parent.totalBytesToTransfer = Int64(downloader.totalSize)
+                parent.bytesTransferred += Int64(blobProperties.contentLength ?? 0)
```
tjprescott, line 108: Would result in the initial block not being "counted", resulting in "completed" blobs only going to 90+%.

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
```diff
@@ -77,7 +77,7 @@ internal class BlockOperation: TransferOperation {
                         return
                     }
                     let blobProperties = BlobProperties(from: responseHeaders)
-                    let bytesTransferred = blobProperties.contentLength ?? 0
+                    let bytesTransferred = (blobProperties.contentLength ?? 1) - 1
```
tjprescott, line 80: Off-by-one error here would result in the assertion always being triggered.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
```diff
@@ -430,7 +430,7 @@ internal class BlobStreamDownloader: BlobDownloader {
                     // extract the file size from the content range. If a specific size request was
                     // not made, then the request is for the entire file.
                     let blobSize = try self.parseLength(fromContentRange: contentRange)
-                    self.totalSize = blobSize
+                    self.totalSize = blobSize - 1
```
tjprescott, line 433: Off-by-one error here would result in the final bytes always being just short of the total.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -496,6 +496,7 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
                     properties: blobProperties,
                     options: uploadOptions
                 )
+                transfer.uploader?.progress = Int(transfer.bytesTransferred)
```
tjprescott, line 499: Total bytes to transfer for upload is determined by the file size, so we can't set it (and don't need to).

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -507,6 +508,8 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
                     destination: destination,
                     options: transfer.downloadOptions
                 )
+                transfer.downloader?.progress = Int(transfer.bytesTransferred)
+                transfer.downloader?.totalSize = Int(transfer.totalBytesToTransfer)
```
tjprescott, line 512: This was resulting in bizarre progress on app restart because the downloader progress wasn't being updated from the transfer.

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
bsiegel: That's pretty bizarre, but that's par for the course for me for CoreData :/

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
bsiegel: Do you understand why this is the case? I would think if we are summing up the content-length of all the blocks, it should exactly equal the total bytes.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
bsiegel: Again I don't fully understand why this is the case :/

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
tjprescott: A difference between the range being inclusive or exclusive.

tjprescott: After correcting the off-by-one errors, the sum of bytes does equal the total, which is how I knew it was cumulative off-by-one errors rather than something weird from the service.

sdk/storage/AzureStorageBlob/Source/StorageStreamDownloader.swift
tjprescott: Exclusive range vs. inclusive range. start 1 end 100  Is that 100 bytes or 99? Depends if you count 100 or not.

bsiegel: Makes sense. We should do the -1 operation inside of parseLength though, so if it gets used elsewhere later we won't have to apply the same adjustment.

sdk/storage/AzureStorageBlob/Source/Operation/BlockOperations.swift
bsiegel: This one still puzzles me, because I would think that regardless of whether we are calculating the range wrong, the content-length returned from the server should always 100% match the exact number of bytes it's returning. Now maybe we are requesting one too many bytes when we split things into chunks (because of Range inclusive vs exclusive), and that's causing each result to provide one too many bytes back?

tjprescott:
There is still some odd behavior I can't trace and will cause problems, so I will close this PR, revert that change, and start a new PR with this and the previous changes.

PR #279, opened 2020-04-23
Revert "Update managed transfer progress reporting to use TransferPro…

tjprescott:
Revert these changes. Trying to use TransferProgress has more unintended consequences that shouldn't derail the UX study.


bsiegel:
@tjprescott the build is failing, looks like due to the change to the transfer(:didUpdateWithState:)

tjprescott:
@bsiegel build fixed and passing. I have tested to verify it does revert the anomalous behavior.

PR #283, opened 2020-04-24
Fix BlobTransfer local URLs are unresolved

### LINKED ISSUE #281: UX Study. Awkwardness working with LocalURL
Identified during the UX Study.

When we tried to play a video using `transfer.destination` on a fully downloaded BlobTransfer object, the resulting URL failed to play. We had to do the following, which we would like to not have to do:

```Swift
let url = LocalUrl(fromAbsoluteUrl: transfer.destination)!.resolvedUrl!
```

bsiegel:
Fixes #281 

The source and destination URLs on a blob transfer are the raw values
    that get stored in CoreData. They aren't appropriate for a user to
    consume directly, so those have been hidden and instead we add the
    sourceUrl and destinationUrl properties that correctly resolve the value
    if it represents a LocalURL.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
@@ -36,7 +36,6 @@ import UIKit
 
 class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
     private var dataSource: PagedCollection<BlobItem>?
-    private var downloadMap = [IndexPath: BlobTransfer]()
     private var noMoreData = false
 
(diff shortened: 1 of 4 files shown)


tjprescott:
@mitchdenny any idea why this PR is stalled?

tjprescott:
@bsiegel not sure why the check-enforcer won't run. Maybe if you rebase your PR off the latest in master?

mitchdenny:
Check Enforcer would have just needed a kick in the pants.

mitchdenny:
Because this PR probably pre-dated when I enabled it in the repo.

PR #284, opened 2020-04-24
Fix single block blob download

### LINKED ISSUE #280: [BUG] UX Study. Download transfers that complete with only one block do not automatically switch to .complete
Identified in the UX study. If a download requires only 1 block (< 4MB) then the state does not trigger identifying the transfer as complete.

tjprescott:
Fixes #280.

### DIFF
--- a/sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
@@ -53,7 +53,10 @@ internal class BlobDownloadInitialOperation: TransferOperation {
         }
 
         // early out if no more blocks are necesasry
-        guard downloader.blockList.count > 0 else { return }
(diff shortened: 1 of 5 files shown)


sdk/storage/AzureStorageBlob/Source/Operation/BlobDownloadOperations.swift
```diff
@@ -53,7 +53,10 @@ internal class BlobDownloadInitialOperation: TransferOperation {
         }
 
         // early out if no more blocks are necesasry
-        guard downloader.blockList.count > 0 else { return }
+        guard downloader.blockList.count > 0 else {
+            transfer.state = .complete
```
tjprescott, line 57: This was the cause of the problem.

bsiegel: Well that was easy. 👍

PR #285, opened 2020-04-24
Perform callbacks on DispatchQueue.main

tjprescott:
- Simplifies client code by having the SDK invoke all completion handlers and delegate callbacks on the main DispatchQueue. 
- Restore the recommended internal labels for the `TransferDelegate` protocol based on UX study feedback.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/CustomTabBarController.swift
+++ b/examples/AzureSDKDemoSwift/Source/CustomTabBarController.swift
@@ -30,10 +30,6 @@ import UIKit
 class CustomTabBarController: UITabBarController {
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
-        do {
-            try StorageBlobClient.stopManaging()
-        } catch {
-            showAlert(error: error)
-        }
+        StorageBlobClient.stopManaging()
     }
 }
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -620,22 +620,22 @@ extension StorageBlobClient: TransferDelegate {
         didUpdateWithState state: TransferState,
         andProgress progress: Float?
     ) {
-        delegate.transfer(transfer, didUpdateWithState: state, andProgress: progress)
+        transferDelegate?.transfer(transfer, didUpdateWithState: state, andProgress: progress)
     }
 
     /// :nodoc:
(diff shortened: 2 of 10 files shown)


sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -595,22 +620,34 @@ extension StorageBlobClient: TransferDelegate {
         didUpdateWithState state: TransferState,
         andProgress progress: Float?
     ) {
-        transferDelegate?.transfer(transfer, didUpdateWithState: state, andProgress: progress)
+        guard let delegate = transferDelegate else { return }
```
bsiegel, line 623: Since these methods just proxy the transferDelegate message coming in from `URLSessionTransferManager`, it might be better to put these changes there instead.

tjprescott: Good point. I'll update.

tjprescott: I did the update and realize that means I don't actually need to wrap calls in the operation classes since they will funnel through these methods.

sdk/storage/AzureStorageBlob/Source/TransferManager.swift
```diff
@@ -63,13 +63,13 @@ internal protocol TransferManager {
 /// A delegate to receive notifications about state changes for all transfers managed by a `StorageBlobClient`.
 public protocol TransferDelegate: AnyObject {
     /// A transfer's state has changed, and progress may be reported.
-    func transfer(_: Transfer, didUpdateWithState: TransferState, andProgress: Float?)
```
tjprescott, line 66: This provides the recommended internal labels when using XCode's "fix-it" function to stub out protocol methods.

examples/AzureSDKDemoSwift/Source/CustomTabBarController.swift
```diff
@@ -30,10 +30,6 @@ import UIKit
 class CustomTabBarController: UITabBarController {
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
-        do {
-            try StorageBlobClient.stopManaging()
-        } catch {
-            showAlert(error: error)
-        }
+        StorageBlobClient.stopManaging()
```
bsiegel, line 33: I think there's one more I had found, in AppDelegate.swift there's a call to `try? StorageBlobClient.stopManaging()` that can have the try removed.

PR #300, opened 2020-05-02
Auto start transfer manager

bsiegel:
Based on #298.

This PR adds a new state, `authenticationRequired`, to which transfers will transition if the request is attempted but is subsequently aborted by the AuthenticationPolicy because interactive authentication would be required.

If a transfer in the `authenticationRequired` state is then resumed by the developer, a value is added to the request context instructing the AuthenticationPolicy to proceed and perform the interactive authentication workflow.

Transfers which are started explicitly by the developer by calling the client's `upload` and `download` methods will have this value added to their requests automatically - they will still invoke the interactive auth workflow if needed.

Conversely, transfers that are being restored from disk and need to have a new client reconnected will not have this value added to their requests and thus will updated to the `authenticationRequired` state, requiring the developer to explicitly resume them in order for the interactive auth workflow to be invoked.


bsiegel:
Closing in favor of alternative approach that will have transfers paused when the app restarts, and the customer will explicitly resume them.

We also discussed having a way for the customer to toggle "auto-starting" transfers so that transfers can be queued up and then started explicitly at some later time.

Issue #289, opened 2020-04-29
Make Client Restoration ID Optional with Default Value


tjprescott:
While this may cause problems in certain scenarios where our library is packaged with or used by another library, the most common intended case is that the client app is invoking the library for one type of client, in which case defaulting this value makes sense and reduces cognitive friction for the developer.

bsiegel:
Addressed in #306

PR #306, opened 2020-05-15
Updates based on review of design guidelines

bsiegel:
This PR makes changes to align with the revised design guidelines in Azure/azure-sdk#1340.

This PR doesn't include two things required by the guidelines which will generally be complicated, and should thus be broken out into their own PRs:
- Refactor `PagedCollection` to match the guideline requirements / API for paging. #326
- Add the ability for in-flight requests to be canceled, return an `AzureTask` from the pipeline that permits cancellation of the request, and make all client methods return an `AzureTask`. #325

### DIFF
--- a/sdk/core/AzureCore/Source/DataStructures/Collections.swift
+++ b/sdk/core/AzureCore/Source/DataStructures/Collections.swift
@@ -168,7 +168,7 @@ public class PagedCollection<SingleElement: Codable> {
     // MARK: Public Methods
 
     /// Retrieves the next page of results asynchronously.
-    public func nextPage(then completion: @escaping Continuation<Element>) {
+    public func nextPage(completionHandler: @escaping Continuation<Element>) {
         // exit if there is no valid continuation token
         guard let continuationToken = continuationToken,
             continuationToken != "" else {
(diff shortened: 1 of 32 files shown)


bsiegel:
Still a few things to be done here - need to add AzureTask and make async methods return them. Also make Transfer type inherit from it. Maybe some other things, I'll have to re-read the guidelines now that they're merged and see what we're missing.

bsiegel:
930a218 resolves #289

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
```diff
@@ -49,7 +49,7 @@ public struct TransferProgress {
 // MARK: Protocols
 
 /// Object that contains information about a transfer operation.
-public protocol Transfer: AnyObject {
+public protocol Transfer: AzureTask, AnyObject {
```
tjprescott, line 52: I think, per our guidelines, even our non-transfer calls like `listBlobs` need to return an AzureTask so they can be canceled... which, I guess, in this context just means we would ignore any results that would come back if the cancel was issued during the networking transit time.

sdk/core/AzureCore/Source/DataStructures/Collections.swift
```diff
@@ -180,15 +176,9 @@ public class PagedCollection<SingleElement: Codable>: PagedCollectionDelegate {
...
-        let delegate = self.delegate ?? self
-
         client.logger.info(String(format: "Fetching next page with: %@", continuationToken))
-        var queryParams = [QueryParameter]()
-        guard let url = delegate.continuationUrl(
-            continuationToken: continuationToken,
-            queryParams: &queryParams,
-            requestUrl: requestUrl
-        ) else { return }
+        guard let url = client.continuationUrl(forRequestUrl: requestUrl, withContinuationToken: continuationToken)
```
tjprescott, line 180: Based on the discussion with TransferDelegate vs. closures, should we expose a closure option here as an option for those who would not want to implement what is essentially a delegate method?

sdk/storage/AzureStorageBlob/Source/Data/Transfer.swift
bsiegel: Yes, that's totally correct. However I started working on implementing that and it became a pretty large refactor so I broke it out into issue #325. Perhaps when working on that issue we can find a way to somehow cancel in-flight network requests.

sdk/core/AzureCore/Source/DataStructures/Collections.swift
bsiegel: I don't particularly think so, since the protocol is internal rather than something the end consumer will be having to code up. So it really only affects library authors, and personally I think the protocol method is more intention-revealing than the closure. Plus if your client has no paginated APIs you can just not conform to `PageableClient` at all.

PR #315, opened 2020-05-19
PagedCollection updates

### LINKED ISSUE #303: PagedCollection: Property to detected when list is exhausted.
Comment from UX Study.
- [x] Expose property to detected when list is exhausted

tjprescott:
This PR:

- Fixes #303 by adding an `isExhausted` property the app developer can check.
- Wraps completion handler calls in `nextPage` and `nextItem` in appropriate calls to dispatch onto the main queue to eliminate the need for the developer to do this.
- The default behavior for "no data received" is to simply return rather than call the completion handler with `nil` data. This allows the user's app to do the right thing when the collection is exhausted without them having to manually check `isExhausted`. The property is still supplied as the app developer may want to affirmatively check this.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
@@ -86,10 +86,8 @@ class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
     private func loadMoreSettings() {
         dataSource?.nextPage { result in
             switch result {
-            case let .success(data):
-                if data != nil {
-                    self.tableView.reloadData()
-                }
+            case .success:
+                self.tableView.reloadData()
             case let .failure(error):
                 self.showAlert(error: error)
             }
--- a/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
@@ -36,7 +36,6 @@ import UIKit
 
 class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
     private var dataSource: PagedCollection<BlobItem>?
-    private var noMoreData = false
 
     @IBOutlet var tableView: UITableView!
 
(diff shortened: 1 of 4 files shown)


tjprescott:
@bsiegel ping :)

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -79,25 +78,20 @@ class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
...
         }
     }
 
     /// Uses asynchronous "nextPage" method to fetch the next page of results and update the table view.
     private func loadMoreSettings() {
-        guard noMoreData == false else { return }
         dataSource?.nextPage { result in
             switch result {
             case let .success(data):
                 if data != nil {
```
bsiegel, line 90: If I read the commentary right, this condition should now always be true right?

bsiegel: If that's accurate, should we change the Result type for nextPage / nextItem to have a non-nullable element in the success case?

tjprescott: Good point. I've updated those return types and adjusted accordingly. Now the client code doesn't need to worry about the "got nothing" scenario at all!

bsiegel: Awesome, yeah that definitely simplifies things. Probably just need to update the demo app here to match that and then this is 👍 to merge.

tjprescott:
@bsiegel I updated the demo app accordingly, and added a second commit that fixes a small bug in the filtering methods. Please make sure you are okay with the fix.

sdk/storage/AzureStorageBlob/Source/DataStructures/TransferCollection.swift
```diff
@@ -96,12 +96,14 @@ public struct TransferCollection {
...
-        blobName _: String? = nil,
-        localUrl _: LocalURL? = nil,
-        state _: TransferState? = nil
+        containerName: String? = nil,
+        blobName: String? = nil,
+        localUrl: LocalURL? = nil,
+        state: TransferState? = nil
     ) -> BlobTransfer? {
-        return items.first { match($0) }
+        return items.first { transfer in
```
bsiegel, line 104: Good catch! That's sort of the one thing I didn't like about making all of the matching arguments optional, is that the autocomplete suggests `transfers.firstWith()` as an option which obviously makes no sense, but there's no way to encode "all of these arguments are optional but at least one must be specified".

That said, I made the same mistake above in `filterWhere` - mind fixing it up there as well? Thanks <3

tjprescott: Done 👍

Issue #313, opened 2020-05-18
UX Study Discussion #5: Starting and Stopping TransferManager


tjprescott:
One question we wanted to answer was whether users felt it was appropriate to need to start and stop the TransferManager engine or whether they would expect that this could just happen automagically. No participants insisted that it should start automatically, though some said that in many simple cases it would likely be appropriate. 

If we do feel that users should start the TransferManager manually in their code, the current behavior is that if `upload` or `download` are called before `startManaging` is called, those transfers will sit in `.pending` status and never execute, which may violate the principle of least surprise. We should consider throwing an error or at least emitting a warning in these instances. 

If we want to pursue automatic starting of the TransferManager, we need to consider the situation where a UI interaction may be needed to refresh an expired credential, such as with MSAL authentication. One possible way would be to have delegate methods on the TransferManager or StorageBlobClient to facilitate this.

tjprescott:
@bsiegel what was the decision we came to with regard to your closed PR? We didn't auto-start, but we did agree that something should happen.

tjprescott:
Per discussion on 6/24, we agreed to try and eliminate these methods as part of PR #327.

PR #318, opened 2020-05-22
Progress handlers and StorageBlobClientDelegate

### LINKED ISSUE #308: UX Study Discussion #3: Resolve callback mechanisms (TransferDelegate vs. closures)
We wanted to know whether people preferred delegates or callbacks. No one was surprised by the use of delegates. One person expressed a strong preference for callbacks. The first question is, can we do both? The answer is probably yes. The second question is, should we do both? 

An exploration into Swift UI may help inform this answer. 

The participant who really wanted to use callback did so in part because it linked the cell to the transfer at the location in code that made the most sense to him. This would (theoretically) eliminate the problem of having to look up in-progress transfers (discussion #2) and correlate them to a cell. 

With the delegate itself, there was a complaint that the delegate does not offer “do-nothing” default implementations. The participant would have preferred to have default implementations in place and only have to implement the specific method they need. When asked how they would discover what they need, they demonstrated how they use XCode’s autocomplete to do this. This is what UITableViewDelegate does and may be very natural for iOS developers. One thing about the way that delegate is structured is that all methods start with the same tag: 

```Swift
tableView(_: willDisplay: forRowAt:) 
tableView(_: indentationLevelForRawAt:) 
tableView(_: shouldSpringLoadRowAt: with:) 
tableView(_: didSelectRowAt:) 
```

This aids discoverability with this method because the user can type “tableView” and that narrows the search results. We might consider a similar scheme for the TransferDelegate: 

```Swift
blobClient(_: didUpdateTransfer: withState: andProgress:) 
blobClient(_: didCompleteTransfer) 
blobClient(_: didFailTransfer: withError) 
```
 
Currently the “tags” (leading part of the name) are different for the various methods. 

In general, it is undocumented and unclear to users when the various delegate methods are called, especially `transfers(_: didUpdateWithState:)`.The idea here was to trigger a single UI refresh when a whole batch of transfers changed in a single operation (such as pauseAll or resumeAll) instead of individual UI refresh calls for each affected item, but that’s not exactly made clear by the existing API.

tjprescott:
Closes #308. 

- [x] Add progress callback to `upload`, `download` and `resume` methods.
- [x] Refactor startManaging to permit adding callback to resume. (Related: #313)
- [x] Update delegate methods to fit the iOS pattern for UITableView and others.
- [x] Add do-nothing default methods to the refactored delegate protocol.

**Questions**

<del>- [ ] If a delegate is set... should we notify the delegate AND the callback or should one trump the other?</del> The consensus was: notify both

<del>- [ ] Should the progress callback also be capable of notifying in the error case?</del> The callback is already capable of communicating errors, so yes!

<del>- [ ] Should we streamline the blob client delegate? Should it be at parity with the progress callback or should they be different?</del> Consensus was to let the delegate be more verbose and the callback more succinct, as that is consistent with their usage throughout iOS's history.


examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -203,8 +203,17 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
...
+                progressHandler: { transfer in
+                    if transfer.transferType == .download {
+                        tableView.reloadData()
+                    }
+                }
+            ) { transfer in
+                if transfer.transferType == .download {
+                    tableView.reloadData()
+                }
+            }
```
tjprescott, line 216: This is an idiomatic sample of what this usage would look like based on existing usage in SwiftUI. You could of course also put the `completionHandler` as a parameter in the method call.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -58,6 +58,8 @@ class BlobDownloadViewController: UIViewController, MSALInteractiveDelegate {
         PHPhotoLibrary.authorizationStatus()
         fetchData(self)
         StorageBlobClient.startManaging()
+        guard let blobClient = blobClient else { return }
+        blobClient.downloads.resumeAll(progressHandler: downloadProgress)
```
tjprescott, line 62: This is how the user would resume downloads and reattach the progress handler.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
```diff
@@ -99,7 +100,15 @@ import Foundation
         // MARK: - Listening
 
         public func registerListener(_ listener: @escaping ReachabilityStatusListener) {
-            self.listener = listener
+            self.listener = { status in
+                // no nothing on the initial status update to force the developer to resume transfers
+                // TODO: Maybe this should be set in a more thoughtful manner
+                guard self.initialized else {
+                    self.initialized = true
+                    return
```
tjprescott, line 108: Initialized flag simply serves to skip the initial listener call that establishes the initial connectivity state.

sdk/storage/AzureStorageBlob/Source/TransferManager.swift
```diff
@@ -50,7 +50,8 @@ internal protocol TransferManager {
     func cancel(transfer: TransferImpl)
     func pause(transfer: TransferImpl)
     func remove(transfer: TransferImpl)
-    func resume(transfer: TransferImpl)
+    // TODO: Resolve issue where handler can't be downcast
+    func resume(transfer: TransferImpl, progressHandler: ((BlobTransfer) -> Void)?)
```
tjprescott, line 54: Could not cast a `((BlobTransfer) -> Void)` closure to a `((TransferImpl) -> Void)` closure, so this is what I had to do just to get it to work for the time being.

bsiegel:
> If a delegate is set... should we notify the delegate AND the callback or should one trump the other?
If both exist, we should probably call both. It would be surprising to do anything else, but we should have docs that describe how to use each option and help the customer avoid using both at the same time.

> Should the progress callback also be capable of notifying in the error case?
It depends. If we have a single callback for status & progress, then obviously yes. If there is a separate callback for progress and another one for status, only status should notify of an error since that is a status change. I would suggest that in that case, we don't do anything with progress at all (rather than, say, resetting it to zero or something).

> Should we streamline the blob client delegate? Should it be at parity with the progress callback or should they be different?
A good question. I would suggest that it should probably match the feature set of the callbacks. That is, if we have only a single callback for progress & status, then the delegate should have one method as well. If we use separate callbacks for progress vs. status then the delegate should have two methods as well.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
bsiegel: If this works, I'm okay with it for now. I think perhaps the "proper" way to do this would be to store the current state somewhere (initially "unknown") and if the previous state was unknown, don't call the callback but only store the new state.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
```diff
@@ -99,7 +100,15 @@ import Foundation
         // MARK: - Listening
 
         public func registerListener(_ listener: @escaping ReachabilityStatusListener) {
-            self.listener = listener
+            self.listener = { status in
+                // no nothing on the initial status update to force the developer to resume transfers
```
bsiegel, line 104:
```suggestion
                // do nothing on the initial status update to force the developer to resume transfers
```

sdk/storage/AzureStorageBlob/Source/TransferManager.swift
bsiegel: I initially considered a conversion function that would wrap a ` ((BlobTransfer) -> Void)` in another closure of type `((TransferImpl) -> Void)` and just do the cast internally, but I don't think that's the right path either. This is fine for now. I think the better approach down the road might involve generics of type `T: TransferImpl` rather than just taking a `TransferImpl`, especially since we switch on the *actual* type of the `TransferImpl` in a lot of our method bodies.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
```diff
@@ -78,7 +79,10 @@ extension URLSessionTransferManager: TransferDelegate {
             for restorationId in Set(restorationIds) {
                 guard let delegate = self.client(forRestorationId: restorationId) as? TransferDelegate else { continue }
                 let transfersForId = transfers.filter { ($0 as? TransferImpl)?.clientRestorationId == restorationId }
-                delegate.transfersDidUpdate(transfersForId)
+                for transfer in transfersForId {
+                    let progress = (transfer as? BlobTransfer)?.progress
+                    delegate.transfer(transfer, didUpdateWithState: transfer.state, andProgress: progress)
```
bsiegel, line 84: Need to also call the progressHandler here?

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
```diff
@@ -91,9 +95,13 @@ extension URLSessionTransferManager: TransferDelegate {
...
     func transferDidComplete(_ transfer: Transfer) {
         guard let restorationId = (transfer as? TransferImpl)?.clientRestorationId else { return }
-        guard let delegate = client(forRestorationId: restorationId) as? TransferDelegate else { return }
+        let delegate = client(forRestorationId: restorationId) as? TransferDelegate
+        guard delegate != nil || (transfer as? BlobTransfer)?.progressHandler != nil else { return }
         DispatchQueue.main.async {
-            delegate.transferDidComplete(transfer)
+            delegate?.transferDidComplete(transfer)
+            if let blobTransfer = transfer as? BlobTransfer {
+                blobTransfer.progressHandler?(blobTransfer)
```
bsiegel, line 103: I think we need to do the same thing in `self.transfer(_:didFailWithError:)`.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
tjprescott: It was a quick fix and I liked your approach better. I also make sure that the status did actually change.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
tjprescott: Comment removed.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
tjprescott: Done

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager+TransferDelegate.swift
tjprescott: Done.

tjprescott:
@bsiegel changes pushed.

sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
```diff
@@ -101,13 +101,10 @@ import Foundation
 
         public func registerListener(_ listener: @escaping ReachabilityStatusListener) {
             self.listener = { status in
-                // no nothing on the initial status update to force the developer to resume transfers
-                // TODO: Maybe this should be set in a more thoughtful manner
-                guard self.initialized else {
-                    self.initialized = true
-                    return
-                }
+                guard self.previousStatus != .unknown,
```
bsiegel, line 104: If `previousStatus` is unknown, we still need to update it. Otherwise it will never move out of the `unknown` state.

sdk/storage/AzureStorageBlob/Source/Data/BlobTransfer+CoreDataClass.swift
```diff
@@ -85,6 +85,8 @@ public class BlobTransfer: NSManagedObject, Transfer {
         }
     }
 
+    internal var previousState: TransferState?
```
tjprescott, line 88: I avoided adding this previously because I didn't want to serialize it... however, unlike the property for previous progress, there really is no need to save this, so I just track it internally.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient+Extensions.swift
```diff
@@ -27,18 +27,13 @@
 import AzureCore
 import Foundation
 
-// MARK: Paged Collection Delegate
+// MARK: PageableClient
```
tjprescott, line 30: Just moved this to `StorageBlobClient+Extensions.swift`

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -430,7 +430,8 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         case let transfer as BlockTransfer:
             operationQueue.add(BlockOperation(withTransfer: transfer, delegate: self))
         case let transfer as BlobTransfer:
-            transfer.progressHandler = progressHandler
+            // if progress handler not provided, do not overwrite with nil
+            transfer.progressHandler = progressHandler ?? transfer.progressHandler
```
tjprescott, line 434: Prevents the reachability listener from wiping out the attached progress handler when calling `resumeAll`.

tjprescott:
@bsiegel comments addressed... and some small fixes based on testing the transitions.

sdk/storage/AzureStorageBlob/Source/Operation/MultiBlobOperation.swift
```diff
@@ -41,7 +41,7 @@ internal class MultiBlobOperation: TransferOperation {
 
     // MARK: Public Methods
 
-    override public func main() {
```
bsiegel, line 44: Looks like a regression from #306, try updating your SwiftLint + SwiftFormat.

Going forward I wonder if there is any way to better protect about rule changes between versions of these tools. Like maybe have a bot that applies the latest version of swiftlint + swiftformat changes and submits PRs automatically.

tjprescott: I would hesitate to do that because, at least drawing my experience in Python, updates to pylint would come out that we absolutely did not want, and we would often pin to an older, specific version of pylint.

bsiegel: Hmm, I wonder if there is a way we can pin the tools versions for swiftlint / swiftformat. At the very least we *could* have the build step that runs them download (or use existing copy) of a specific version of the tool rather than relying on devs to have their own system-wide copies acquired from e.g. Homebrew

PR #321, opened 2020-06-01
Add delete blob operation

### LINKED ISSUE #215: AzureStorageBlob must be able to remove blobs in a container

tjprescott:
Closes #215.

Adds delete operation to the `BlobStorageClient` and to the demo app.

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
@@ -161,6 +161,13 @@ extension HTTPHeaders {
     public init(_ values: [HTTPHeader: String]) {
         self.init(minimumCapacity: values.underestimatedCount)
         for (key, value) in values {
+            self[key.rawValue] = value
+        }
+    }
+
+    public init(_ values: [String: String]) {
+        self.init(minimumCapacity: values.underestimatedCount)
+        for (key, value) in values {
             self[key] = value
         }
     }
--- a/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
@@ -164,11 +164,4 @@ extension HTTPHeaders {
             self[key.rawValue] = value
         }
     }
-
-    public init(_ values: [String: String]) {
-        self.init(minimumCapacity: values.underestimatedCount)
-        for (key, value) in values {
-            self[key] = value
-        }
-    }
 }
--- a/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
+++ b/sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
@@ -411,7 +411,7 @@ public final class StorageBlobClient: PipelineClient {
         blob: String,
         inContainer container: String,
         withOptions options: DeleteBlobOptions? = nil,
-        then completion: @escaping HTTPResultHandler<Data?>
+        completionHandler: @escaping HTTPResultHandler<Data?>
     ) {
         // Construct URL
         let urlTemplate = "{container}/{blob}"
@@ -426,7 +426,7 @@ public final class StorageBlobClient: PipelineClient {
 
         // Construct headers
         var headers = HTTPHeaders([
-            HTTPHeader.apiVersion.rawValue: self.options.apiVersion
+            .apiVersion: self.options.apiVersion
         ])
         if let deleteSnapshots = options?.deleteSnapshots {
             headers[.deleteSnapshots] = deleteSnapshots.rawValue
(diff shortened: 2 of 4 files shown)


sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
```diff
@@ -159,6 +159,13 @@ extension HTTPHeaders {
     /// - Parameters:
     ///   - values: A dictionary of `HTTPHeader` values to populate the `HTTPHeaders` instance.
     public init(_ values: [HTTPHeader: String]) {
+        self.init(minimumCapacity: values.underestimatedCount)
+        for (key, value) in values {
+            self[key.rawValue] = value
+        }
+    }
+
+    public init(_ values: [String: String]) {
```
bsiegel, line 168: I think we intentionally avoided having such an initializer in order to ensure that any HTTP headers that needed to accompany a request would be added to the appropriate structures (`HTTPHeader`, `StorageHTTPHeader`, etc)

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -401,6 +401,70 @@ public final class StorageBlobClient: PipelineClient {
...
+    /// - Parameters:
+    ///   - blob: The blob name to delete.
+    ///   - container: The container name containing the blob to delete.
+    ///   - options: A `DeleteBlobOptions` object to control the delete operation.
+    ///   - completion: A completion handler that receives a `PagedCollection` of `BlobItem` objects on success.
+    public func delete(
+        blob: String,
+        inContainer container: String,
+        withOptions options: DeleteBlobOptions? = nil,
+        then completion: @escaping HTTPResultHandler<Data?>
```
bsiegel, line 414: Guidelines say this should be called `completionHandler`. It probably doesn't need separate internal & external names.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -401,6 +401,70 @@ public final class StorageBlobClient: PipelineClient {
...
+            "blob": blob
+        ]
+        guard let url = self.url(forTemplate: urlTemplate, withKwargs: pathParams) else { return }
+
+        // Construct query
+        var queryParams: [QueryParameter] = []
+
+        // Construct headers
+        var headers = HTTPHeaders([
+            HTTPHeader.apiVersion.rawValue: self.options.apiVersion
```
bsiegel, line 429: Does this not work?

```swift
var headers = HTTPHeaders([
    .apiVersion: self.options.apiVersion
])
```

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: This was essentially copy-pasta. I could:

1) update it here now in this one place
2) update it here everywhere
3) leave as is here and have it updated in a different PR

sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
tjprescott: The problem is that the existing initializer will not work with mixed types: `[HTTPHeader, StorageHTTPHeader]`, even though both of these are simply type aliases for string-based dictionaries! Swift will complain that `StorageHTTPHeader` is not an `HTTPHeader`. Perhaps an enum is not the best structure to use here?

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Yes, that will work. I switched to the String version because originally I was trying to also include the `StorageHTTPHeader.deleteSnapshots` here before I pulled it out (since it's optional). I can certainly fix this up.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: I'm fine with #1 - we can go back and update the others later, but no need to add new code now that we know we need to change later.

sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
bsiegel: Yeah, we have initializers for both types, but as you point out it doesn't work if they're mixed. Thankfully it's easy enough to create it with one type and then add members of another, IIRC only the initializers suffer from this problem. I'm not sure if it could be fixed with the appropriate use of generics or not, but maybe worth investigating. That said, it's not really consumer-facing (more a problem for library authors) so I'm less motivated to spend the cycles to make this more elegant.

tjprescott: Makes sense. I reverted the second initializer. Should be good now.

tjprescott:
@bsiegel comments should be addressed. Let me know what you think.

PR #327, opened 2020-06-08
Network policy configurability

### LINKED ISSUE #295: TransferManager network configurability
Policies needed to control the automatic pause/resume features of TransferManager. For example: download only on WiFi. Currently, TM will resume given any network connectivity such as Wifi or cellular data.

This feature would also need to be documented.

tjprescott:
Closes #295.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/Common.swift
+++ b/examples/AzureSDKDemoSwift/Source/Common.swift
@@ -96,15 +96,8 @@ struct AppState {
                 tenant: AppConstants.tenant, clientId: AppConstants.clientId, application: application,
                 account: AppState.currentAccount
             )
-            let downloadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet], enableAutoResume: [])
-            let uploadPolicy = TransferNetworkPolicy(
-                transferOver: [.wifiOrEthernet, .cellular],
-                enableAutoResume: [.wifiOrEthernet, .cellular]
-            )
             let options = StorageBlobClientOptions(
-                logger: ClientLoggers.none,
-                downloadNetworkPolicy: downloadPolicy,
-                uploadNetworkPolicy: uploadPolicy
+                logger: ClientLoggers.none
             )
             AppState.internalBlobClient = try? StorageBlobClient(
                 credential: credential,
--- a/examples/AzureSDKDemoSwift/Source/Common.swift
+++ b/examples/AzureSDKDemoSwift/Source/Common.swift
@@ -96,7 +96,13 @@ struct AppState {
                 tenant: AppConstants.tenant, clientId: AppConstants.clientId, application: application,
                 account: AppState.currentAccount
             )
-            let options = StorageBlobClientOptions(logger: ClientLoggers.none)
+            let downloadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet], enableAutoResume: false)
+            let uploadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet, .cellular], enableAutoResume: true)
+            let options = StorageBlobClientOptions(
+                logger: ClientLoggers.none,
+                downloadNetworkPolicy: downloadPolicy,
+                uploadNetworkPolicy: uploadPolicy
+            )
             AppState.internalBlobClient = try? StorageBlobClient(
                 credential: credential,
                 endpoint: AppConstants.storageAccountUrl,
--- a/examples/AzureSDKDemoSwift/Source/Common.swift
+++ b/examples/AzureSDKDemoSwift/Source/Common.swift
@@ -96,8 +96,11 @@ struct AppState {
                 tenant: AppConstants.tenant, clientId: AppConstants.clientId, application: application,
                 account: AppState.currentAccount
             )
-            let downloadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet], enableAutoResume: false)
-            let uploadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet, .cellular], enableAutoResume: true)
+            let downloadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet], enableAutoResume: [])
+            let uploadPolicy = TransferNetworkPolicy(
+                transferOver: [.wifiOrEthernet, .cellular],
+                enableAutoResume: [.wifiOrEthernet, .cellular]
+            )
             let options = StorageBlobClientOptions(
                 logger: ClientLoggers.none,
                 downloadNetworkPolicy: downloadPolicy,
--- a/sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
+++ b/sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
@@ -102,8 +102,7 @@ import Foundation
         public func registerListener(_ listener: @escaping ReachabilityStatusListener) {
             self.listener = { status in
                 defer { self.previousStatus = status }
-                guard self.previousStatus != .unknown,
-                    self.previousStatus != status else { return }
+                guard self.previousStatus != status else { return }
                 listener(status)
             }
         }
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -34,6 +34,10 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     /// The `ClientLogger` to be used by this `StorageBlobClient`.
     public let logger: ClientLogger
 
+    internal let downloadNetworkPolicy: TransferNetworkPolicy
+
+    internal let uploadNetworkPolicy: TransferNetworkPolicy
+
     // Blob operations
 
     /// The maximum size of a single chunk in a blob upload or download.
@@ -48,11 +52,21 @@ public struct StorageBlobClientOptions: AzureConfigurable {
     public init(
         apiVersion: StorageBlobClient.ApiVersion = .latest,
         logger: ClientLogger = ClientLoggers.default(tag: "StorageBlobClient"),
-        maxChunkSizeInBytes: Int = 4 * 1024 * 1024 - 1
+        maxChunkSizeInBytes: Int = 4 * 1024 * 1024 - 1,
+        downloadNetworkPolicy: TransferNetworkPolicy? = nil,
+        uploadNetworkPolicy: TransferNetworkPolicy? = nil
     ) {
         self.apiVersion = apiVersion.rawValue
         self.logger = logger
         self.maxChunkSizeInBytes = maxChunkSizeInBytes
+        self.downloadNetworkPolicy = downloadNetworkPolicy ?? TransferNetworkPolicy(
+            transferOver: [.wifiOrEthernet, .cellular],
+            enableAutoResume: true
+        )
+        self.uploadNetworkPolicy = uploadNetworkPolicy ?? TransferNetworkPolicy(
+            transferOver: [.wifiOrEthernet, .cellular],
+            enableAutoResume: true
+        )
     }
 }
 
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -59,13 +59,15 @@ public struct StorageBlobClientOptions: AzureConfigurable {
         self.apiVersion = apiVersion.rawValue
         self.logger = logger
         self.maxChunkSizeInBytes = maxChunkSizeInBytes
+
+        // apply default `TransferNetworkPolicy` if none supplied
         self.downloadNetworkPolicy = downloadNetworkPolicy ?? TransferNetworkPolicy(
             transferOver: [.wifiOrEthernet, .cellular],
-            enableAutoResume: true
+            enableAutoResume: [.wifiOrEthernet, .cellular]
         )
         self.uploadNetworkPolicy = uploadNetworkPolicy ?? TransferNetworkPolicy(
             transferOver: [.wifiOrEthernet, .cellular],
-            enableAutoResume: true
+            enableAutoResume: [.wifiOrEthernet, .cellular]
         )
     }
 }
--- a/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
+++ b/sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
@@ -61,14 +61,8 @@ public struct StorageBlobClientOptions: AzureConfigurable {
         self.maxChunkSizeInBytes = maxChunkSizeInBytes
 
         // apply default `TransferNetworkPolicy` if none supplied
-        self.downloadNetworkPolicy = downloadNetworkPolicy ?? TransferNetworkPolicy(
-            transferOver: [.wifiOrEthernet, .cellular],
-            enableAutoResume: [.wifiOrEthernet, .cellular]
-        )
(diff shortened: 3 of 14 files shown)


sdk/core/AzureCore/Source/Util/ReachabilityManager.swift
```diff
@@ -100,12 +99,7 @@ import Foundation
         // MARK: - Listening
 
         public func registerListener(_ listener: @escaping ReachabilityStatusListener) {
-            self.listener = { status in
-                defer { self.previousStatus = status }
-                guard self.previousStatus != .unknown,
-                    self.previousStatus != status else { return }
```
tjprescott, line 106: This might get reintroduced.

tjprescott:
@bsiegel ping :)

tjprescott:
@bsiegel please take a look at this PR.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -100,6 +113,30 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         return container
     }()
 
+    /// Retrieve all managed transfers created by the transfer manager.
+    internal var transferCollection: TransferCollection {
+        let matching: [BlobTransfer] = StorageBlobClient.manager.transfers.compactMap { $0 as? BlobTransfer }
```
bsiegel, line 118: Shouldn't this (and the other instances of this line, below) just be `transfers.compactMap`? Since the TM has its list of transfers already.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -100,6 +113,30 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         return container
     }()
 
+    /// Retrieve all managed transfers created by the transfer manager.
```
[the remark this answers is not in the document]
bsiegel: (If we decide not to do that, then it appears `transferCollection` is unused.)

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -48,11 +48,22 @@ public struct StorageBlobClientOptions: AzureConfigurable {
...
         logger: ClientLogger = ClientLoggers.default(tag: "StorageBlobClient"),
-        maxChunkSizeInBytes: Int = 4 * 1024 * 1024 - 1
+        maxChunkSizeInBytes: Int = 4 * 1024 * 1024 - 1,
+        downloadConnectivityPolicy: TransferNetworkPolicy? = nil,
+        uploadConnectivityPolicy: TransferNetworkPolicy? = nil
     ) {
         self.apiVersion = apiVersion.rawValue
         self.logger = logger
         self.maxChunkSizeInBytes = maxChunkSizeInBytes
+        URLSessionTransferManager.shared.uploadConnectivityPolicy = uploadConnectivityPolicy ?? TransferNetworkPolicy(
```
bsiegel, line 58: This logic isn't quite right. If the customer initializes a `StorageBlobClientOptions`, the settings it contains shouldn't have any effect until it's used (at the earliest - when the client is created).

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -446,6 +515,28 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
         self.transfer(transfer, didUpdateWithState: transfer.state, andProgress: nil)
     }
 
+    // MARK: Misc Methods
+
+    func handleNetworkTransition() {
+        let status = networkStatus
+        if let downloadPolicy = downloadConnectivityPolicy {
+            if downloadPolicy.transferOver.contains(status) {
```
bsiegel, line 523: We should call `shouldTransfer` instead (here and below) for consistency.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
```diff
@@ -0,0 +1,58 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import AzureCore
+import Foundation
+
+/// Describes the type of network connection
+public enum NetworkType {
+    case unknown
```
bsiegel, line 32: I think we probably need to have two different enums, and map between them:

1.) The network types that the developer can choose to allow or deny transfers on. This only has two values, wifiOrEthernet and cellular, and is public.
2.) The current network state. This has all four values and is internal. We use this to keep track of the current network state.

Mostly because we don't want a developer to be able to say "transfer only over `disconnected`", that doesn't make sense. But we want to know when we are going from a connected network to `disconnected` or vice versa.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
```diff
@@ -0,0 +1,58 @@
...
+/// Describes the network state
+public struct NetworkState {
+    var type: NetworkType
+}
+
+public struct TransferNetworkPolicy {
+    /// Permit transfers only on permitted values of `NetworkType`.
+    public let transferOver: [NetworkType]
+
+    public let enableAutoResume: Bool
```
bsiegel, line 47: I propose we also make this a list of `NetworkType`. It makes sense to me to be able to say "allow transfers over any network type, but only auto-resume them on wifi". I'd also want to make it very clear what "auto-resume" means. These are two different scenarios:

1.) The device goes from disconnected to cellular. In this case, auto-resume would permit the TM to resume the transfer that was stopped when we got disconnected.

2.) The device goes from wifi to cellular. In this case, the TM should /not/ auto-resume because if a transfer were manually paused by the developer, it would be surprising that our handy auto-resume feature would kick in at this time and force it to resume.

We need to make sure this is documented well, and also make sure we account for the difference between these two scenarios in our TM auto-resume logic.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -265,6 +302,20 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
             }
             transfer.totalBlocks = Int64(transfer.transfers.count)
         }
+
+        // if transfer should not occur on specific network type, it should immediately pause
+        switch transfer.transferType {
```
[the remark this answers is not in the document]
tjprescott: Updated.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -416,22 +467,40 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
+            return downloadConnectivityPolicy?.shouldTransfer(withStatus: networkStatus) ?? true
+        case .upload:
+            return uploadConnectivityPolicy?.shouldTransfer(withStatus: networkStatus) ?? true
+        }
+    }
+
     func resumeAll(withRestorationId restorationId: String? = nil, progressHandler: ((BlobTransfer) -> Void)? = nil) {
         let toResume = restorationId == nil ? transfers : transfers.filter { $0.clientRestorationId == restorationId }
         for transfer in toResume {
+            if let blobTransfer = transfer as? BlobTransfer { print(blobTransfer.hash) }
```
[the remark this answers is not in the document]
tjprescott: Removed.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -446,6 +515,28 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
+            if downloadPolicy.transferOver.contains(status) {
+                if downloadPolicy.enableAutoResume {
+                    downloadCollection.resumeAll()
+                }
+            } else {
+                downloadCollection.pauseAll()
+            }
+        }
+        if let uploadPolicy = uploadConnectivityPolicy {
+            if uploadPolicy.transferOver.contains(status), uploadPolicy.enableAutoResume {
```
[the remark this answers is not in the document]
tjprescott: The download logic is the intended behavior.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: Done.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
tjprescott: In Android, @anuchandy distinguishes between system paused and user paused. I'm okay with this change--you might want it to autoresume on Wifi but not necessarily on cellular, for example--but should we then validate for consistency if you only allow transfer over Wifi but pass both Wifi and cellular to autoresume?

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
tjprescott: Done.

bsiegel: Also, not sure if this is how you implemented it or not, but for #1, this should probably be an OptionSet rather than a list of enum values.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
bsiegel: We could, but I don't think we necessarily need to. If you allow = [wifi] and autoresume = [wifi, cellular] I think it's pretty clear that on wifi, transfers will be allowed and will auto resume, and on cellular transfers will not be allowed (the auto resume doesn't even come into play)

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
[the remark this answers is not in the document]
tjprescott: I updated to have the StorageBlobClient variables pull from the transfer manager implementations. The real difference is the TM version pulls all transfers, whereas the blob client version returns only the subset that was created by that client (matches restorationId).

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
tjprescott: Updated the logic so the transfer manager can reference the policy directly from the client object.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
tjprescott: I converted both to an OptionSet.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
tjprescott: Pushed a change that implements this. I included an assert so that having this scenario will fail during development, but not crash a production app.

tjprescott:
@bsiegel all comments should be addressed. I opened a new issue to track a bug that was masked by the old implementation (#336). I would still prefer to remove `stopManaging` in this PR, and tackle this small bug in a separate PR.

tjprescott:
@bsiegel please take a look at this PR.

examples/AzureSDKDemoSwift/Source/Common.swift
```diff
@@ -96,7 +96,16 @@ struct AppState {
                 tenant: AppConstants.tenant, clientId: AppConstants.clientId, application: application,
                 account: AppState.currentAccount
             )
-            let options = StorageBlobClientOptions(logger: ClientLoggers.none)
+            let downloadPolicy = TransferNetworkPolicy(transferOver: [.wifiOrEthernet], enableAutoResume: [])
```
bsiegel, line 99: Since these are both parameters of the same type, maybe we should give them similar name. E.g. `transferOn` and `autoResumeOn` ?

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
```diff
@@ -48,11 +52,23 @@ public struct StorageBlobClientOptions: AzureConfigurable {
...
+        maxChunkSizeInBytes: Int = 4 * 1024 * 1024 - 1,
+        downloadNetworkPolicy: TransferNetworkPolicy? = nil,
+        uploadNetworkPolicy: TransferNetworkPolicy? = nil
     ) {
         self.apiVersion = apiVersion.rawValue
         self.logger = logger
         self.maxChunkSizeInBytes = maxChunkSizeInBytes
+
+        // apply default `TransferNetworkPolicy` if none supplied
+        self.downloadNetworkPolicy = downloadNetworkPolicy ?? TransferNetworkPolicy(
```
bsiegel, line 64: May be a good idea to DRY this up into a `TransferNetworkPolicy.always` (naming?) or something? It could be internal if we don't want to expose it to the customer (I'm undecided on whether that would be a good idea or not).

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
```diff
@@ -0,0 +1,95 @@
...
+    public static let cellular = NetworkType(rawValue: 1 << 1)
+
+    // MARK: Initializers
+
+    public init(rawValue: Self.RawValue) {
+        self.rawValue = rawValue
+    }
+}
+
+internal struct NetworkTypeInternal: OptionSet {
```
bsiegel, line 45: I might just call this `NetworkState` instead, since it tracks the current state of the network.

Also, I understand your comment about making them both OptionSets to be consistent, but it doesn't accurately model reality that way. The network can only be in one state at a time, but the customer can express a desire for multiple network states. So unless it makes the implementation extremely ugly, I would suggest we change this one back to a simple Enum type.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -446,6 +492,32 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
...
+    // MARK: Misc Methods
+
+    func handleNetworkTransition() {
+        let status = networkStatus.publicValue
+
+        let clientEnumerator = clients.objectEnumerator()
+        while let client: StorageBlobClient = clientEnumerator?.nextObject() as? StorageBlobClient {
+            let downloadPolicy = client.options.downloadNetworkPolicy
+            if downloadPolicy.shouldTransfer(withStatus: status) {
+                if downloadPolicy.enableAutoResume.contains(status!) {
```
bsiegel, line 504: We should be able to avoid these force-unwraps right? Something like the following at the beginning of the while-loop?

```swift
guard let status = status else {
    client.transfers.pauseAll()
    continue
}
```

examples/AzureSDKDemoSwift/Source/Common.swift
tjprescott: Done

sdk/storage/AzureStorageBlob/Source/DataStructures/StorageBlobClientOptions.swift
tjprescott: I made it an internal static variable named `default`.

sdk/storage/AzureStorageBlob/Source/NetworkingPolicy.swift
tjprescott: Done

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: Done

PR #328, opened 2020-06-09
Create AzureIdentity project

bsiegel:
This pulls out the MSAL-related code from core into AzureIdentity. AzureIdentity is now the only part of the SDK that imports `MSAL`. AzureStorageBlob is updated to conditionally compile in the MSALCredential initializer only if AzureIdentity is importable.

This also attempts to get SwiftPM support working, to some degree. It seems like building AzureCore and AzureStorageBlob is working via `swift build --target <LibraryName>`. You can also build AzureIdentity if you first build MSAL.framework and put it somewhere in your framework search path, and then specify the proper `-sdk` and `-target` flags to swiftc:

```bash
swift build --target AzureIdentity \
  -Xswiftc "-sdk" -Xswiftc "`xcrun --sdk iphonesimulator --show-sdk-path`" \
  -Xswiftc "-target" -Xswiftc "x86_64-apple-ios12.0-simulator" \
  -Xswiftc -Fpath/to/folder/containing/MSALframework
```

EDIT: You can now build AzureIdentity without extra flags because we conditionally-compile the MSAL related code only if MSAL is available. However there's (currently) nothing left in AzureIdentity if you don't have MSAL available. You still need to run the full command above if you want to build / test with the MSAL related code included.


bsiegel:
Need to write / update docs. Also not sure whether we should consider pulling other credential / identity related things into AzureIdentity. Will discuss.

sdk/core/AzureCore/Source/Pipeline/Policies/AuthenticationPolicy.swift
```diff
@@ -24,16 +24,21 @@
...
-import MSAL
 
 public typealias TokenCompletionHandler = (AccessToken?, Error?) -> Void
-public typealias MSALResultCompletionHandler = (MSALResult?, Error?) -> Void
 
 public struct AccessToken {
     // MARK: Properties
 
     public let token: String
     public let expiresOn: Int
```
tjprescott, line 34: Per guidelines, should this be a Date that is internally converted to Int?

sdk/core/AzureCore/Source/Pipeline/Policies/AuthenticationPolicy.swift
```diff
@@ -24,16 +24,21 @@
...
+
+    // MARK: Initializers
+
+    public init(token: String, expiresOn: Int) {
```
tjprescott, line 38: Since this is a struct, an initializer of this sort should be a freebie and you shouldn't need to code it.

bsiegel: I thought so too, but the auto-created initializer wasn't public - I was unable to access it from other projects. Was very confused but this got it going. I will investigate further.

sdk/core/AzureCore/Source/Pipeline/Policies/AuthenticationPolicy.swift
bsiegel: Yep, good call :)

bsiegel:
> Is it fair to say that this simply relocated code, or did you also change the underlying implementation?

None of the code itself changed, just things were relocated. I added conditional imports / compiles to support building this together with or without MSAL. I had a few thoughts overnight on how I may want to refactor that part a little bit. The biggest thing I need to understand at this point is why we separate out Identity from Core in general, and what pieces should be put in Identity

tjprescott:
> None of the code itself changed, just things were relocated. I added conditional imports / compiles to support building this together with or without MSAL. I had a few thoughts overnight on how I may want to refactor that part a little bit. The biggest thing I need to understand at this point is why we separate out Identity from Core in general, and what pieces should be put in Identity

I think the main rationale behind it is to allow someone to use Azure Core without brining in MSAL unless they opted for that auth strategy. Good question whether other things should go in there. My initial thought would be that any auth implementations from AzureCore would live there but the protocols would stay in AzureCore, while service-specific implementations would be packaged with their library.

Then again, that means if you wanted to use a non-MSAL auth with AzureCore, you'd be pulling in AzureIdentity and... thus MSAL! The conditional compilation directives imply that you could import AzureIdentity but leave out MSAL. Is that true?

bsiegel:
> Then again, that means if you wanted to use a non-MSAL auth with AzureCore, you'd be pulling in AzureIdentity and... thus MSAL! The conditional compilation directives imply that you could import AzureIdentity but leave out MSAL. Is that true?

Yep, pretty much - we just check `#if canImport(MSAL)` and we can omit just the MSAL-specific stuff. Then MSAL no longer becomes a hard dependency of any of our libraries, but if the developer includes MSAL in their app then the MSAL support in our library will be available.

tjprescott:
@bsiegel, in that case, could we not leave it in AzureCore with the MSAL conditional compilation? @adrianhall do you have thoughts on this?

bsiegel:
> @bsiegel Brandon Siegel FTE, in that case, could we not leave it in AzureCore with the MSAL conditional compilation? @adrianhall Adrian Hall FTE do you have thoughts on this?

In theory yes. I wanted to understand more the motivation for having a separate identity library so I pinged Scott, he wants to have a chat around 4 today.

bsiegel:
Believe this is now ready to go.

AzureIdentity has a soft (optional) dependency on MSAL. If it is not present, basically no code will be compiled in because the only credential AzureIdentity has right now is MSALCredential. In the future we may have other shared types of credentials that don't use MSAL.

AzureStorageBlob now has a soft dependency on both AzureIdentity *and* MSAL. If both are present in the app, the initializer that takes the MSALCredential will be compiled in.

Issue #160, opened 2019-11-27
Pre-authenticated URL (SAS token) AuthenticationPolicy


bsiegel:
This needs to include work for the public API interface that customers will use to provide a method we can call to get a new pre-authenticated URL when the token expires.

- [x] Engineering work #330 
- [x] Sample docs and code #358

PR #330, opened 2020-06-12
Allow SAS token to be refreshed

bsiegel:
Completes the engineering work for #160

We still need to add a sample azure function to show how to vend SAS tokens.

### DIFF
--- a/sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
+++ b/sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
@@ -57,15 +57,31 @@ extension Date {
             formatter.dateFormat = "EEE, dd MMM yyyy HH:mm:ss z"
             return formatter
         }()
+
+        static var iso8601: DateFormatter = {
+            let formatter = DateFormatter()
+            formatter.calendar = Calendar(identifier: .iso8601)
+            formatter.locale = Locale(identifier: "en_US_POSIX")
+            formatter.timeZone = TimeZone(abbreviation: "GMT")
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZZ"
+            return formatter
+        }()
     }
 
     public enum Format {
+        /// Format the `Date` as an RFC 1123-formatted string.
         case rfc1123
 
+        /// Format the `Date` as an ISO-8601-formatted string (in RFC 3339 format).
+        case iso8601
+
+        /// Retrieve a `DateFormatter` for the given date format.
         public var formatter: DateFormatter {
             switch self {
             case .rfc1123:
                 return Formatters.rfc1123
+            case .iso8601:
+                return Formatters.iso8601
             }
         }
     }
--- a/sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
+++ b/sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
@@ -35,8 +35,29 @@ public struct StorageSASCredential: AzureCredential {
     internal var queueEndpoint: String?
     internal var fileEndpoint: String?
     internal var tableEndpoint: String?
-    internal var sasToken: String?
     internal var error: Error?
+    internal var sasToken: String? {
+        didSet { expirationDate = StorageSASCredential.expiration(forToken: sasToken) }
+    }
+
+
+    // Treat tokens that expire in <= 2 minutes as expired
+    private static let expirationOffset: TimeInterval = 2 * 60
+
+    public var expirationDate: Date?
+    public var expired: Bool {
+        guard let expirationDate = expirationDate else { return true }
+        return (expirationDate - StorageSASCredential.expirationOffset) >= Date()
+    }
+
+    // MARK: Static methods
+
+    private static func expiration(forToken sasToken: String?) -> Date? {
+        guard let token = sasToken else { return nil }
+        guard let comps = URLComponents(string: "?\(token)") else { return nil }
+        guard let expiration = comps.queryItems?.filter({ $0.name == "se" }).first?.value else { return nil }
+        return Date(expiration, format: .iso8601)
+    }
 
     // MARK: Initializers
 
@@ -166,6 +187,7 @@ public struct StorageSASCredential: AzureCredential {
 
     private mutating func clear() {
         sasToken = nil
+        expirationDate = nil
         blobEndpoint = nil
         queueEndpoint = nil
         fileEndpoint = nil
@@ -322,7 +344,7 @@ internal class StorageSASAuthenticationPolicy: Authenticating {
     public var next: PipelineStage?
 
     /// A shared access signature credential.
-    public let credential: StorageSASCredential
+    public internal(set) var credential: StorageSASCredential
 
     // MARK: Initializers
 
@@ -339,6 +361,15 @@ internal class StorageSASAuthenticationPolicy: Authenticating {
     ///   - request: A `PipelineRequest` object.
     ///   - completionHandler: A completion handler that forwards the modified pipeline request.
     public func authenticate(request: PipelineRequest, completionHandler: @escaping OnRequestCompletionHandler) {
+        if credential.expired {
+            do {
+                try credential.refresh()
+            } catch {
+                completionHandler(request, error)
+                return
+            }
+        }
+
         let queryParams = parse(sasToken: credential.sasToken ?? "")
         if let requestUrl = request.httpRequest.url.appendingQueryParameters(queryParams) {
             request.httpRequest.url = requestUrl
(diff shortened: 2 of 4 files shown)


sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
```diff
@@ -57,23 +57,64 @@ extension Date {
...
+            formatter.locale = Locale(identifier: "en_US_POSIX")
+            formatter.timeZone = TimeZone(abbreviation: "GMT")
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
+            return formatter
+        }()
     }
 
     public enum Format {
         case rfc1123
+        case iso8601
```
tjprescott, line 82: Why are you not exposing `iso8601WithMilliseconds` here? Is there a way to specify the `WithMilliseconds` aspect with associated values?

sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
```diff
@@ -57,23 +57,64 @@ extension Date {
...
         public var formatter: DateFormatter {
             switch self {
             case .rfc1123:
                 return Formatters.rfc1123
+            case .iso8601:
+                return Formatters.iso8601
+            }
+        }
+
+        fileprivate var alternateFormatters: [DateFormatter] {
```
tjprescott, line 93: I think exposing `.iso8601(withMilliseconds: true)` would be conceptually simpler.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
```diff
@@ -29,23 +29,102 @@ import Foundation
...
+    internal var queueEndpoint: String?
+    internal var fileEndpoint: String?
+    internal var tableEndpoint: String?
+    internal var error: Error?
+    internal var sasToken: String? {
+        didSet { expirationDate = StorageSASCredential.expiration(forToken: sasToken) }
+    }
+
+
+    // Treat tokens that expire in <= 2 minutes as expired
```
tjprescott, line 44: What's the rationale for this? What is the typical expiration?

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
```diff
@@ -29,23 +29,102 @@ import Foundation
...
+        self.blobSasUriProvider = nil
+        self.connectionStringProvider = connectionStringProvider
+        self.error = configureFrom(connectionString: connectionString)
+    }
+
+    /// Create a shared access signature credential from a container- or blob-level shared access signature.
+    /// - Parameters:
+    ///   - blobSasUri: The initial container- or blob-level shared access signature URI.
+    ///   - blobSasUriProvider: A closure that returns a container- or blob-level shared access signature URI.
+    public init(blobSasUri: String, blobSasUriProvider: (() -> String)? = nil) {
```
tjprescott, line 96: Is it possible to provide just a provider and not an initial value... where it would simply need to contact the provider initially? Same question for the connection string variety. Or is it that the initial connection string contains some important info that the connection string provider can't retrieve?

sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
bsiegel: I actually spent a lot of time on this, because it was annoying. In fact, one of the passes I took at it was exactly the suggestion you provided above. However I was trying to reconcile a few things that at first blush don't want to be reconciled:

- The SDK should be opinionated about how dates are formatted as strings
- The SDK must be flexible as to how to parse dates out of strings sent by the server
- The DateFormatter class (and the ISO8601DateFormatter class, which, incredibly, is *NOT* a subclass of DateFormatter) has no ability to be flexible when parsing strings to dates - the string MUST follow the precise format specified in the formatting string.

My attempt to resolve these constraints resulted in the solution you see here. It's asymmetric - the SDK can parse ISO8601 dates in two formats (note: this is still extremely insufficient to be a global solution - the [spec](https://en.wikipedia.org/wiki/ISO_8601) denotes a wide variety of formats that are valid ISO8601 date strings) but has an opinion about the only "correct" format in which a date should be rendered as an ISO8601 date string (and the opinion is that millisecond-accuracy is unnecessary, so we never emit fractional seconds).

Frustratingly, the ISO8601DateFormatter class exists that is aware of all of the different "options" you have available when formatting a date into an ISO8601 date string. But unfortunately it suffers the same limitation as DateFormatter, in that you must set the options to exactly match those present in the date string before parsing it into a Date, otherwise it will return nil even if the string is a valid ISO8601 date string.

I'm totally open to other suggestions here. If there were a way to accept *any* valid ISO8601 date string when parsing a string to a Date, I would much prefer that. And likewise if it could be encoded into the API elegantly, I would love to allow an option set of `ISO8601DateFormatter.Options` to be provided when formatting a Date to a string in ISO8601 format.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
bsiegel: So, I was under the impression that the server would provide a meaningful error response if the token was expired or otherwise invalid, but it seems that's not (always?) the case. In at least some cases, an expired token simply returns a 404 as if the entity did not exist at all. So the logic for determining whether the token is expired or not is (unfortunately) entirely based on the expiration date stamped into the token itself.

In order to have a greater chance of success and to prevent the customer from getting an unexpected 404 or other error response, we optimistically refresh the token just a little bit before it expires. I'm trying to recall where this value came from, but it was discussed last week and it was suggested that other SDKs used a value of two minutes in order to be as small as possible while mitigating (small amounts of) clock skew between the client and the server.

I would definitely want to get some feedback from other language SDK developers and/or the storage service developers on this.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
bsiegel: Yep, those are above. I couldn't collapse them into a single initializer each because then both parameters would be optional and a customer could choose to provide neither of them. So now you can either provide a credential provider closure, or a static credential and an optional credential provider to be used to refresh the credential.

bsiegel: I actually debated removing the version with the static credential entirely, since we know tokens WILL expire eventually, so a static credential will eventually require to be refreshed. But I figured there might be scenarios where a developer makes a token that expires far in the future because they simply don't want to ever have to deal with the complexity of refreshing it, so they treat it like a static key.

sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
bsiegel: I decided to remove the fallback logic for now. iOS's built-in `JSONDecoder.DateDecodingStrategy` exposes `iso8601` as a strategy which also only parses dates without milliseconds so I feel like there is at least some prior art. And right now I'm only using it for parsing the date in the SAS token which lacks the milliseconds part anyhow. If/when we need to support iso8601 dates with milliseconds or other less common formatting options we can revisit.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
tjprescott: Recommend reaching out to @adrianhall on this question.

bsiegel: Discussed, we will ditch these initializers and go with only the version that takes the callback.

tjprescott:
@bsiegel is this PR ready to be merged?

tjprescott:
@bsiegel what needs to be done to get this PR merged?

bsiegel:
I think this is ready to go. I will add the sample function in a separate PR.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageCredentials.swift
```diff
@@ -29,23 +29,74 @@ import Foundation
...
-    internal let fileEndpoint: String?
-    internal let tableEndpoint: String?
-    internal let sasToken: String?
-    internal let error: Error?
+    internal let credentialProvider: () throws -> String
+    internal var blobEndpoint: String?
+    internal var queueEndpoint: String?
+    internal var fileEndpoint: String?
+    internal var tableEndpoint: String?
+    internal var error: Error?
```
tjprescott, line 37: Why are these changed from `var` to `let`?

bsiegel: Whenever we need to refresh the token, we mutate these properties so they have to be `var`s and the functions `clear()` and `refresh()` have to be marked as `mutating`.

PR #333, opened 2020-06-15
Error handling updates

### LINKED ISSUE #292: Default Result type should include AzureError rather than Error
This should aid in documenting which errors are throwable from methods, and aid the user in error handling.

tjprescott:
Closes #292.

Consolidates to a single `AzureError` type that has three values: 
- `sdk` - any errors generated and thrown by the SDK.
- `system` - any errors passed from the service through the SDK.
- `service` - any error generated from the system or OS that is then passed through the SDK. 

The distinction will hopefully allow us to better triage bug reports by focusing on those errors and error messages that come directly from the SDK.

### DIFF
--- a/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
+++ b/examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
@@ -168,7 +168,7 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
 
     func downloadProgress(transfer: BlobTransfer) {
         guard transfer.state != .failed else {
-            let error = transfer.error ?? AzureError.sdk("An error occurred.")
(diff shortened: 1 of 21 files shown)


tjprescott:
@bsiegel could you take a look at this?

bsiegel:
Pushed 49595ba as a starting point to discuss how we might better model different classes of errors.

tjprescott:
> Pushed [49595ba](https://github.com/Azure/azure-sdk-for-ios/commit/49595baf4547f6018f9c38752322cbdf3c24cfc7) as a starting point to discuss how we might better model different classes of errors.

My main concern with this approach is that the standard expectation for errors in Swift is that they are enums, not structs. I will see if I can get an version that purely uses enums. I agree that have a mix of enums and structs is non-ideal.

bsiegel:
Related / alternative implementation: #348

tjprescott:
Closing this in favor of #348.

PR #334, opened 2020-06-16
Add header validation and etag policy. 

### LINKED ISSUE #161: AzureCore response validation policies
Create policies that operate on the response:
  - [ ] Response validation policy (See: [Android](https://github.com/Azure/azure-sdk-for-android/blob/dev/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/ResponseHeadersValidationInterceptor.java))
  - [ ] Scrub eTag policy (See: [Android](https://github.com/Azure/azure-sdk-for-android/blob/1d7f20437e061255e017de37aad4c6f3ec829001/sdk/storage/azure-storage-blob/src/main/java/com/azure/android/storage/blob/interceptor/NormalizeEtagInterceptor.java))

The "Response validation policy" as implemented in e.g. Java is a generic framework that allows a developer to assert any arbitrary validation on the request/response. However, we only ever used this to do one assertion: assert that for a provided set of headers, the values in the request exactly match the values in the response.

When implementing, just implement a policy that validates this "echo" behavior. Don't make a policy that implements a generic validation framework.

tjprescott:
Closes #161.
- Adds header validation policy
- Adds etag normalization policy
- Fix pipeline stage to allow exceptions.


tjprescott:
@bsiegel could you take a look at this PR. The two new policies are simple and tested. The logic to make `on(response:...)` throwable probably needs the most scrutiny.

bsiegel:
Trying to understand why we went with throwing from `on(response:)` rather than making the `OnResponseCompletionHandler` accept an optional `AzureError` just like `OnRequestCompletionHandler` currently does? Something like the following:

```diff
diff --git a/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift b/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
index e5ed720..206a959 100644
--- a/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
@@ -30,7 +30,7 @@ public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TErr
 public typealias HTTPResultHandler<T> = ResultHandler<T, Error>
 public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
 public typealias OnRequestCompletionHandler = (PipelineRequest, Error?) -> Void
-public typealias OnResponseCompletionHandler = (PipelineResponse) -> Void
+public typealias OnResponseCompletionHandler = (PipelineResponse, Error?) -> Void
 public typealias OnErrorCompletionHandler = (PipelineError, Bool) -> Void
 
 /// Protocol for implementing pipeline stages.
@@ -74,7 +74,7 @@ extension PipelineStage {
     }
 
     public func on(response: PipelineResponse, completionHandler: @escaping OnResponseCompletionHandler) {
-        completionHandler(response)
+        completionHandler(response, nil)
     }
 
     public func on(error: PipelineError, completionHandler: @escaping OnErrorCompletionHandler) {
@@ -106,7 +106,12 @@ extension PipelineStage {
             self.next!.process(request: request) { result, httpResponse in
                 switch result {
                 case let .success(pipelineResponse):
-                    self.on(response: pipelineResponse) { response in
+                    self.on(response: pipelineResponse) { response, error in
+                        if let error = error {
+                            let pipelineError = PipelineError(fromError: error, pipelineResponse: pipelineResponse)
+                            completionHandler(.failure(pipelineError), httpResponse)
+                            return
+                        }
                         completionHandler(.success(response), httpResponse)
                     }
                 case let .failure(pipelineError):
```

tjprescott:
> Trying to understand why we went with throwing from `on(response:)` rather than making the `OnResponseCompletionHandler` accept an optional `AzureError` just like `OnRequestCompletionHandler` currently does? Something like the following:
> 

Good point. I pushed a new commit that makes this change. Please take a look. I would like to merge this before #333.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -107,6 +107,11 @@ public final class StorageBlobClient: PipelineClient {
                 UserAgentPolicy(for: StorageBlobClient.self),
                 RequestIdPolicy(),
                 AddDatePolicy(),
+                NormalizeETagPolicy(),
```
bsiegel, line 110: I checked against the Java client, and it looks like these should be added at the end of the policy. Specifically the order these are inserted in the pipeline should be:

1.) (Everything else)
2.) HeadersValidationPolicy
3.) LoggingPolicy
4.) ScrubEtagPolicy

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
```diff
@@ -0,0 +1,64 @@
...
+        let response = pipelineResponse.httpResponse!
+        var error: AzureError?
+
+        for key in headers {
+            let requestValue = request.headers[key] ?? "NIL"
+            let responseValue = response.headers[key] ?? "NIL"
+            if requestValue != responseValue {
+                error = AzureError
+                    .general(
+                        "Value for header '\(key)' did not match. Expected: \(requestValue) Actual: \(responseValue)"
```
bsiegel, line 58: 👍 on this for now since it matches other clients, however this appears to be a way to potentially avoid the `allowHeaders` list and force the SDK to write out an unredacted header value.

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
```diff
@@ -0,0 +1,64 @@
...
+        completionHandler: @escaping OnResponseCompletionHandler
+    ) {
+        let request = pipelineResponse.httpRequest
+        let response = pipelineResponse.httpResponse!
+        var error: AzureError?
+
+        for key in headers {
+            let requestValue = request.headers[key] ?? "NIL"
+            let responseValue = response.headers[key] ?? "NIL"
+            if requestValue != responseValue {
```
bsiegel, line 55: The logic isn't quite right here. Examining the Java SDK, it appears that the logic should only verify the values match if the response includes a value for the given header. If the response doesn't include the given header, then the check should pass regardless of the request's value for that header (I believe this is intentional given the name of the method that adds the check, `addOptionalEcho`).

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Done.

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
tjprescott: Good point. I will try and put the redacting function in here. Better now than later...

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
tjprescott: Updated the logic and added a test for this case.

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
tjprescott: Updated this logic and added a test for this scenario.

tjprescott:
@bsiegel changes pushed. Please take another look.

bsiegel:
Addressed my remaining comments in c54530e - if you agree with these changes then it's good to merge from my side.

tjprescott:
> Addressed my remaining comments in [c54530e](https://github.com/Azure/azure-sdk-for-ios/commit/c54530e65c1781c723cd02862f89969d0fab7be5) - if you agree with these changes then it's good to merge from my side.

Yep, looks good to me. Thanks for the commit!

PR #335, opened 2020-06-22
Make pipeline calls cancellable

### LINKED ISSUE #325: Allow pipeline operations to be canceled
The guidelines specify that client methods must return an `AzureTask` instance, allowing async operations to be canceled.

To do this, we must:

1. Add the ability for in-flight requests to be canceled
2. Return an AzureTask from the pipeline that permits cancellation of the request
3. Make all client methods return an AzureTask
4. Ensure methods that involve multiple operations propagate the cancellation - we must both cancel the currently in progress task AND do not begin any further operations

tjprescott:
Closes #325. 

Created as Draft PR to gather feedback.

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
@@ -33,13 +33,13 @@ public protocol Cancellable {
 }
 
 public enum TaskState {
-    /// The transfer is currently in progress.
+    /// The task is currently in progress.
     case inProgress
-    /// The transfer completed successfully.
+    /// The task completed successfully.
     case complete
-    /// The transfer failed. This failure may or may not be retryable.
+    /// The task failed. This failure may or may not be retryable.
     case failed
-    /// The transfer was explicitly canceled.
+    /// The task was explicitly canceled.
     case canceled
 
     /// A string representation of the task state.
@@ -65,7 +65,7 @@ public class AzureTask: Cancellable {
     public internal(set) var error: Error?
 
     public func cancel() {
-        // TODO: Should this cancel if it is already failed?
+        guard state == .inProgress else { return }
         AzureTaskManager.shared[id]?.state = .canceled
     }
 
--- a/sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
@@ -1,0 +1,108 @@
+// --------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// The MIT License (MIT)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the ""Software""), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+/// A handle for an asynchronous operation that can be canceled.
+public protocol Cancellable {
+    /// Cancel this operation.
+    func cancel()
+}
+
+public enum TaskState {
+    /// The transfer is currently in progress.
+    case inProgress
+    /// The transfer completed successfully.
+    case complete
+    /// The transfer failed. This failure may or may not be retryable.
+    case failed
+    /// The transfer was explicitly canceled.
+    case canceled
+
+    /// A string representation of the task state.
+    public var label: String {
+        switch self {
+        case .inProgress:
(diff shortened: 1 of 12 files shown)


sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
```diff
@@ -0,0 +1,108 @@
...
+import Foundation
+
+/// A handle for an asynchronous operation that can be canceled.
+public protocol Cancellable {
+    /// Cancel this operation.
+    func cancel()
+}
+
+public enum TaskState {
+    /// The transfer is currently in progress.
```
bsiegel, line 36: Since this is a core class, comments should refer to "task" rather than "transfer"

sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
```diff
@@ -0,0 +1,108 @@
...
+
+public class AzureTask: Cancellable {
+    public internal(set) var id: UUID
+
+    public internal(set) var state: TaskState
+
+    public internal(set) var error: Error?
+
+    public func cancel() {
+        // TODO: Should this cancel if it is already failed?
```
bsiegel, line 68: Probably only if current state is `inProgress`

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -295,13 +295,14 @@ public final class StorageBlobClient: PipelineClient {
...
+    @discardableResult
     public func listContainers(
         withOptions options: ListContainersOptions? = nil,
         completionHandler: @escaping HTTPResultHandler<PagedCollection<ContainerItem>>
-    ) {
+    ) -> AzureTask {
         // Construct URL
         let urlTemplate = ""
-        guard let url = self.url(forTemplate: urlTemplate) else { return }
+        guard let url = self.url(forTemplate: urlTemplate) else { return AzureTask.for(error: "Precondition failure") }
```
bsiegel, line 305: General question - when we error out inside the body of the public API method like this, should we also be calling the completionHandler with the error?

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
```diff
@@ -106,6 +128,14 @@ public class URLSessionTransport: HTTPTransportStage {
                 logger: logger,
                 context: responseContext
             )
+            // Check if clientRequestId is canceled and back out if it is.
```
bsiegel, line 131: We should probably revert this. At this point, the request has been sent already, it's too late to cancel - the operation is complete, so even if cancellation was requested, we should return back the completion to the caller. Not doing so could cause problems. For example, if the caller tried to create a resource, and then canceled the task after the request was already sent. The resource would be created by the service, but this logic here would make them believe the task had been cancelled successfully.

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
```diff
@@ -84,6 +84,28 @@ public class URLSessionTransport: HTTPTransportStage {
             pipelineRequest.logger.error("Invalid session.")
             return
         }
+        guard let clientRequestIdHeader = pipelineRequest.httpRequest.headers[.clientRequestId],
```
bsiegel, line 87: We should do this at the very top, so that any failures further down can also set the task's `state` to `failed`. I noticed that we're not ever setting the state to either `failed` or `completed` during the execution of the task. We should do that any time before we call a callback.

sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
tjprescott: Done

sdk/core/AzureCore/Source/Pipeline/AzureTask.swift
tjprescott: Done

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: That's a good question. Philosophically I would argue no since it is an HTTPResultHandler, and if it errors out as a precondition, there's really no HTTP result to speak of. On the other hand, it would be inconvenient/annoying to need to check for an error both in the completion handler and in the task object. Should probably be a discussion with @adrianhall  and the Android folks as well.

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
tjprescott: Or in lieu of stating that "Request canceled" (since it wasn't) we change the error message to something like "Request cancellation failed," which I would argue is still an error. Requesting cancellation and then getting a success result would seem bizarre to me. Then again, the right answer is probably what do other SDKs do. @adrianhall do you have any thoughts here?

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
tjprescott: Without the client request ID we can't access the task, so we won't be able to set the task state to anything without this unless we attach the task to the pipeline process. I'm not opposed to that, though the intent was really just to have a way of marking request IDs as canceled in order to interrupt the pipeline.

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
tjprescott: After discussion with @JeffreyRichter, this is the correct behavior (albeit not necessarily the correct implementation). Cancellation is best effort and not guaranteed, but if they issue a cancellation request, they should get a cancellation "response". The key is to document this behavior clearly.

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
bsiegel: Yep, that makes sense. My comment was more along the lines of we want to get the task right away, so that any failure like the one above on L84 can also set the task state to failed.

As a separate comment, I suggested we should set the task state to failed or completed any time we call the callback with either a success or an error.

However it sounds like this may no longer be applicable if we go with the cancellation token approach.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
bsiegel: In discussing with Adrian, he mentioned it would make sense to call the callback with the failure immediately. This makes more sense if we decide to not go with the approach of returning an AzureTask.

bsiegel:
Related / alternative implementation: #345

tjprescott:
I think we are all preferring the CancellationToken approach, so I will close this PR for now. We could always reopen it later if need be.

PR #344, opened 2020-07-02
Remove azure-sdk-build-tools repo resource

chidozieononiwu:
Remove reference to `azure-sdk-build-tools` as it's no longer needed.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -1,9 +1,5 @@
 resources:
   repositories:
-  - repository: azure-sdk-build-tools
-    type: git
-    name: internal/azure-sdk-build-tools
-    ref: refs/heads/feature/engsys-ios-barebones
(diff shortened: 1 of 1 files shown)


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -1,9 +1,5 @@
 resources:
   repositories:
-  - repository: azure-sdk-build-tools
```
bsiegel, line 3: This is actually still used, on Line 31 below.

chidozieononiwu: Right.

chidozieononiwu:
Closing this as its not needed.

eng/pipelines/templates/stages/archetype-sdk-client.yml
weshaggard: @mitchdenny do you know if there is any reason the iOS templates cannot be moved out into this repo to remove this repository reference?

PR #345, opened 2020-07-07
Cancellation token implementation

### LINKED ISSUE #325: Allow pipeline operations to be canceled

tjprescott:
Closes #325. Alternate approach to that in PR #335.

Draft PR for initial feedback. If good, I'll need to plumb it through the rest of the SDK calls.


tjprescott:
@bsiegel please take a look.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -0,0 +1,61 @@
...
+    public internal(set) var isCanceled: Bool
+
+    internal var timeout: Double?
+    internal var timeoutRunning: Bool
+
+    public func cancel() {
+        isCanceled = true
+    }
+
+    public init(timeout: Double? = nil) {
```
bsiegel, line 39: IIRC the guidelines specify that fields / parameters should express the units in their name. So presumably this would be `timeoutInSeconds`?

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.timeout = timeout
+        self.timeoutRunning = false
+        self.isCanceled = false
+    }
+
+    // MARK: Equatable Protocol
+
+    public static func == (lhs: CancellationToken, rhs: CancellationToken) -> Bool {
+        return lhs.timeout == rhs.timeout && lhs.isCanceled == rhs.isCanceled
```
bsiegel, line 48: Slightly suspicious about this. Will wait to see how it's used, but I would assume that two tokens are never equal except by hashCode (which I don't believe is a concept in swift but we could easily implement it).

Otherwise I see a few questions - what if one token's timer is running and the other's isn't? And furthermore, are they really equal if their timeout values are the same? Or is it more correct to say they are equal if they expire at the same time (and thus we need to do some time math here)?

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
```diff
@@ -95,7 +91,41 @@ public class URLSessionTransport: HTTPTransportStage {
         let responseContext = pipelineRequest.context
         let logger = pipelineRequest.logger
 
+        if let cancellationToken = pipelineRequest.context?.value(forKey: .cancellationToken) as? CancellationToken {
```
bsiegel, line 94: Maybe just add a helper property `PipelineRequest.isCanceled` that performs this logic since we do this in a few places.

sdk/storage/AzureStorageBlob/Source/DataStructures/TransferCollection.swift
```diff
@@ -47,7 +47,7 @@ public struct TransferCollection {
     }
 
     /// Remove all transfers in this `TransferCollection` from the database. All currently active transfers will be
-    /// cancelled.
+    /// canceled.
```
bsiegel, line 50: Never been quite sure on this, but looking it up online suggests that American english uses one `l` for this word (British english uses two), but two are always used for "cancellation". 🙃

sdk/storage/AzureStorageBlob/Source/Options/ClientMethodOptions.swift
```diff
@@ -38,6 +38,9 @@ public struct ListContainersOptions: AzureOptions {
     /// A client-generated, opaque value with 1KB character limit that is recorded in analytics logs.
     public let clientRequestId: String?
 
+    /// A token used to make a best-effort attempt at cancelling a request.
```
bsiegel, line 41: Soooo... based on the above, this should be "canceling" 🤦

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
bsiegel: Actually I didn't see where this is used at all, can you show me where we compare cancellation tokens for equality?

bsiegel:
LGTM. Had one actual question about the logic, the rest are just naming / code hygiene things.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
tjprescott: It's not used anywhere. Conformance was required so that Options objects remained serializable.

sdk/storage/AzureStorageBlob/Source/DataStructures/TransferCollection.swift
tjprescott: I looked up the same and that's why I changed it 😆

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/Transport/URLSessionTransport.swift
tjprescott: I don't think it would really simplify anything. I would still need to retrieve the cancellation token using this syntax to be able to start the timeout. Best case I could simplify 

```Swift
if let cancellationToken = pipelineRequest.context?.value(forKey: .cancellationToken) as? CancellationToken
```

to

```Swift
if let cancellationToken = pipelineRequest.cancellationToken
```

But we don't do that for anything else we embed in the context, so I'm not sure why we would do that now.

sdk/storage/AzureStorageBlob/Source/Options/ClientMethodOptions.swift
tjprescott: Fixed.

bsiegel:
I was reviewing the general guidelines and they suggest we need to have a global timeout property. I feel like it would make sense to add in this PR as follows:

1. Create a `TransportOptions` structure with a global `timeout` property, and add this as a property called `AzureConfigurable.transportOptions`.

2. When any network call is made, if the `timeout` property is not `nil`, create and attach a cancellation token with the appropriate timeout. If the user already supplied a cancellation token with the call, set the timeout value on the existing cancellation token only if it was nil previously.

bsiegel:
I implemented the global timeout as described above.

As I was working on it I realized that we're not currently respecting the cancellation token for upload/download. I'm guessing that's part of what needs to be done yet to move this out of the draft state? Just making sure it wasn't missed.

tjprescott:
> As I was working on it I realized that we're not currently respecting the cancellation token for upload/download. I'm guessing that's part of what needs to be done yet to move this out of the draft state? Just making sure it wasn't missed.

That's right--I wanted to make sure the cancellation mechanics were correct before applying it throughout.

tjprescott:
Opened #376 to track the completion of this scenario for TransferManager since it has a release date later than AzureCore.

tjprescott:
@bsiegel comments addressed. I have tested the following scenarios:
- Paged collection instantiates the cancellation token on a per-page basis
- TransferManager uses the cancellation on a per-operation (per up/download) basis.

examples/AzureSDKDemoSwift/Source/LoginViewController.swift
```diff
@@ -62,7 +62,11 @@ class LoginViewController: UIViewController {
             redirectUri: nil,
             authority: authority
         )
-        AppState.application = try? MSALPublicClientApplication(configuration: msalConfiguration)
+        do {
```
tjprescott, line 65: this can be reverted.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -0,0 +1,63 @@
...
+            && lhs.isStarted == rhs.isStarted
+            && lhs.isCanceled == rhs.isCanceled
+    }
+
+    // MARK: Methods
+
+    /// Start the cancellation token countdown. If the countdown is already running, this return immediately.
+    public func start() {
+        guard !isStarted, let timeout = timeoutInSeconds else { return }
+        print("STARTING \(timeout) SECOND TIMEOUT!")
```
tjprescott, line 57: Remove this testing line.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -27,39 +27,24 @@
...
     /// The API version of the service to invoke.
     var apiVersion: String { get }
     /// The `ClientLogger` to be used by the service client.
     var logger: ClientLogger { get }
     /// Options for configuring telemetry sent by the service client.
     var telemetryOptions: TelemetryOptions { get }
-}
-
-/// Options for configuring telemetry sent by the service client.
-public struct TelemetryOptions {
```
tjprescott, line 40: Moved this to its own file.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -692,4 +689,17 @@ public final class StorageBlobClient: PipelineClient {
         StorageBlobClient.manager.add(transfer: blobTransfer)
         return blobTransfer
     }
+
+    // MARK: Internal Methods
+
+    internal func addCancellationToken(to context: PipelineContext, from options: AzureOptions?) {
+        if let cancellationToken = options?.cancellationToken {
```
bsiegel, line 696: Slight change in behavior here - I'm not sure which is better. Previously, if the customer provided a cancellation token without a timeout, we would apply the globalTimeout to the existing token. Now, the global timeout is only applied if no cancellation token is supplied at all.

I suspect the previous behavior is preferable - we probably should not make it easy to do network operations with an unlimited timeout. In the previous behavior, the customer could make use of the default timeout and still pass in a token if they wished to cancel manually.

tjprescott: I'm also relocating this logic to be centralized in the `PipelineContext` object.

tjprescott:
@bsiegel comments addressed.

sdk/core/AzureCore/Source/DataStructures/Collections.swift
```diff
@@ -228,18 +228,11 @@ public class PagedCollection<SingleElement: Codable> {
...
-            context.add(
-                value: CancellationToken(timeoutInSeconds: cancellationTimeout) as AnyObject,
-                forKey: .cancellationToken
-            )
-        } else if let globalTimeout = client.commonOptions.transportOptions.timeoutInSeconds {
-            context.add(
-                value: CancellationToken(timeoutInSeconds: globalTimeout) as AnyObject,
-                forKey: .cancellationToken
-            )
+        let cancellationToken = CancellationToken()
```
bsiegel, line 231: I'm pretty sure this doesn't work. Here we're replacing the customer's provided token with a new one, copying over the timeout. If the customer calls `cancel()` on the token they have, nothing will happen.

sdk/core/AzureCore/Source/Pipeline/PipelineContext.swift
```diff
@@ -139,6 +159,26 @@ public class PipelineContext {
...
     }
+
+    /// Add a `CancellationToken` while applying smart defaulting logic. If the client transport options
+    /// specify a timeout, this will be used to automatically create `CancellationToken`s for each call,
+    /// even when a token is not specified. If the client call options contain a `CancellationToken` with
+    /// no timeout the default timeout will be applied, if specified in `AzureClientOptions`.
+    /// - Parameters:
+    ///   - cancellationToken: Optional `CancellationToken` object.
+    ///   - clientOptions: Optional `AzureClientOptions`
+    public func add(cancellationToken: CancellationToken?, applying clientOptions: AzureClientOptions?) {
```
bsiegel, line 170: `clientOptions` is never nil, so we can make this non-optional.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -29,7 +29,7 @@ import Foundation
 public final class CancellationToken: Codable, Equatable {
     public private(set) var isCanceled: Bool
     public private(set) var isStarted: Bool
-    public private(set) var timeoutInSeconds: Double?
+    public var timeoutInSeconds: Double?
```
bsiegel, line 32: We should revert this and add an internal method to set the timeout. We don't want a customer to be able to change the timeout value, because changing it while the timer is running could lead to confusion.

sdk/core/AzureCore/Source/DataStructures/Collections.swift
tjprescott: Good point. Originally I had a restart method but I was concerned what would happen if I just reused the token. If I reset and restarted the timer after it had already been started, I could envision a frustrating race condition. I'll try it out and see how that works. Otherwise we'd need to return a cancellation token, which would be not great.

sdk/core/AzureCore/Source/Pipeline/PipelineContext.swift
tjprescott: Done.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/DataStructures/Collections.swift
tjprescott: I basically use an eTag concept to accomplish this. The timer captures the value at the time it is started so if reset is called, the value will change and it will not mark the token as cancelled.

tjprescott:
@bsiegel please take a look.

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -39,6 +43,7 @@ public final class CancellationToken: Codable, Equatable {
         self.timeoutInSeconds = timeoutInSeconds
         self.isStarted = false
         self.isCanceled = false
+        self.runId = Int.random(in: Int.min ... Int.max)
     }
 
     // MARK: Equatable Protocol
```
bsiegel, line 49: We should replace this implementation with one that just compares the runIDs since they are supposed to be unique

sdk/core/AzureCore/Source/DataStructures/CancellationToken.swift
```diff
@@ -39,6 +43,7 @@ public final class CancellationToken: Codable, Equatable {
         self.timeoutInSeconds = timeoutInSeconds
         self.isStarted = false
         self.isCanceled = false
+        self.runId = Int.random(in: Int.min ... Int.max)
```
bsiegel, line 46: Any reason not to make this a UUID?

PR #348, opened 2020-07-08
Simplify Errors using Enums only

tjprescott:
Alternative to #333.


sdk/storage/AzureStorageBlob/Source/Operations/BlockOperation.swift
```diff
@@ -84,8 +84,10 @@ internal class BlockOperation: TransferOperation {
                     self.notifyDelegate(withTransfer: parent)
                 case let .failure(error):
                     var err = error
-                    if let pipelineError = error as? PipelineError {
-                        err = pipelineError.innerError
+                    switch error {
+                    case let .wrapped(innerError, _):
+                        err = AzureError.service(innerError.localizedDescription)
```
bsiegel, line 89: I'm not sure if this error is actually one that is returned from the service. IIRC "service" is intended to only be for deserialized error response objects actually returned back by the service in its response.

examples/AzureSDKDemoSwift/Source/Common.swift
```diff
@@ -89,7 +89,7 @@ struct AppState {
     static func blobClient(withDelegate delegate: StorageBlobClientDelegate? = nil) throws -> StorageBlobClient {
         if AppState.internalBlobClient == nil {
             guard let application = AppState.application else {
-                let error = AzureError.general("Application is not initialized. Unable to create Blob Storage Client.")
+                let error = AzureError.sdk("Application is not initialized. Unable to create Blob Storage Client.")
```
bsiegel, line 92: I would suggest that consuming applications should not create instances of AzureError at all. Obviously we can't prevent that since there's no "group internal" concept in Swift, but we should maybe document it in the API doc comments for AzureError and its cases.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -168,7 +168,7 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
 
     func downloadProgress(transfer: BlobTransfer) {
         guard transfer.state != .failed else {
-            let error = transfer.error ?? AzureError.general("An error occurred.")
+            let error = transfer.error ?? AzureError.sdk("An error occurred.")
```
bsiegel, line 171: Likewise here about client apps not creating AzureErrors

sdk/core/AzureCore/Source/Errors.swift
```diff
@@ -44,43 +44,27 @@ extension BaseError {
...
-        case let .general(msg),
-             let .decode(msg),
-             let .resourceExists(msg),
-             let .resourceNotFound(msg),
-             let .clientAuthentication(msg),
-             let .resourceModified(msg),
-             let .tooManyRedirects(msg),
-             let .statusCode(msg):
-            return msg
+extension Error {
```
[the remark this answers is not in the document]
tjprescott: Agreed!

sdk/core/AzureCore/Source/Pipeline/PipelineError.swift
```diff
@@ -26,16 +26,16 @@
...
-    public var pipelineResponse: PipelineResponse
-
-    // MARK: Initializers
-
-    init(fromError innerError: Error, pipelineResponse: PipelineResponse) {
-        self.innerError = innerError
-        self.pipelineResponse = pipelineResponse
-    }
-}
+// public final class PipelineError: Error {
```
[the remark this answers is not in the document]
tjprescott: Deleted!

sdk/core/AzureCore/Source/Errors.swift
```diff
@@ -44,43 +44,27 @@ extension BaseError {
 }
 
 public enum AzureError: BaseError {
-    case general(_ message: String)
-    case fileSystem(_ message: String)
-    case serviceRequest(_ message: String)
-    case serviceResponse(_ message: String)
+    case sdk(String)
+    case service(String)
+    case wrapped(Error, PipelineResponse)
```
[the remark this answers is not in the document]
tjprescott: Great points. I simplified to:

```Swift
    case sdk(String, Error?)
    case service(String, PipelineResponse, Error?)
```

sdk/core/AzureCore/Source/Pipeline/Policies/LoggingPolicy.swift
```diff
@@ -87,9 +87,14 @@ public class LoggingPolicy: PipelineStage {
         completionHandler(response, nil)
     }
 
-    public func on(error: PipelineError, completionHandler: @escaping OnErrorCompletionHandler) {
-        LoggingPolicy.queue.async { self.log(response: error.pipelineResponse, withError: error.innerError) }
-        completionHandler(error, false)
+    public func on(error: AzureError, completionHandler: @escaping OnErrorCompletionHandler) {
+        switch error {
+        case let .wrapped(_, pipelineResponse):
+            LoggingPolicy.queue.async { self.log(response: pipelineResponse, withError: error) }
```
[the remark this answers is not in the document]
tjprescott: Fixed.

sdk/identity/AzureIdentity/Source/MSALCredential.swift
```diff
@@ -133,11 +133,13 @@ import Foundation
         public func token(forScopes scopes: [String], completionHandler: @escaping TokenCompletionHandler) {
             let group = DispatchGroup()
             var accessToken: AccessToken?
-            var returnError: Error?
+            var returnError: AzureError?
             group.enter()
             if let account = account {
                 acquireTokenSilently(forAccount: account, withScopes: scopes) { result, error in
-                    returnError = error
+                    if let errorMessage = error?.localizedDescription {
```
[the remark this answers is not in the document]
tjprescott: Fixed.

sdk/storage/AzureStorageBlob/Source/Operations/BlockOperation.swift
tjprescott: Updated.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -119,9 +119,9 @@ extension PipelineStage {
                         completionHandler(.success(response), httpResponse)
                     }
                 case let .failure(pipelineError):
-                    self.on(error: pipelineError) { error, handled in
+                    self.on(error: pipelineError) { _, handled in
                         if !handled {
-                            completionHandler(.failure(error), httpResponse)
+                            completionHandler(.failure(pipelineError), nil)
```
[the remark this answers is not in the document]
tjprescott: I removed the pipeline response from any of the error types. This does mean that `on(error:)` now has to include `on(error:pipelineResponse:)` once again.

examples/AzureSDKDemoSwift/Source/Common.swift
tjprescott: I replaced it with `fatalError()`

tjprescott:
@bsiegel all comments have been addressed. Please take another look.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -97,7 +97,7 @@ open class PipelineClient {
                 let allowedStatusCodes = pipelineResponse.value(forKey: .allowedStatusCodes) as? [Int] ?? [200]
                 if !allowedStatusCodes.contains(httpResponse?.statusCode ?? -1) {
                     self.logError(withData: deserializedData)
-                    let error = HTTPResponseError.statusCode("Service returned invalid status code [\(statusCode)].")
+                    let error = AzureError.service("Service returned invalid status code [\(statusCode)].", nil)
```
bsiegel, line 100: This is probably a `.sdk` - our SDK is defining what the allowed status codes should be.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -27,11 +27,11 @@
...
 
 public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse?) -> Void
-public typealias HTTPResultHandler<T> = ResultHandler<T, Error>
-public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
+public typealias HTTPResultHandler<T> = ResultHandler<T, AzureError>
+public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, AzureError>
 public typealias OnRequestCompletionHandler = (PipelineRequest, Error?) -> Void
 public typealias OnResponseCompletionHandler = (PipelineResponse, Error?) -> Void
-public typealias OnErrorCompletionHandler = (PipelineError, Bool) -> Void
+public typealias OnErrorCompletionHandler = (AzureError, Bool) -> Void
```
bsiegel, line 34: Does it make sense any more to have an `OnErrorCompletionHandler` when our onRequest / onResponse handlers both take an optional Error object?

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -27,11 +27,11 @@
 import Foundation
 
 public typealias ResultHandler<TSuccess, TError: Error> = (Result<TSuccess, TError>, HTTPResponse?) -> Void
-public typealias HTTPResultHandler<T> = ResultHandler<T, Error>
-public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, PipelineError>
+public typealias HTTPResultHandler<T> = ResultHandler<T, AzureError>
+public typealias PipelineStageResultHandler = ResultHandler<PipelineResponse, AzureError>
 public typealias OnRequestCompletionHandler = (PipelineRequest, Error?) -> Void
```
bsiegel, line 32: Should these be updated to take `AzureError` instead?

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -95,10 +104,10 @@ extension PipelineStage {
                     logger: request.logger,
                     context: request.context
                 )
-                let pipelineError = PipelineError(fromError: error, pipelineResponse: pipelineResponse)
-                self.on(error: pipelineError) { _, handled in
+                let error = AzureError.service("Service error.", error)
```
bsiegel, line 107: Is this actually always a service error? IIRC, any pipeline stage can return an error. If we update the signature of onRequest / onResponse handlers to return AzureError, we won't need to wrap it in another error here, and the stage itself can determine whether it's an sdk or a service error.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -108,8 +117,8 @@ extension PipelineStage {
                 case let .success(pipelineResponse):
                     self.on(response: pipelineResponse) { response, error in
                         if let error = error {
-                            let pipelineError = PipelineError(fromError: error, pipelineResponse: response)
-                            self.on(error: pipelineError) { _, handled in
+                            let pipelineError = AzureError.service("Service error.", error)
```
bsiegel, line 120: Same comment as above

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
```diff
@@ -119,9 +128,15 @@ extension PipelineStage {
...
+                    let pipelineResponse = PipelineResponse(
+                        request: request.httpRequest,
+                        response: httpResponse,
+                        logger: request.logger,
+                        context: request.context
+                    )
+                    self.on(error: pipelineError, pipelineResponse: pipelineResponse) { _, handled in
                         if !handled {
-                            completionHandler(.failure(error), httpResponse)
+                            completionHandler(.failure(pipelineError), nil)
```
bsiegel, line 139: Should we not be passing in the response here rather than `nil`?

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -218,7 +219,7 @@ public class ContentDecodePolicy: PipelineStage {
             jsonData = try JSONSerialization.data(withJSONObject: arrayObj, options: [])
         }
         guard let finalJsonData = jsonData else {
-            throw HTTPResponseError.decode("Failure decoding XML.")
+            throw AzureError.sdk("Failure decoding XML.")
```
bsiegel, line 222: Let's capture the thrown error above and include it as the inner error here.

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
```diff
@@ -62,8 +62,9 @@ public class HeadersValidationPolicy: PipelineStage {
                     requestValue = "REDACTED"
                     responseValue = "REDACTED"
                 }
-                error = AzureError.general(
-                    "Value for header '\(key)' did not match. Expected: \(requestValue ?? "nil") Actual: \(responseValue)"
+                error = AzureError.sdk(
+                    "Value for header '\(key)' did not match. Expected: \(requestValue ?? "nil") Actual: \(responseValue)",
+                    nil
```
bsiegel, line 67: Inner error is optional here, can omit the nil parameter.

sdk/storage/AzureStorageBlob/Source/Operations/BlockOperation.swift
```diff
@@ -84,8 +84,10 @@ internal class BlockOperation: TransferOperation {
                     self.notifyDelegate(withTransfer: parent)
                 case let .failure(error):
                     var err = error
-                    if let pipelineError = error as? PipelineError {
-                        err = pipelineError.innerError
+                    switch error {
```
bsiegel, line 87: I'm unclear what this is doing. It seems like it's a no-op?

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
```diff
@@ -537,7 +537,7 @@ internal final class URLSessionTransferManager: NSObject, TransferManager, URLSe
             """
             assertionFailure(errorMessage)
 
-            transfer.error = AzureError.general(errorMessage)
+            transfer.error = AzureError.sdk(errorMessage, nil)
```
bsiegel, line 540: Can omit optional second parameter here.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
tjprescott: The Swagger defines what the allowed status codes can be.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
tjprescott: Fixed.

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
tjprescott: This is the error thrown when none of the above `try` statements fail but there's still no JSON payload, so there's no inner error to wrap. See the updated logic in the `on(response:...)` for this policy.

sdk/core/AzureCore/Source/Pipeline/Policies/HeadersValidationPolicy.swift
tjprescott: Missed that one when I was undoing adding `nil` to everything :)

sdk/storage/AzureStorageBlob/Source/Operations/BlockOperation.swift
tjprescott: It was some wonky logic. Based on other changes, I was able to simplify this so it just says `parent.error = error`.

sdk/storage/AzureStorageBlob/Source/Transport/URLSessionTransferManager.swift
tjprescott: Done.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
tjprescott: So.... arguably yes or no? Maybe? This method is called if either `on(request:...)` or `on(response:...)` returns an error as a way to hopefully handle the error and avoid a failure scenario. So you could handle that logic in those respective methods, and if the logic is different that would certainly make sense. Currently, logging policy is the only policy we have that actually uses this method, so perhaps it would make sense to do away with it.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
bsiegel: Fair enough, I don't feel strongly.

sdk/core/AzureCore/Source/Pipeline/PipelineStage.swift
bsiegel: Filed #357 to look into this another time

Issue #220, opened 2020-02-27
Convert to single pipeline


kurtzeborn:

tjprescott:
PR #354 merged.

PR #354, opened 2020-07-14
Combined pipelines

mitchdenny:
This PR adds a new pipeline (```ios - client``` and ```ios - client - ci```) which is the pipeline that will replace ```ios - [service] - ci``` etc. At this point in time I've left all the CocoaPods wire-up in there since ripping that out will require a few other changes in the project files and I didn't want to overcomplicate this PR.

/cc @bsiegel @tjprescott


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -113,10 +110,10 @@ jobs:
         displayName: 'Publish unit test results'
 
       - script: |
-          echo "Getting list of projects in scheme ${{parameters.ServiceDirectory}}"
+          echo "Getting list of projects in scheme 'ci'"
```
bsiegel, line 113: This should probably print the actual scheme i.e. "all"

bsiegel: Actually is there a reason we don't just keep this file as-is, rename "ServiceDirectory" to "scheme", and pass in "scheme: all" in the ci.yml? It seems like it would let us keep the ability to run against other schemes in the future (e.g. maybe for a manual verification build of some sort).

tjprescott:
@mitchdenny will this PR effectively complete #220?

mitchdenny:
@tjprescott yep

eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -1,30 +1,12 @@
 resources:
   repositories:
-  - repository: azure-sdk-build-tools
-    type: git
-    name: internal/azure-sdk-build-tools
-    ref: refs/heads/feature/engsys-ios-barebones
+  - repository: azure-sdk-tools
```
weshaggard, line 3: Do we still need this repository reference? If so what are we using it for?

cc @chidozieononiwu

PR #355, opened 2020-07-15
Add TelemetryOptions

### LINKED ISSUE #150: AzureCore Global Configuration

- [ ] Support global configuration of pipeline policies:
  - [x] Logging: log level, logger implementation
  - [x] Telemetry: (PR #355)
  - [ ] Retry: (Issue #356)
  - [x] Timeout: (PR #345)

bsiegel:
Allows the user to set the applicationId and telemetryDisabled properties.

Partially completes #150.

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
@@ -32,6 +32,27 @@ public protocol AzureConfigurable {
     var apiVersion: String { get }
     /// The `ClientLogger` to be used by the service client.
     var logger: ClientLogger { get }
+    /// Options for configuring telemetry sent by the service client.
+    var telemetryOptions: TelemetryOptions { get }
+}
+
+/// Options for configuring telemetry sent by the service client.
+public struct TelemetryOptions {
+    /// Whether platform information will be omitted from the user agent string sent by the service client.
+    public let telemetryDisabled: Bool
+    /// An optional user-specified application ID included in the user agent string sent by the service client.
+    public let applicationId: String?
+
+    /// Initialize a `TelemetryOptions` structure.
+    /// - Parameters:
+    ///   - telemetryDisabled: Whether platform information will be omitted from the user agent string sent by the
+    ///   service client.
+    ///   - applicationId: An optional user-specified application ID included in the user agent string sent by the
+    ///   service client.
+    public init(telemetryDisabled: Bool = false, applicationId: String? = nil) {
+        self.telemetryDisabled = telemetryDisabled
+        self.applicationId = applicationId
+    }
 }
 
 /// Protocol for baseline options for individual client API calls.
--- a/sdk/core/AzureCore/Source/Pipeline/Policies/UserAgentPolicy.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Policies/UserAgentPolicy.swift
@@ -57,24 +57,24 @@ public class UserAgentPolicy: PipelineStage {
 
     // MARK: Initializers
 
-    public convenience init(for clazz: AnyClass, applicationId: String? = nil) {
+    public convenience init(for clazz: AnyClass, telemetryOptions: TelemetryOptions) {
         let libraryBundleInfo = DeviceProviders.bundleInfo(for: clazz)
         let sdkName = libraryBundleInfo?.name ?? ""
         let sdkVersion = libraryBundleInfo?.version ?? ""
-        self.init(sdkName: sdkName, sdkVersion: sdkVersion, applicationId: applicationId)
+        self.init(sdkName: sdkName, sdkVersion: sdkVersion, telemetryOptions: telemetryOptions)
     }
 
     public init(
         sdkName: String,
         sdkVersion: String,
-        applicationId: String? = nil,
+        telemetryOptions: TelemetryOptions,
         platformInfoProvider: PlatformInfoProvider? = DeviceProviders.platformInfo,
         appBundleInfoProvider: BundleInfoProvider? = DeviceProviders.appBundleInfo,
         localeInfoProvider: LocaleInfoProvider? = DeviceProviders.localeInfo
     ) {
         var userAgent = String(format: UserAgentPolicy.userAgentFormat, sdkName, sdkVersion)
 
-        let applicationId = applicationId ?? appBundleInfoProvider?.identifier
+        let applicationId = telemetryOptions.applicationId ?? appBundleInfoProvider?.identifier
         if var applicationId = applicationId, !applicationId.isEmpty {
             // From the design guidelines, applicationId must not contain a space
             if applicationId.rangeOfCharacter(from: .whitespacesAndNewlines) != nil {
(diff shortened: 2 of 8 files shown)


sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -32,6 +32,27 @@ public protocol AzureConfigurable {
...
+    /// Options for configuring telemetry sent by the service client.
+    var telemetryOptions: TelemetryOptions { get }
+}
+
+/// Options for configuring telemetry sent by the service client.
+public struct TelemetryOptions {
+    /// Whether platform information will be omitted from the user agent string sent by the service client.
+    public let telemetryDisabled: Bool
+    /// An optional user-specified application ID included in the user agent string sent by the service client.
+    public let applicationId: String?
```
tjprescott, line 44: Can this be defaulted to the bundle identifier?

sdk/core/AzureCore/Source/Pipeline/Policies/UserAgentPolicy.swift
```diff
@@ -57,24 +57,24 @@ public class UserAgentPolicy: PipelineStage {
 
     // MARK: Initializers
 
-    public convenience init(for clazz: AnyClass, applicationId: String? = nil) {
+    public convenience init(for clazz: AnyClass, telemetryOptions: TelemetryOptions) {
```
tjprescott, line 60: `clazz`? How did I miss this before?

tjprescott: Also, considering that the default of TelemetryOptions is all `nil`, why not make `telemetryOptions` optional here and simply instantiate a blank one if not specified? It would cut down on a lot of the other changes in this PR.

sdk/core/AzureCore/Tests/CollectionsTests.swift
```diff
@@ -60,7 +60,7 @@ class CollectionsTests: XCTestCase {
             baseUrl: URL(string: "http://www.microsoft.com")!,
             transport: URLSessionTransport(),
             policies: [
-                UserAgentPolicy(sdkName: "Test", sdkVersion: "1.0")
+                UserAgentPolicy(sdkName: "Test", sdkVersion: "1.0", telemetryOptions: TelemetryOptions())
```
tjprescott, line 63: Making `telemetryOptions` optional would eliminate the need to do this here and throughout.

sdk/storage/AzureStorageBlob/Source/Options/ClientOptions.swift
```diff
@@ -33,34 +33,57 @@ public struct StorageBlobClientOptions: AzureConfigurable {
...
+    public let restorationId: String
+
     /// The maximum size of a single chunk in a blob upload or download.
     public let maxChunkSizeInBytes: Int
 
+    /// The `TransferNetworkPolicy` to use for managed downloads.
+    public let downloadNetworkPolicy: TransferNetworkPolicy
+
+    /// The `TransferNetworkPolicy` to use for managed uploads.
+    public let uploadNetworkPolicy: TransferNetworkPolicy
```
tjprescott, line 56: Why are these public instead of internal?

sdk/core/AzureCore/Source/Pipeline/Policies/UserAgentPolicy.swift
bsiegel: Because `class` is typically a reserved keyword, "clazz" is a common name for such a parameter. However I could change it.

bsiegel: I considered making TelemetryOptions optional, but I liked making it required because it contains the very important flag to disable telemetry, and I felt like setting the default value would be the job of the client. Honestly there's no strong reason to go either way.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
bsiegel: We do this in the UserAgentPolicy, L77.

sdk/core/AzureCore/Tests/CollectionsTests.swift
bsiegel: It's true, but I felt it was important enough to make it a required parameter. Again, it's basically just a judgement call.

sdk/storage/AzureStorageBlob/Source/Options/ClientOptions.swift
bsiegel: Since the customer passes them in to the constructor of the `Options` object, it seems like they should be able to inspect them from the object. The `Options` objects are really just structs that hold onto a bunch of user-supplied options and contain no logic, so I feel like each option should be:

1. Provided to the initializer by the customer.
2. Have a default value in the initializer so it can be optional.
3. Be reflected in a public immutable property inside the struct.

I didn't see a reason why the policies should be an exception to this.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
bsiegel: I guess we could do here, but it seems better to do it in the policy since that avoids adding business logic to the options structure, and the policy has the various providers injected to its initializer already in order to decouple it from the operating system.

tjprescott: Sounds good to me.

sdk/core/AzureCore/Source/Pipeline/Policies/UserAgentPolicy.swift
tjprescott: I'm thinking mainly from a maintainability standpoint of having to add boilerplate telemetryOptions every time a test is created.

But on that note, what is the policy on the default setting for telemetry being enabled? In CLI it was a whole thing because of privacy. @adrianhall @mayurid is this something we need to worry about?

bsiegel: That's fair, I don't have a strong preference for this so I added a default value for the parameter in UserAgentPolicy.

bsiegel: For the telemetry setting, I looked at the other SDKs and they have a flag to "disable" telemetry, which is false by default. (Rather than say, having a flag to "enable" telemetry which is true by default.) So this should match the semantics and naming that we use in our other language SDKs.

bsiegel: I just recalled the specifics of why I did not make the TelemetryOptions parameter in the UserAgentPolicy initializer have a default value - it FORCES the clients to pass in a value when adding the policy to the pipeline, which will prevent the situation where a client accidentally forgets to pass the customer's TelemetryOptions and the default value is used instead (and since the default is that telemetry is enabled, we could accidentally be sending telemetry even if the customer disabled it which is VERY bad).

I added an extension to the test project which provides a convenience initializer that lets the tests omit the telemetryOptions parameter instead, this should resolve the concern while still maintaining the requirement for clients to always provide a TelemetryOptions.

tjprescott:
Needs the latest from master, but otherwise LGTM.

tjprescott:
One thing though is that this doesn't close #150 because there are two other subitems in that issue.

PR #358, opened 2020-07-16
Add sample token vendor Azure Function

bsiegel:
Completes #160


examples/AzureFunctionSASTokenVendor/GetSasToken/__init__.py
```diff
@@ -0,0 +1,50 @@
+import datetime, json, os
+
+import azure.functions as func
+import azure.storage.blob as blob
+
+account_name = os.environ['AZURE_STORAGE_ACCOUNT_NAME']
+account_key = os.environ['AZURE_STORAGE_ACCOUNT_KEY']
+container_name = os.environ['AZURE_STORAGE_CONTAINER_NAME']
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
```
tjprescott, line 10: So this is the only Python-specific code needed for the Azure Function? @adrianhall any feedback on this file?

bsiegel: Yep, everything else in this was generated by the VSCode extension when I used it to create a new Python-based Azure Function project. I also had to update requirements.txt and and I force-added the local.settings.json file (normally excluded by gitignore) to include an example of the three environment variables the function needs to access.

bsiegel:
> Have you tested this with the demo app for upload and download?

I haven't yet, no. I have verified that the function is correctly returning valid SAS tokens, but I haven't tested the client changes to make sure the Credential can pull them correctly. I've got to figure out how to get the app (running on the simulator) to access the Azure Function test server running on my local machine, but once I do that it should be easy to validate.

Wanted to get your feedback on the new credential from a feature / architectural standpoint as well since the new semantics are quite different: every time the client issues a request, the SASTokenCredential is asked for credentials corresponding to the URL being accessed. It will be the customer's job to implement a caching strategy for these credentials currently if they want to avoid spamming their backend with every request, which isn't exactly trivial. I considered trying to build one, but it's difficult because the customer might return say, an account- or container-level credential, which would be valid for every blob within that account/container, or they might return a credential that is just valid for the one blob being accessed - we wouldn't know which type they are returning so we couldn't make a smart decision about how to cache it.

tjprescott:
> Wanted to get your feedback on the new credential from a feature / architectural standpoint as well since the new semantics are quite different: every time the client issues a request, the SASTokenCredential is asked for credentials corresponding to the URL being accessed. It will be the customer's job to implement a caching strategy for these credentials currently if they want to avoid spamming their backend with every request, which isn't exactly trivial. I considered trying to build one, but it's difficult because the customer might return say, an account- or container-level credential, which would be valid for every blob within that account/container, or they might return a credential that is just valid for the one blob being accessed - we wouldn't know which type they are returning so we couldn't make a smart decision about how to cache it.

Since I suppose the scenario is that there would be a different SAS token for each file, I think that architecture makes sense.

Generally, I'm not a fan of letting customers figure it out, especially if it's not trivial. I feel like we should at least have some implementation for the example, even if it doesn't fit their exact scenario. @adrianhall do you have any input here?

Should we discuss this at tomorrow's meeting? It will be very relevant to Android as well.

bsiegel:
> Should we discuss this at tomorrow's meeting? It will be very relevant to Android as well.

Yes, definitely. There's a thread going on in @vcolin7's identity PR that is basically the same discussion: https://github.com/Azure/azure-sdk-for-android/pull/238#discussion_r457806326

tjprescott:
@bsiegel is this awaiting any additional work or is it ready to go?

bsiegel:
@tjprescott I'm carving out time this evening to do an end-to-end test with this to make sure it all works. Assuming that's all good I will merge it tonite.

tjprescott:
@bsiegel what's the word on this? Ready to merge?

bsiegel:
Ran through the end-to-end test here today, I found two small things that needed to be fixed in the sample code for consuming the token within the iOS application. Otherwise everything worked as expected - once all CI passes I will merge.

PR #361, opened 2020-07-17
Remove AzureIdentity from Package.swift.

mitchdenny:
This PR removes AzureIdentity from the Package.swift file. This is because we won't initially have the MSAL dependency hooked up in the shipping package so we don't need it (although we can still use it for development purposes and building out the demo app).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,7 +36,6 @@ let package = Package(
     ],
     products: [
         .library(name: "AzureCore", targets: ["AzureCore"]),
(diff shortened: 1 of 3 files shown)


bsiegel:
This is good, but we also need to update the places we have conditional-compile statements, and I think it makes sense to do in this PR:

1. Anywhere we have `#if canImport(MSAL)` in AzureIdentity, remove it
2. Anywhere we have `#if canImport(AzureIdentity) && canImport(MSAL)` in Storage, change it to just `#if canImport(AzureIdentity)`

PR #381, opened 2020-08-03
JSON Patch Experiment

tjprescott:
This PR has some draft code for support PATCH operations. It uses a sentinel value `JSONPatch.null` to resolve the difference between a do-nothing nil default and the act of actually setting something equal to nil. Since the sentinel is string-based, there will need to be some way of overriding this to provide the sentinel string as a literal.


sdk/core/AzureCore/Tests/JSONPatchTests.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// Represents a simple method with a simple, optional type.
+    func testMethod(value: String? = nil) -> String? {
+        let patch = JSONPatchObject()
+        if let val = value {
+            patch.replace(atPath: "value/", withValue: val)
+        }
+        let encoder = JSONEncoder()
+        let encoded = try! encoder.encode(patch)
+        return String(data: encoded, encoding: .utf8)
+    }
```
tjprescott, line 43: This is a prototype of what generated body code for a PATCH operation might look like (except the encoding part)

sdk/core/AzureCore/Tests/JSONPatchTests.swift
```diff
@@ -0,0 +1,77 @@
...
+            "{\"operations\":[{\"path\":\"value\\/\",\"operation\":\"replace\",\"value\":null,\"from\":null}]}"
+        XCTAssertEqual(patch!, expected)
+    }
+
+    func test_JSONPatch_EscapedSentinelValueWorks() {
+        let patch = testMethod(value: "__NULL__")
+        let expected =
+            "{\"operations\":[{\"path\":\"value\\/\",\"operation\":\"replace\",\"value\":\"\\\"__NULL__\\\"\",\"from\":null}]}"
+        XCTAssertEqual(patch!, expected)
+    }
```
tjprescott, line 76: This test fails, but you should be able to set a value equal to the sentinel raw value.

sdk/core/AzureCore/Source/Util/JSONPatchUtil.swift
```diff
@@ -0,0 +1,117 @@
...
+
+//    /// Inserts a new value at an array index or adds a new property.
+//    public func add<T: Codable>(atPath path: String, withValue value: T) {
+//        operations.append(PatchOperation(operation: .add, from: nil, path: path, value: value))
+//    }
+//
+//    /// Remove the property or entry at an array index. Property must exist.
+//    public func remove(atPath path: String) {
+//        operations.append(PatchOperation(operation: .remove, from: nil, path: path, data: nil))
+//    }
```
tjprescott, line 80: These are comment out for now since the only operation being used right now is `replace`.

PR #383, opened 2020-08-10
Fixed broken links in ios

sima-zhu:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -56,6 +56,10 @@ jobs:
       - script: sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app
         displayName: 'Use Xcode $(XcodeVersion)'
 
+      - template: ../../../common/pipelines/templates/steps/verify-links.yml
+        parameters:
+          Directory: sdk/${{ parameters.ServiceDirectory }}
```
weshaggard, line 61: @mitchdenny @bsiegel correct me if I'm wrong but I don't think there are service directories in the ios repo. I thought there is only one large client build. In which case we should just run the link checker on the entire repo.

eng/ignore-links.txt
```diff
@@ -0,0 +1,2 @@
+#Place holder for the temporary links
+not-valid-link
```
weshaggard, line 2: I would remove "not-valid-link" and just leave the comment `# Placeholder file for links that need exceptions from link checking`

sima-zhu: We have some #note-valid-link in README-TEMPLATE.md which failed the scripts. How do we plan to suppress the check if remove here.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
bsiegel: We do have service directories, but currently we have a single pipeline that builds all libraries. It can easily be swapped between the two options by simply changing the scheme name (from "all" to the name of the service directory) when calling xcodebuild.

That said, yes I'd support running the link checker on the entire repo during each build.

eng/ignore-links.txt
weshaggard: Ok we should remove that link from the template.md then and just add a comment, see an example in https://github.com/Azure/azure-sdk/pull/1587/files.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
sima-zhu: Will check the entire repo here in new PR.

PR #415, opened 2020-09-02
Permit client method callback on background threads.

### LINKED ISSUE #413: syncIterator stops prematurely
Assuming a paged response, the following should work:
```Swift
// get all results
let iterator = paged.syncIterator()
while iterator.next != nil {}
print(paged.underestimatedCount)
print(paged.isExhausted)
```

This should output the total number of items, and `isExhausted` should be true (indicating there are no more pages to retrieve). Instead, this code will iterate through each item of the first page and then output the number of items in the first page and that isExhausted is false.

### LINKED ISSUE #414: Cannot make client calls on a background dispatch queue
Attempting to run an operation on a background DispatchQueue hangs because the client always calls the completionHandler on `DispatchQueue.main`. This is a convenience to the app developer who will often be processing the results of an operation on the main thread. We should support passing in a DispatchQueue to enable background processing.

tjprescott:
Fixes #414. Fixes #413.


sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -36,6 +36,8 @@ public protocol AzureClientOptions {
     var telemetryOptions: TelemetryOptions { get }
     /// Global transport options
     var transportOptions: TransportOptions { get }
+    /// The default dispatch queue on which to call all completion handlers. Defaults to `DispatchQueue.main`.
+    var dispatchQueue: DispatchQueue? { get }
```
bsiegel, line 40: This doesn't seem to be true. I only see the `dispatchQueue` property being used for the PagedCollection (where it defaults to `.utility`), and the other client operations don't appear to use the dispatchQueue at all anymore.

bsiegel:
I'm unclear how or why this fixes #413 ??

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
tjprescott: The generated client code in autorest.swift uses:
```Swift
let dispatchQueue = self.commonOptions.dispatchQueue ?? DispatchQueue.main
dispatchQueue.async { 
  stuff
}
```

I removed any references to dispatch queues inside the collections file because those calls were not working. They were hanging even though `commonOptions.dispatchQueue` was set to the right value (and thus these were being called on the right queue). Removing any reference to the dispatchQueue fixed the issue and still permits the final completion handler call on the desired queue.

tjprescott:
This fixes #413 because that issue was caused by async callbacks being swallowed and ignored. With these changes, it correctly iterates through all the pages until the collection is exhausted, at which point it returns `nil`. There was no specific error in the iterator logic--it all had to do with the dispatch queue issues.

The tests work with these changes for our in-progress SDK (and hang with AzureCore as-is) and I've check the Storage demo  with these changes to ensure it still works (which it does).

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
tjprescott: Here's the autorest.swift PR link that shows how this change is translating to generated code:
https://github.com/Azure/autorest.swift/pull/177

sdk/core/AzureCore/Source/DataStructures/Collections.swift
```diff
@@ -246,16 +246,12 @@ public class PagedCollection<SingleElement: Codable> {
                 }
             }
             if let returnError = returnError {
-                DispatchQueue.main.async {
-                    completionHandler(.failure(returnError))
-                }
+                completionHandler(.failure(returnError))
```
sacheu, line 249: Should the completionHandler be called inside the `queue` (either the DispatchQueue.main or the queue in the ClientOption)?

tjprescott: This caused the demo test to hang, even when called on the right dispatch queue. Removing this entirely caused the demo test to work correctly and still works correctly with the AzureSwiftSDKDemo.

tjprescott:
@sacheu @bsiegel ping

sdk/core/AzureCore/Source/DataStructures/Collections.swift
sacheu: Thanks. Do you know why removing this will cause the demo app to work, but keeping it inside the queue failed? Maybe add comments to explain why calling completionHandler out of DispatchQueue will work, but keeping them inside does not work.

sacheu (approved):

PR #445, opened 2020-09-21
ContentDecodePolicy should pass through data to client

### LINKED ISSUE #444: AzureCore ContentDecodePolicy should pass through data and let the client decode it
ContentDecodePolicy always try to decode the http response using JSONSerialization to deserialize http response body.
But JSONSerialization can't handle data which is not a JSON object.
49 (2 bytes), null (4 bytes), "foo" (5 bytes) are part of the JSON spec is valid JSON so ContentDecodePolicy should be passing the data to client instead of throw an error.

Solution is to pass through the data if JSONSerialization can't decode the data.

tjprescott:
Fixes #444.

Previously, we would find content type JSON and do a round-trip deserialized/serialize loop to validate that the data was a valid JSON **object**. However, content type `application/json` is any valid JSON, not just JSON objects, which will cause this check to fail for cases within the autorest testserver since there are bodies that are valid JSON but not JSON objects.

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
@@ -231,7 +231,7 @@ public class ContentDecodePolicy: PipelineStage {
     internal func deserialize(from response: PipelineResponse, contentType: String) throws -> AnyObject? {
         guard let data = response.httpResponse?.data else { return nil }
         if jsonRegex.hasMatch(in: contentType) {
-            return try JSONSerialization.jsonObject(with: data, options: []) as AnyObject
+            return data as AnyObject
         } else if contentType.contains("xml") {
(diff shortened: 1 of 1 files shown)


bsiegel:
This looks good to me. One question - what is the developer experience if the client receives invalid JSON? Currently it would throw due to the validation step, what would happen with this change?

tjprescott:
> This looks good to me. One question - what is the developer experience if the client receives invalid JSON? Currently it would throw due to the validation step, what would happen with this change?

@bsiegel basically it defers the validation to the client. With the autorest generated client, they will still get an error, but the code generator has more information to decode the correct type and only throw if that can't be decoded. 

Currently, that logic would all work just fine, but it gets pre-empted in this policy and instead throws a DecodingError when it should succeed.

bsiegel (approved):
LGTM, build failure is unrelated - seems like we need to update a link in a README.

PR #446, opened 2020-09-22
Add generated code

sacheu:


Package.swift
```diff
@@ -35,7 +35,10 @@ let package = Package(
         .macOS(.v10_15), .iOS(.v12), .tvOS(.v12)
     ],
     products: [
-        .library(name: "AzureCore", targets: ["AzureCore"])
+        .library(name: "AzureCore", targets: ["AzureCore"]),
```
raosanat, line 38: Can we remove the library here that ends with "Dynamic" and keep the rest as it is

tjprescott (approved):

RezaJooyandeh (approved):

Package.swift
sacheu: We have opened PR # 447 to address this comment.

PR #447, opened 2020-09-22
Remove dynamic in Package.swift

sacheu:


Package.swift
```diff
@@ -37,8 +37,6 @@ let package = Package(
     products: [
         .library(name: "AzureCore", targets: ["AzureCore"]),
```
raosanat, line 38: this needs to be dynamic as well ?

tjprescott (changes requested):
Awaiting approval from @raosanat.

Package.swift
sacheu: Sanath is testing without making this dynamic.

tjprescott (approved):

Package.swift
sacheu: It is removed.

PR #457, opened 2020-09-29
[CommunicationIdentifier] Align identifier types in Swift with other languages

JoshuaLai:
Removing the `getId()` method to align with other sdk's

### DIFF
--- a/sdk/communication/AzureCommunication/Source/Identifiers.swift
+++ b/sdk/communication/AzureCommunication/Source/Identifiers.swift
@@ -26,16 +26,10 @@
 
 import Foundation
 
-@objc public protocol CommunicationIdentifier: NSObjectProtocol {
-    func getId() -> String
-}
+@objc public protocol CommunicationIdentifier: NSObjectProtocol {  }
 
 @objcMembers public class CommunicationUser: NSObject, CommunicationIdentifier {
     public let identifier: String
-
-    public func getId() -> String {
-        return self.identifier
-    }
 
     public init(identifier: String) {
         self.identifier = identifier
@@ -45,10 +39,6 @@ import Foundation
 @objcMembers public class CallingApplication: NSObject, CommunicationIdentifier {
     public let identifier: String
 
-    public func getId() -> String {
-        return self.identifier
-    }
-
     public init(identifier: String) {
         self.identifier = identifier
     }
@@ -57,10 +47,6 @@ import Foundation
 @objcMembers public class UnknownIdentifier: NSObject, CommunicationIdentifier {
     public let identifier: String
 
-    public func getId() -> String {
-        return self.identifier
-    }
-
     public init(identifier: String) {
         self.identifier = identifier
     }
@@ -68,10 +54,6 @@ import Foundation
 
 @objcMembers public class PhoneNumber: NSObject, CommunicationIdentifier {
     public let value: String
-
-    public func getId() -> String {
-        return self.value
-    }
 
     public init(phoneNumber: String) {
         self.value = phoneNumber


sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -26,17 +26,11 @@
 
 import Foundation
 
-@objc public protocol CommunicationIdentifier: NSObjectProtocol {
-    func getId() -> String
-}
+@objc public protocol CommunicationIdentifier: NSObjectProtocol {  }
 
 @objcMembers public class CommunicationUser: NSObject, CommunicationIdentifier {
     public let identifier: String
```
DominikMe, line 32: Can we rename the field to `id` so that it matches Java and C#? Same for the other types.

```suggestion
    public let id: String
```

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -26,17 +26,11 @@
...
 
-    public func getId() -> String {
-        return self.identifier
-    }
-
     public init(identifier: String) {
         self.identifier = identifier
```
DominikMe, line 35:
```suggestion
        self.id= identifier
```

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -45,10 +39,6 @@ import Foundation
 @objcMembers public class CallingApplication: NSObject, CommunicationIdentifier {
     public let identifier: String
```
DominikMe, line 40:
```suggestion
    public let id: String
```

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -45,10 +39,6 @@ import Foundation
...
 
-    public func getId() -> String {
-        return self.identifier
-    }
-
     public init(identifier: String) {
         self.identifier = identifier
```
DominikMe, line 43:
```suggestion
        self.id= identifier
```

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -57,10 +47,6 @@ import Foundation
 @objcMembers public class UnknownIdentifier: NSObject, CommunicationIdentifier {
     public let identifier: String
```
DominikMe, line 48:
```suggestion
    public let id: String
```

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -57,10 +47,6 @@ import Foundation
...
 
-    public func getId() -> String {
-        return self.identifier
-    }
-
     public init(identifier: String) {
         self.identifier = identifier
```
DominikMe, line 51:
```suggestion
        self.id= identifier
```

DominikMe (changes requested):
Thanks for the quick change!

Let's align this even more by renaming to `id` where applicable. We should double check if our docs need to change after this.

sdk/communication/AzureCommunication/Source/Identifiers.swift
JoshuaLai: In Objective-C `id` is considered an actual type. This may raise some confusion.

DominikMe: I didn't know, let's keep `identifier` then.

(Talked offline, just commenting here for transparency :-) )

DominikMe (approved):

RezaJooyandeh (approved):

raosanat (approved):

raosanat (approved):
I have changed the implementation in the SDK, please merge this so that i can commit my changes to master as well

raosanat:
@mariusu-msft  this was the PR for the change of removing `getID()` from `CommunicationIdentifier` 
As i mentioned previously this change is not ideal as we have to now first know what kind of a class it is  , cast it and then read the member variable `identifier`. But i was told that this was already cleared by Azure review board and have to remove `getId()`

mariusu-msft:
@RezaJooyandeh @JoshuaLai @tjprescott i get the desire here to align, but this seems to be making the developer experience more complicated.

getId works as an interface for all Identifier types.

Is there any other requirement driving this change?

RezaJooyandeh:
@mariusu-msft We did consider the id as a shared method. But id of all types are not supposed to be exposed hence we did not had it in the design and the previous `getId()` was not actually returning the id on every type.

These types as we had originally implemented did not suppose to have `getId` (https://github.com/Azure/azure-sdk-for-ios-pr/pull/10) and the addition of `getId` seems to have happen by mistake against the decision that we had in the arch board and internally (https://github.com/Azure/azure-sdk-for-ios-pr/pull/15/commits/a7462b38789b8a65b681501b1a9e506b2deecce0). We spotted this mismatch right before the first release, but we kept it to avoid last minute breakage in the Calling package, but now we are aligning back with the collective decision.

PR #471, opened 2020-10-05
Bump SDK version to 1.0.0-beta.2 and update / improve readmes

bsiegel:


README.md
```diff
@@ -70,41 +75,11 @@ Next, add each client library you wish to use in a target to the target's array
         ...
         .target(
             name: "MyTarget",
```
mitchdenny, line 77: Is this name correct?

mitchdenny (approved):

eng/ignore-links.txt
```diff
@@ -1 +1,2 @@
-# Placeholder file for links that need exceptions from link checking
\ No newline at end of file
+# Placeholder file for links that need exceptions from link checking
+https://github.com/Azure/azure-sdk-for-ios/releases/tag/1.0.0-beta.2
```
weshaggard, line 2: I'd like us to try and avoid having links in the repo that aren't valid even if they aren't valid even temporarily. We should instead try and stick with links that exist already.

PR #479, opened 2020-10-07
Add flag in PipelineClient to skip url encoding

sacheu:
For url.json test swagger, we need an optional flag to skip url encoding in PipelineClient

### DIFF
--- a/sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
+++ b/sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
@@ -59,6 +59,7 @@ open class PipelineClient {
     public var baseUrl: URL
     public var logger: ClientLogger
     public var commonOptions: AzureClientOptions
+    public var skipUrlEncoding: Bool
 
     // MARK: Initializers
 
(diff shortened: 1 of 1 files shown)


sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -59,6 +59,7 @@ open class PipelineClient {
     public var baseUrl: URL
     public var logger: ClientLogger
     public var commonOptions: AzureClientOptions
+    public var skipUrlEncoding: Bool
```
bsiegel, line 62: Haven't dug into this deeply, but should this not be in the commonOptions?

bsiegel, line 62: Should we perhaps have an "internal" options object that is not exposed to customers, for things like this? I can imagine more flags being needed in the future and we don't want to explode the parameter list.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
```diff
@@ -59,6 +59,7 @@ open class PipelineClient {
     public var baseUrl: URL
     public var logger: ClientLogger
```
bsiegel, line 60: Unrelated, but I just noticed we accept a logger here, but the 1commonOptions` also contains a logger. Maybe we can clean this up.

sdk/core/AzureCore/Source/Pipeline/PipelineClient.swift
sacheu: Thanks for your review Brandon. After further discussion with Travis, we are changing our solution. Changes in this PR is not needed.

sacheu:
Change solution. Changes in the PR is not needed.

PR #481, opened 2020-10-08
[HMACAuthenticationPolicy] Work in progress of the HMACAuthentication policy

JoshuaLai:
Set up the new HMACAuthenticationPolicy and adding unit tests to test the encryption is working like the other platforms.

### DIFF
--- a/sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
+++ b/sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
@@ -29,16 +29,23 @@ import AzureCore
 #endif
 
 import Foundation
-import CommonCrypto
 import CryptoKit
 
-@objcMembers public class HMACAuthenticationPolicy: Authenticating {
+public class HMACAuthenticationPolicy: Authenticating {
     public var next: PipelineStage?
     private let accessKey: String
     
     static let dateHeader = "date"
     static let hostHeader  = "host"
     static let contentHashHeader = "x-ms-content-sha256"
+
+    struct HMACAuthenticationProperties {
+        let url: URL
+        let httpMethod: HTTPMethod
+        let contents: Data
+        let date: Date
+    }
+
     
     public init(accessKey: String) {
         self.accessKey = accessKey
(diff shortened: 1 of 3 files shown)


sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+
+#if canImport(AzureCore)
+import AzureCore
+#endif
+
+import Foundation
+import CommonCrypto
+import CryptoKit
+
+@objcMembers public class HMACAuthenticationPolicy: Authenticating {
```
sacheu, line 35: Do we need to export this to Objective C?

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+    static let contentHashHeader = "x-ms-content-sha256"
+    
+    public init(accessKey: String) {
+        self.accessKey = accessKey
+    }
+
+    public func authenticate(
+        request: PipelineRequest,
+        completionHandler: @escaping OnRequestCompletionHandler) {
+        let contents = request.httpRequest.data ?? Data() // Is this the body of the request?
```
sacheu, line 50: this can be moved to after the guard

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+    
+    public func addAuthenticationHeaders(
+        url: URL,
+        httpMethod: String,
+        contents: Data) -> HTTPHeaders {
+        var headers: HTTPHeaders = [:]
+        headers[HMACAuthenticationPolicy.contentHashHeader] = contents.sha256
+        
+        let dateFormatter = DateFormatter()
```
sacheu, line 74: There is date formatter in AzureCore

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            return self.hash(algorithm: .sha256).base64EncodedString()
+        }
+    }
+
+}
+
+extension String {
+    func generateHmac(using secret: String) -> String {
+        if #available(iOS 13.0, *) {
```
sacheu, line 136: will this 2 approaches produce the same Hmac values? if yes, can we just do `self.hmac(algorithm: .sha256, key: Data(base64Encoded: secret)!).base64EncodedString()`

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+        url: URL,
+        httpMethod: String,
+        date: String,
+        contentHashed: String) -> HTTPHeaders {
+        // Order of the headers are important here for generating correct signature
+        let signedHeaderNames = "\(HMACAuthenticationPolicy.dateHeader);\(HMACAuthenticationPolicy.hostHeader);\(HMACAuthenticationPolicy.contentHashHeader)"
+        let signedHeaderValues =  "\(date);\(url.host ?? "");\(contentHashed)" // date;urlhost;content hash
+        
+        // Add unit test for different ports
+        // 1 with port 1 without port
```
sacheu, line 109: does `port` include in the url.path?

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+        headers[HMACAuthenticationPolicy.contentHashHeader] = contents.sha256
+        
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "E, dd MMM YYYY HH:mm:ss 'GMT'" // Is this the right date format? Try ISO8601DateFormatter
+        dateFormatter.timeZone = TimeZone(abbreviation: "UTC")
+        
+        let date = Date()
+        let utcNow = dateFormatter.string(from: date)
+        headers[HMACAuthenticationPolicy.dateHeader] = utcNow
+        headers[HMACAuthenticationPolicy.hostHeader] = url.host
```
sacheu, line 81: can these be set in addSignatureHeader() as well? so that we don't  need to do `headers.merge`?

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
JoshuaLai: I thought I saw the other policies export to Obj-C but it looks like I was mistaken so I will remove.

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
JoshuaLai: No. Port is returned under `url.port`. This was a suggestion from Dominik to write unit tests in the case different ports are used so I will have those in an update

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
JoshuaLai: Both of them return the same thing. The reason I wanted to leave both implementations is I wanted to have a conversation about which one to use. Within the if is an implementation that uses `CryptoKit` whereas the bottom one uses `CommonCrypto` I know as swift continues to mature Apple is looking for developers to use `CryptoKit` more.

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
JoshuaLai: Yes, but this feels a little backwards, would it make more sense to get rid of the `addSignatureHeader` method and add its implementation in here instead?

sdk/communication/AzureCommunication/Source/Authentication/HMACAuthenticationPolicy.swift
```diff
@@ -0,0 +1,144 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+#if canImport(AzureCore)
+import AzureCore
+#endif
+
+import Foundation
+import CommonCrypto
```
JoshuaLai, line 32: I should be able to remove this import

JoshuaLai:
After conversations with @RezaJooyandeh we don't think there is a need for this policy. 
So closing this PR

PR #496, opened 2020-10-19
[CommunicationAccessToken] Communication created version of access token 

JoshuaLai:
Currently, AzureCommunication depends on AzureCore's AccessToken. Chat is built in obj-c so requires an obj-c members. For chat to use AzureCommunication properly we need to create our own obj-c class access token. At the same time the Access token in Core can be changed back to a pure struct.

### DIFF
--- a/sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
+++ b/sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
@@ -25,8 +25,13 @@
 // --------------------------------------------------------------------------
 
 import Foundation
-
-public struct CommunicationAccessToken {
+// This is class
+@objcMembers public class CommunicationAccessToken: NSObject {
     let token: String
     let expiresOn: Date
+    
+    public init(token: String, expiresOn: Date) {
+        self.token = token
+        self.expiresOn = expiresOn
+    }
 }
--- a/sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
+++ b/sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
@@ -24,37 +24,9 @@
 //
 // --------------------------------------------------------------------------
 
-#if canImport(AzureCore)
-import AzureCore
-#endif
 import Foundation
-/**
- The Azure Communication Services User token credential. 
- */
-internal class StaticUserCredential: CommunicationTokenCredential {
-    private let accessToken: CommunicationAccessToken
 
-    /**
-     Creates a static `CommunicationUserCredential` object from the provided token.
-        
-     - Parameter token: The static token to use for authenticating all requests.
-     
-     - Throws: `AzureError` if the provided token is not a valid token.
-     
-     - SeeAlso: ` CommunicationUserCredential.init(...)`
-     */
-    public init(token: String) throws {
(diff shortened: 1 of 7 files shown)


sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
```diff
@@ -0,0 +1,37 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+// This is class
```
sacheu, line 28: remove this comment.

sdk/communication/AzureCommunication/Source/Authentication/CommunicationUserCredential.swift
```diff
@@ -29,13 +29,13 @@ import AzureCore
...
 
-public typealias AccessTokenRefreshOnCompletion = (AccessToken?, Error?) -> Void
+public typealias AccessTokenRefreshOnCompletion = (CommunicationAccessToken?, Error?) -> Void
 public typealias TokenRefreshOnCompletion = (String?, Error?) -> Void
 
 /**
  The Azure Communication Services User token credential. This class is used to cache/refresh the access token required by Azure Communication Services.
 */
-@objcMembers public class CommunicationUserCredential: NSObject {
+public class CommunicationUserCredential: NSObject {
```
sacheu, line 38: Why 'objcMembers' is removed?  i think this is needed for callingSDK to use this class.

JoshuaLai: Yes sorry this was done by mistake

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
JoshuaLai: Removed

tjprescott (commented):
This looks good to me. Have you verified that the example app continues to compile with this change?

RezaJooyandeh:
Is there any problem with keeping the annotation for AccessToken in Azure.Core? It is not scalable to duplicate Azure.Core types and I think it will lead us towards being less align with the Azure SDKs.

bsiegel:
> Is there any problem with keeping the annotation for AccessToken in Azure.Core? It is not scalable to duplicate Azure.Core types and I think it will lead us towards being less align with the Azure SDKs.

We definitely understand the scalability issue, but unfortunately adapting Azure Core to be usable by ObjC libraries is not possible without significant re-engineering. We've been having discussions to determine the best way forward for this scenario, but we don't have a general solution yet, so we definitely want to revert this change for now.

JoshuaLai:
> This looks good to me. Have you verified that the example app continues to compile with this change?
@tjprescott: I have been working with @raosanat and @sacheu to make sure their sample apps are still working

tjprescott (approved):

PR #503, opened 2020-10-20
[CommunciationUserCredentialTests] Adding new obj-c unit tests

JoshuaLai:
Create a ObjC unit test against CommunicationSDK to prevent future breaks due to ObjC/Azure Communication.


JoshuaLai:
@bsiegel I had to do a few things to get all the unit tests to pass in Obj-C. It seems like the unit tests themselves are having trouble communicating with the simulator, so in the script I now boot up the simulator first to give a little extra time to connect. The second thing I did to help I noticed running too many unit tests in parallel seemed to throw code 65 as well so I broke it up into two classes. The third (probably more questionable) I turned off all the parallel runs for all the targets, this makes the entire suite run a little slower but the theory was in case one test suite interfered with another one.

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
```diff
@@ -27,8 +27,8 @@
 import Foundation
 
 @objcMembers public class CommunicationAccessToken: NSObject {
-    let token: String
-    let expiresOn: Date
+    public let token: String
```
sacheu, line 30: This change has been merged right? I am curious why it shows up as a diff in this PR?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+    NSString *token = [self generateTokenValidForMinutes: 1];
+    CommunicationUserCredential *credential = [[CommunicationUserCredential alloc]
+                                               initWithInitialToken:token
+                                               refreshProactively:YES
+                                               error:nil
+                                               tokenRefresher:
+                                               ^(void (^ block)
+                                                 (NSString * _Nullable newToken,
+                                                  NSError * _Nullable error)) {
+        weakSelf.fetchTokenCallCount += 1;
```
sacheu, line 93: Do we have assert to validate the value of `weakSelf.fetchTokenCallCount` ?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+    NSString *token = [self generateTokenValidForMinutes: 9];
+    CommunicationUserCredential *credential = [[CommunicationUserCredential alloc]
+                                               initWithInitialToken:token
+                                               refreshProactively:YES
+                                               error:nil
+                                               tokenRefresher:
+                                               ^(void (^ block)
+                                                 (NSString * _Nullable newToken,
+                                                  NSError * _Nullable error)) {
+        weakSelf.fetchTokenCallCount += 1;
```
sacheu, line 123: Do we have assert to validate the value of weakSelf.fetchTokenCallCount ?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+                                             NSError * _Nullable error) {
+        XCTAssertNotNil(accessToken);
+        XCTAssertNil(error);
+        XCTAssertEqual(accessToken.token, weakSelf.sampleToken);
+        XCTAssertEqual(accessToken.expiresOn.timeIntervalSince1970, weakSelf.sampleTokenExpiry);
+        
+        [expectation fulfill];
+    }];
+    
+    [self waitForExpectations:@[expectation] timeout:5.0];
```
sacheu, line 76: Why some test cases use 
`[self waitForExpectations:@[expec`tation] timeout:5.0]; `
and other use 
`block(weakSelf.sampleToken, nil);`

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+    [credential tokenWithCompletionHandler:^(CommunicationAccessToken * _Nullable accessToken,
+                                             NSError * _Nullable error) {
+        XCTAssertNotNil(accessToken);
+        XCTAssertEqual(accessToken.token, weakSelf.sampleToken);
+        XCTAssertEqual(weakSelf.fetchTokenCallCount, 1);
+        
+        [expectation fulfill];
+    }];
+
+    [self waitForExpectations:@[expectation] timeout:2.0];
```
sacheu, line 106: why some test cases use 5 and others use 2 for timeout values?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+    dateFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ssZ";
+    
+    NSDate *rightNow = [[dateFormatter dateFromString:d]
+                        dateByAddingTimeInterval:(60 * minutes)];
+    NSTimeInterval timeInterval = [rightNow timeIntervalSince1970];
+    NSString *tokenString = [NSString stringWithFormat:@"{\"exp\":%f}", timeInterval];
+    NSData *tokenStringData = [tokenString dataUsingEncoding: NSASCIIStringEncoding];
+
+    NSString *validToken = [NSString stringWithFormat:@"%@.%@.%@",
+                            @"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9",
```
sacheu, line 152: should we move these constants to the list of constants in the setup method?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
```diff
@@ -0,0 +1,159 @@
...
+                            [tokenStringData base64EncodedStringWithOptions:NSDataBase64Encoding76CharacterLineLength],
+                            @"adM-ddBZZlQ1WlN3pdPBOF5G4Wh9iZpxNP_fSvpF4cWs"];
```
sacheu, line 154: should we move these constants to the list of constants in the setup method?

sacheu (changes requested):
Looks good. Have a few comments

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAccessToken.swift
JoshuaLai: Yes I merged in that change. This should be resolved when I update the fork

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
JoshuaLai: Yes this is being asserted in line 101 in the completion block

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
JoshuaLai: We can change them all to 2 I was previously following the the swift tests. But I noticed that some async calls were doing dispatch after. Since I removed the that little logic we can probably just have all the time outs be 2 seconds

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
JoshuaLai: Yes this is being asserted in line 131

sdk/communication/AzureCommunication/Tests/ObjCCommunicationUserCredentialAsyncTests.m
JoshuaLai: Good idea1

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -108,6 +108,8 @@ jobs:
 
       - script: |
           set -o pipefail
+          xcrun simctl shutdown all
```
bsiegel, line 111: Did we determine if these are necessary to make the tests work? In theory they shouldn't since we should always get a fresh VM / simulator every run, so if we don't need them we should back this change out.

bsiegel (approved):
Looks good but if we can back out that one change to the CI configuration that would be ideal.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JoshuaLai: I think you are right so I will remove those now

sacheu (approved):

PR #510, opened 2020-10-21
ArchBoard fixes and refactors

tjprescott:
Small, easy fixes and refactors from the arch board review.

Will have implications on Autorest.Swift

Renames:

AzureError.sdk -> AzureError.client
AzureOptions -> RequestOptions
AzureClientOptions -> ClientOptions
AzureCredential -> Credential

cc/ @sacheu


sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
```diff
@@ -116,7 +116,7 @@ public final class AzureCommunicationChatClient: PipelineClient, PageableClient
         self.request(request, context: context) { result, httpResponse in
             let dispatchQueue = options?.dispatchQueue ?? self.commonOptions.dispatchQueue ?? DispatchQueue.main
             guard let data = httpResponse?.data else {
-                let noDataError = AzureError.sdk("Response data expected but not found.")
+                let noDataError = AzureError.client("Response data expected but not found.")
```
simokhwang, line 119: Since the changes under sdk/communication are generated, should we fix the Autorest.swift code generator first and then republished the generated code here?

simokhwang (changes requested):
I think we need to make corresponding changes in AutoRest swift code generator

sacheu (changes requested):
I think we need to make the corresponding change to Autorest swift code generator

sacheu (approved):

bsiegel (approved):
Looks good to me, let's make sure we encode these changes in the guidelines.

bsiegel:
Oh, let's make sure to capture (either here in this PR or a separate issue/PR) migrating from Double to TimeInterval for durations (and removing the suffix - TimeInterval is always in seconds).

PR #511, opened 2020-10-21
Split AzureSDK.podspec and add AzureCommunicationCalling.podspec

raosanat:
Split AzureSDK.podspec into individual podspecs and add new podspec for AzureCommunicationCalling


AzureCommunicationCalling.podspec
```diff
@@ -25,40 +25,27 @@
...
+  
+    s.ios.deployment_target = '12.0'
+    s.ios.vendored_frameworks = "AzureCommunicationCalling.framework"
+
+    s.swift_version = '5.0'
+  
+    s.source = { :http => 'https://github.com/Azure/Communication/releases/download/v1.0.0-beta.4/azurecommunicationcalling.framework-1.0.0-beta.4.zip' }
+    s.source_files = 'Headers/*.h'
+    s.public_header_files = "Headers/*.h"
+    s.dependency 'AzureCore', '~> 1.0.0-beta.2'
```
bsiegel, line 48: I don't believe Calling depends on AzureCore any longer, double check but we should be able to remove this.

EULA.txt
```diff
@@ -0,0 +1,79 @@
+MICROSOFT SOFTWARE LICENSE TERMS
```
bsiegel, line 1: Can we move this into e.g. `sdk/communiation/AzureCommunicationCalling/` ? I think seeing this in the root of the repo might concern users. I'm fine if this EULA is the only file in that directory.

AzureCommunicationCalling.podspec
raosanat: Dependency needs to be still there for older releases we have done.

bsiegel: If the podspec is starting at 1.0.0-beta.4 and 1.0.0-beta.4 doesn't require AzureCore I don't see any reason it couldn't be removed.

EULA.txt
raosanat: Moved

AzureCommunicationCalling.podspec
raosanat: Yes now i get it, AzureCommunication already specifes deps on AzureCore , so yes we dont directly dont depend on AzureCore

bsiegel (approved):
:tada: Looks good, all specs passed validation and got pushed to my Specs repo https://github.com/bsiegel/Specs

:shipit:

PR #515, opened 2020-10-22
Expose Pipeline primitives to end user

### LINKED ISSUE #502: App developers should be able to override the pipeline and supply their own context
Currently, the pipeline policies are controlled by the SDK author, and the end-user (app developer) can only customize them in ways exposed by the SDK through initializers or other convenience methods. As an escape hatch, we should allow the end user to create, modify and supply their own structure of pipeline policies, including custom policies, and since policies are stateless, they must also be able to submit their own pipeline context objects.

### LINKED ISSUE #509: Storage and Chat Library client initializers do not follow guidelines
https://azure.github.io/azure-sdk/ios_design.html#client-interface

The initializer here differs from that used in AzureStorageBlob and that generated by Autorest.Swift

**Guidelines**
```public init(endpoint: String, credential: String? = nil, options: Options? = nil)```

**Actual**
```public init(baseUrl: URL, authPolicy: Authenticating, withOptions options: AzureOptions? = nil)```

The "right" answer probably lies closer to the guidelines.

### PR DESCRIPTION
**Pull request by tjprescott**
Fixes #509 
Fixes #502

### CODE REVIEW

File: sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift [hunk 1]
```diff
@@ -194,8 +194,8 @@ public class ContentDecodePolicy: PipelineStage {
         else { return }
         contentType = contentType.lowercased().trimmingCharacters(in: .whitespacesAndNewlines)
         do {
-            if let deserializedData = try deserialize(from: returnResponse, contentType: contentType) {
-                returnResponse.add(value: deserializedData as AnyObject, forKey: .deserializedData)
+            if let deserializedData = try deserialize(from: returnResponse, contentType: contentType) as? AnyObject {
```
bsiegel, line 197: Unsure why this changed - does this affect behavior or is it just cosmetic?

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
```diff
@@ -109,7 +109,7 @@ public final class AzureCommunicationChatClient: PipelineClient, PageableClient
         }
 
         // Send request
-        let context = PipelineContext.of(keyValues: [
+        let context = options?.context ?? PipelineContext.of(keyValues: [
```
bsiegel, line 112: Should we be merging in the "default" context entries here rather than just taking the user's context object as-is? I can see arguments both ways - the user shouldn't need to know to set the allowed status codes for example, but also if for some reason the service changes and a new status code is added the user needs an escape hatch to work around our default value for that context key.

sdk/communication/AzureCommunicationChat/Source/Options/AzureCommunicationChatClientOptions.swift
```diff
@@ -26,6 +26,11 @@ public struct AzureCommunicationChatClientOptions: AzureClientOptions {
     public let transportOptions: TransportOptions
     /// The default dispatch queue on which to call all completion handler. Defaults to `DispatchQueue.main`.
     public let dispatchQueue: DispatchQueue?
+    /// An array of `PipelineStage` policies to use in lieu of the default ones.
+    public let pipeline: [PipelineStage]?
```
bsiegel, line 30: Unclear whether this is what the arch board intended, or whether they were asking for some way to inject additional policies at the end of the pipeline instead. I thought I had noted that it was the latter but I could be wrong. I think that might make sense regardless since a policy at the end of the pipeline could pretty much overwrite anything done by any earlier policy.

sdk/communication/AzureCommunicationChat/Source/Options/AzureCommunicationChatClientOptions.swift
```diff
@@ -26,6 +26,11 @@ public struct AzureCommunicationChatClientOptions: AzureClientOptions {
...
+    /// An `HTTPTransportStage` policy to use in lieu of the default.
+    public let transport: HTTPTransportStage?
```
bsiegel, line 32: Can we add these (pipeline, transport) to `TransportOptions` instead? That way they don't have to be repeated in every ClientOptions object and they seem like they make more sense there anyhow.

bsiegel (changes requested):

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
tjprescott: It was necessary with a variant I was trying, but it is no longer necessary so I removed it.

sdk/communication/AzureCommunicationChat/Source/Options/AzureCommunicationChatClientOptions.swift
tjprescott: Done.

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
tjprescott: If we update the default context with any user keys, they should be able to do both.

sdk/communication/AzureCommunicationChat/Source/Options/AzureCommunicationChatClientOptions.swift
tjprescott: This also makes sense within the context of distinguishing between "per request" and "per retry" policies.

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
vcolin7: @bsiegel Is it possible to offer both public APIs? (i.e. something like `replaceContext` and `mergeContext`).

tjprescott:
@bsiegel pushed another iteration with an extensible pipeline. Please take a second look. I still need to add unit tests for extending the PipelineContext

tjprescott:
@bsiegel I've added tests for the Pipeline and Context cases. This PR will not actually implement the RetryPolicy, but does include the logic for structuring the pipeline around the Retry policy.

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
bsiegel: Sounds like the answer is that we just need essentially, `mergeContext`

sdk/storage/AzureStorageBlob/Source/Options/OptionsModels.swift
```diff
@@ -154,7 +154,7 @@ public struct CustomerProvidedEncryptionKey: Codable, Equatable {
     }
 
     /// Specifies the algorithm to use when encrypting data using the given key. Must be AES256.
-    public let algorithm = "AES256"
+    var algorithm = "AES256"
```
bsiegel, line 157: Unrelated change?

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -84,16 +84,16 @@ public final class StorageBlobClient: PipelineClient {
...
     ) throws {
         self.options = options
         super.init(
-            baseUrl: baseUrl,
-            transport: URLSessionTransport(),
-            policies: [
-                UserAgentPolicy(for: StorageBlobClient.self, telemetryOptions: self.options.telemetryOptions),
+            endpoint: endpoint,
+            transport: options.transportOptions.transport ?? URLSessionTransport(),
+            policies: options.transportOptions.perRequestPolicies ?? [
```
bsiegel, line 95: Now that we merge policies I think this needs to be reverted

sdk/storage/AzureStorageBlob/Source/Options/OptionsModels.swift
tjprescott: It was to get rid of an XCode warning.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Yes, it does.

sdk/storage/AzureStorageBlob/Source/Options/OptionsModels.swift
bsiegel: Hmm, odd. Is it just because it's unused? If so I guess maybe just delete it?

bsiegel (approved):

PR #522, opened 2020-10-23
Updates to Support Autorest.Swift.

### LINKED ISSUE #501: Enums that accept any string may not be discoverable
Enums such as HTTPHeader or ContextKey offer strongly typed enum cases to communicate to the end user (app developer) what values are actively supported in AzureCore. This is merely an aid, as the actual APIs accept any string. However, the presence of the enum might confuse developers into not realizing that they can supply a custom value. 

One possible solution would be to expose an enum case of `custom(String)`.

### LINKED ISSUE #514: Helper Structure for QueryParams
We can simplify Autorest.Swift codegen if the syntax for a required query param and an optional one work the same way. If an optional param evaluates to nil, it should be ignored and not added to the query params.

**current**
```swift
func test(var1: String, var2: String?) {
  var queryParams: [QueryParameter] = [
   ("var1": var1)
  ]
}
if let value = var2 {
  queryParams.append(("var2": value))
}
```

**desired**
```swift
func test(var1: String, var2: String?) {
  var queryParams = QueryParameters([
   ("var1": var1),
   ("var2": var2) // ignored if value is nil
  ])
}
```

cc/ @sacheu

### LINKED ISSUE #516: [FEATURE REQ] Helper struct for PathParams
In generated code, to get the generated code compile, we need to manually add empty string as key , value pairs in PathParams when there is no specific Path parameters.

This is because PathParams is a dictionary. Swift does not allow empty dictionary.
Example:
```
        let pathParams = [                                                                                                
            "": ""                                                                                                        
        ] 
```

It will be nice if we convert  PathParams into an array of [String,String], just like QueryParams. So that we can just do something like:
```
let pathParams: [PathParameter] = [                                                                                                                                                                                       
        ]              
```

### LINKED ISSUE #534: [FEATURE REQ] Create helper method to construct request url for Autorest swift code generator
Create a helper method to construct request url in AutoRest swift code generator 

* Remove allowable characters from Swift foundation urlqueryallowed when encode query parameters
* Remove allowable characters from Swift foundation urlqueryallowed when encode query parameters
* Add base64url to encode Data object (it is base64 encoding and a few characters are changes). Difference between base64 and base64url can be found in http://www.base64url.com/
* In the construct url help method, allow caller to encode /skip encoding in path parameters, query parameters

tjprescott:
Closes #514.
Closes #516.
Closes #501.
Closes #534.

- adds generic `RequestParameters` object that can work with query, header or path parameters with a syntax that is consistent regardless of whether a nullable value is supplied, for use in Autorest.Swift.
- adds logic to construct the url with proper regard to the `skipUrlEncoding` parameter from Swagger.
- fixes discoverability of undiscoverable enums and what we will need to do in any event for extensible enums (modelAsString = true).


sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,96 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+public protocol QueryStringConvertible {
+    var queryValue: String { get }
+}
```
tjprescott, line 31: @sacheu when we generate enums, we will need to ensure they conform to this protocol.

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,86 @@
...
+        let stringVal = map { value in
+            if let string = value {
+                return String(string)
+            } else {
+                return ""
+            }
+        }.joined(separator: ",")
+        return stringVal
+    }
+}
```
tjprescott, line 86: We can add implementations for any built-in types here.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -0,0 +1,168 @@
...
+
+// MARK: Test Object Defintions
+
+enum AnimalEnum: String, QueryStringConvertible {
+    case cat
+    case dog
+
+    var queryValue: String {
+        return rawValue
+    }
```
tjprescott, line 38: @sacheu this implementation would need to be added to all String-backed enums. It might make sense to treat all enums like ShapeEnum.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -0,0 +1,168 @@
...
+}
+
+enum FlavorEnum: Int, QueryStringConvertible {
+    case chocolate
+    case vanilla
+
+    var queryValue: String {
+        return String(rawValue)
+    }
```
tjprescott, line 47: This default implementation would need to be added to all Int-backed enums. There's no way (that I can find) to extend one protocol (RawRepresentable) with another (QueryStringConvertible) to put a generic default in AzureCore.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        switch self {
+        case let .custom(val):
+            return val
+        case .square:
+            return "square"
+        case .circle:
+            return "circle"
+        }
+    }
+}
```
tjprescott, line 65: @sacheu @bsiegel this would be the proposed solution for extensible enums (Swagger defines modelAsString true). This would also work for our HTTPHeader and ContextKey enums, since it advertises the custom (extensible) option.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -0,0 +1,168 @@
...
+typealias MyDate = Date
+extension MyDate: QueryStringConvertible {
+    internal static let dateFormat = "yyyy-MM-dd"
+
+    public var queryValue: String {
+        let formatter = DateFormatter()
+        formatter.dateFormat = MyDate.dateFormat
+        return formatter.string(from: self)
+    }
+}
```
tjprescott, line 76: While we can put a generic implementation into AzureCore, I don't think it will work, generally. @sacheu I think we'd need to actually generate the Date schemas from Swagger in a manner similar to this, where each one can specify its own format. Curious what you think.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        ])
+        XCTAssert(query.parameters.count == 1)
+        XCTAssert(query.parameters[0].value == "cat")
+    }
+
+    func test_QueryParameters_WithExtensibleStringBackedEnum() throws {
+        let query = QueryParameters([
+            ("var1", ShapeEnum.custom("parallelogram")),
+            ("var2", ShapeEnum.circle)
+        ])
```
tjprescott, line 147: An example of how the extensible enum would work. If someone wanted to use the value "custom", the would supply `.custom("custom")` so I think this implementation will suffice for all our purposes.

sdk/storage/AzureStorageBlob/Source/Credentials/StorageSASCredential.swift
```diff
@@ -251,19 +251,19 @@ internal class StorageSASAuthenticationPolicy: Authenticating {
...
-    private func parse(sasToken: String) -> [QueryParameter] {
-        var queryItems = [QueryParameter]()
+    private func parse(sasToken: String) -> QueryParameters {
+        var queryItems = QueryParameters()
         for component in sasToken.components(separatedBy: "&") {
             let splitComponent = component.split(separator: "=", maxSplits: 1).map(String.init)
             let name = splitComponent.first!
             let value = splitComponent.count == 2 ? splitComponent.last : ""
-            queryItems.append(name, value?.removingPercentEncoding)
+            queryItems.add(key: name, value: value?.removingPercentEncoding)
```
tjprescott, line 266: @sacheu this logic was ported over--do you think this introduces problems?

sdk/storage/AzureStorageBlob/Source/Options/ClientMethodOptions.swift
```diff
@@ -419,3 +423,11 @@ public struct UploadBlobOptions: AzureOptions, Codable, Equatable {
             encoding, timeoutInSeconds
     }
 }
+
+public typealias SnapshotDate = Date
+extension SnapshotDate: QueryStringConvertible {
```
bsiegel, line 428: Not convinced this alias+extension is optimal, see below.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
```diff
@@ -430,7 +417,10 @@ public final class StorageBlobClient: PipelineClient {
         guard let url = self.url(forTemplate: urlTemplate, withKwargs: pathParams) else { return }
 
         // Construct query
-        var queryParams: [QueryParameter] = []
+        let queryParams = QueryParameters([
+            ("snapshot", options?.snapshot),
```
bsiegel, line 421: Feels like the right place to convert this to a RFC1123 format string is here rather than making an alias and adding protocol conformance. But if you still want to do that without our special QueryStringConvertible protocol, you could do it by defining the alias, and then conforming to CustomStringConvertible instead.

bsiegel (changes requested):

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,91 @@
...
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+public protocol QueryStringConvertible {
+    var queryValue: String { get }
+}
+
+public typealias QueryParameter = (key: String, value: String)
```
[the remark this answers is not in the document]
tjprescott: @sacheu can you comment on this? You are more familiar with the testserver cases.

sdk/storage/AzureStorageBlob/Source/StorageBlobClient.swift
tjprescott: Doing the conversion here means that Autorest.Swift, its stencil files and its backing code models need to be increasingly more complicated and hard to maintain in order to know which string to render. That is why the purpose of this PR is to ensure that AzureCore handles all complexity so that all lines can be resolved to the following pattern: `("{key}": {path}{value})`

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,91 @@
...
+
+public struct QueryParameters {
+    internal var parameters = [QueryParameter]()
+
+    public init(_ params: [(key: String, value: QueryStringConvertible?)]? = nil) {
```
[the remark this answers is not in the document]
tjprescott: Done.

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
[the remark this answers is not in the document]
tjprescott: If this is true, then it won't work in autorest.swift anyways today. Currently, the pattern generated would be:

```swift
if let unwrapped = someOptional {
  queryParams.add(("someOptional": unwrapped))
}
```

This would prevent us from every being able to supply nil as you describe, absent some sentinel value.

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,91 @@
...
+        for param in params ?? [] {
+            add(key: param.key, value: param.value)
```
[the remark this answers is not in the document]
tjprescott: Throwing away the nil values is crucial to this PR, but I replace this logic with a filter/map statement. Not sure if that really improves the efficiency, but it avoids using the mutating function.

sdk/core/AzureCore/Source/Pipeline/Transport/QueryParameters.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public mutating func add(key: String, value: QueryStringConvertible?) {
+        // skip values that evaluate to nil
+        guard let val = value else { return }
+        parameters.append((key: key, value: val.queryValue))
+    }
+}
+
+extension Int: QueryStringConvertible {
```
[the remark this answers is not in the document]
tjprescott: I initially tried an approach taking an array of `[AnyObject]` and doing a switch on the type. However, this failed because the value 1 of type `Any` successfully casts to a `Bool` and an `Int`, making it impossible to meaningfully distinguish between the two. It essentially boils down to whichever case statement is listed first.

I opted for `QueryStringConvertible` over `CustomStringConvertible` since someone might want to use `CustomStringConvertible` for other, more general purposes and not have that hijacked by us.

The other implementation I pursued was having many overloads of the `add` method, but this resulted in a lot of methods when you dot on the struct. Thus, having a single method that takes a protocol type seemed like the cleanest implementation.

The purpose of this is essentially to relocate the complexity of generating the right string from the generated code (via Autorest.Swift) into AzureCore, which is much much easier to do because of the limits of the Swift Stencil framework.

While these extensions seem awfully special case, they are special cased based on what the testserver expects us to send on the wire in these instances.

sdk/storage/AzureStorageBlob/Source/Options/ClientMethodOptions.swift
tjprescott: Hmm.. yes, this won't work because:
```swift
typealias CatDate = Date
extension CatDate: QueryStringConvertible {
...
}

typealias DogDate = Date
extension DogDate: QueryStringConvertible {
...
}
```

This raises an invalid redundant conformance of Date to QueryStringConvertible.

tjprescott: Unfortunately, `structs` can't be extended, so I can't do this either:
```swift
struct SnapshotDate: Date, CustomStringConvertible {
  let format = "blah"
  ...
}
```

sdk/core/AzureCore/Tests/QueryParametersTests.swift
tjprescott: This solution didn't work and has been replaced.

sdk/storage/AzureStorageBlob/Source/Options/ClientMethodOptions.swift
tjprescott: This solution has been replaced in the latest commit.

sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift
```diff
@@ -47,54 +47,8 @@ extension URL {
     }
 }
 
-extension Date {
```
tjprescott, line 50: These have been relocated and consolidated to an `AzureDateFormat` enum that accepts `custom(String)`.

sdk/core/AzureCore/Tests/QueryParametersTests.swift
```diff
@@ -64,14 +64,28 @@ enum ShapeEnum: QueryStringConvertible {
...
+    }
+
+    init(_ date: Date) {
+        self.date = date
+    }
+
+    public var queryValue: String {
+        return MyDate.formatter.string(from: date)
     }
 }
```
tjprescott: @sacheu @bsiegel this is an example of how we can handle dates in AzureCore instead of in the code gen itself. We could have a `Dates.swift` file with these generated the same way we have an `Enumerations.swift` file.

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    // MARK: Comparable
+
+    public static func < (lhs: Iso8601Date, rhs: Iso8601Date) -> Bool {
+        return lhs.date < rhs.date
+    }
+}
+
+public struct Rfc1123Date: RequestStringConvertible, Codable, Equatable, Comparable {
+    static var dateFormat: AzureDateFormat = .rfc1123
```
tjprescott, line 74: These are reusable date definitions for services that use these very common formats (like Storage). Custom formats will require what is essentially a copy-paste-edit of these definitions.

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+    public init() {
+        self.date = Date()
+    }
+
+    public init?(string: String?) {
+        guard let date = Self.formatter.date(from: string ?? "") else { return nil }
+        self.date = date
+    }
+
+    public init?(_ date: Date?) {
```
sacheu, line 55: Does input parameter `date` needs to be optional?

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+        return Self.formatter.string(from: date)
+    }
+
+    // MARK: Initializers
+
+    public init() {
+        self.date = Date()
+    }
+
+    public init?(string: String?) {
```
sacheu, line 50: Does input parameter `string` needs to be optional?   I think we can make it a required parameter

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+        guard let date = Self.formatter.date(from: string ?? "") else { return nil }
```
sacheu, line 95: Does input parameter `string` needs to be optional?

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+        self.date = date
+    }
+
+    public init?(_ date: Date?) {
+        guard let unwrapped = date else { return nil }
+        self.date = unwrapped
+    }
+
```
sacheu, line 103: Does input parameter `date` needs to be optional?

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+import Foundation
+
+public struct Iso8601Date: RequestStringConvertible, Codable, Equatable, Comparable {
+    static var dateFormat: AzureDateFormat = .iso8601
+
+    static var formatter: DateFormatter {
+        return Self.dateFormat.formatter
+    }
+
+    public var date: Date
```
sacheu, line 36: why this is a `var` instead of `let`?

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
```diff
@@ -0,0 +1,150 @@
...
+    case iso8601
+
+    /// Retrieve a `DateFormatter` for the given date format.
+    public var formatter: DateFormatter {
+        switch self {
+        case let .custom(format):
+            let formatter = DateFormatter()
+            formatter.dateFormat = format
+            return formatter
+        case .rfc1123:
```
sacheu, line 134: Azure core already has these date formatter defined. Can we resue them? https://github.com/Azure/azure-sdk-for-ios/blob/65a2dba0447ba20d58e74af0f96ce2a2f123c8d3/sdk/core/AzureCore/Source/Util/ConvertingInitializers.swift

tjprescott: They have been relocated here.

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
tjprescott: It is so the user can decide whether they want the date to be `var` or `let`.  

If the inner date is marked `let`:
```Swift
    let date1 = Rfc1123Date()
    date1.date = Date() // fails because inner date is let

    var date2 = Rfc1123Date()
    date2.date = Date() // fails because inner date is let
```

If the inner date is marked `var`:
```Swift
    let date1 = Rfc1123Date()
    date1.date = Date() // fails because outer date1 is let

    var date2 = Rfc1123Date()
    date2.date = Date() // succeeds
```

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
tjprescott: The nullability is crucial to being able to pass in something that might be nil and not have to do the unwrapping dance.

tjprescott:
@bsiegel @sacheu I pushed the changes we discussed. The CI is failing for a reason wholly unrelated to the PR (some link broken in azure-sdk-tools).

tjprescott:
@bsiegel @sacheu the CI has been fixed and the PR is passing.

sdk/core/AzureCore/Source/DataStructures/AzureDate.swift
bsiegel: Maybe just call this `value` instead? Seeing `date1.date = Date()` is weird to me.

sdk/core/AzureCore/Source/Errors.swift
```diff
@@ -64,3 +64,19 @@ public enum AzureError: BaseError {
         }
     }
 }
+
+extension String: LocalizedError {
```
bsiegel, line 68: I don't recall if we discussed this in the meeting - what is the use case here?

sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
```diff
@@ -26,142 +26,350 @@
...
+/// Extensions to work with `HTTPHeader` values within a collection of `HTTPHeaders`.
 
 /// Common HTTP headers.
-public enum HTTPHeader: String {
+public enum HTTPHeader: RequestStringConvertible, Equatable, Hashable {
+    /// Use when the header value you want is not in the list.
+    case custom(String)
     /// Accept
-    case accept = "Accept"
+    case accept
```
bsiegel, line 40: Can we use associated value defaults here, e.g. `case accept(requestString: String = "Accept")`?

Then we could remove any special knowledge inside the getter for `requestString`, and also in the `init` we can just loop through all the cases and find one whose requestString matches (or return .custom otherwise) rather than having to do a big switch on string values there.

sdk/core/AzureCore/Source/Errors.swift
tjprescott: This was something I upleveled from Autorest.Swift. @sacheu can you touch on this?

sdk/core/AzureCore/Source/Pipeline/Transport/HTTPHeader.swift
tjprescott: If you do that, you would be able to override the value (`.accept("MadeUpValue"`), which at least in my mind would break the principle of least surprise.  The switch statements are annoying, but they are also fairly trivial to generate.

bsiegel: That's true, it is ugly but I'm fine with dealing with that if it's auto-generated. Maybe just add some comments to indicate that the file should not be hand-modified, or to be careful if doing so.

tjprescott: This file isn't auto-generated since it's in Core, but ones generated by the Swagger (where modelAsString is true) would be. Here, you would need to add a few annotations when adding new convenience values to AzureCore, but that's probably not going to happen with regularity. The compiler would force you to update the requestString method and, if you forgot to add it to the string init it would just consider it a custom value and everything would continue to work.

sdk/core/AzureCore/Source/Errors.swift
tjprescott: This is is to support the proper generation of AzureError when the error response body is something other than a JSON object (i.e. string, int, etc). This was created in Autorest.Swift to support the error-body testserver swagger, but should really live in AzureCore.

sacheu: This is added in order to support xms-error-responses.json autorest testsever test swagger

bsiegel (approved):
Looks good, I think all my comments were addressed.

sacheu (approved):

PR #542, opened 2020-11-02
Remove test artifacts from AzureCore public interface

tjprescott:
These mistakenly appear in the AzureCore public interface and don't need to.


bsiegel (approved):
Looks good to me :shipit:

PR #547, opened 2020-11-06
Regnerate AzureCommunicationChat for AzureCore updates

tjprescott:


sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
```diff
@@ -39,18 +44,22 @@ public final class AzureCommunicationChatClient: PipelineClient, PageableClient
...
     /// Create a AzureCommunicationChatClient client.
     /// - Parameters:
-    ///   - baseUrl: Base URL for the AzureCommunicationChatClient.
+    ///   - endpoint: Base URL for the AzureCommunicationChatClient.
     ///   - authPolicy: An `Authenticating` policy to use for authenticating client requests.
     ///   - options: Options used to configure the client.
     public init(
-        baseUrl: URL,
+        endpoint: String,
+        api - version: String,
```
tjprescott, line 52: @sacheu something is wrong here. Is this the Swagger or autorest?

tjprescott (commented):
Problem with the apiVersion parameter name. Also, we shouldn't regenerate the README if it already exists. cc/ @sacheu

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
sacheu: We need to fix this in autorest swift. I will take a look.

tjprescott: There's a problem in autorest since this shouldn't even be in the signature. It should be in the ClientOptions.

PR #563, opened 2020-11-18
Regenerate AzureCommunicationChat

tjprescott:
Necessary for release of AzureSDK.


sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
```diff
@@ -16,19 +16,24 @@ import Foundation
...
-        case v20200921preview2 = "2020-09-21-preview2"
+        /// API version ""
+        case v = ""
 
         /// The most recent API version of the
         public static var latest: ApiVersion {
-            return .v20200921preview2
+            return .v
         }
     }
```
tjprescott, line 38: @sacheu looks like there's still an issue here with how Autorest.Swift is (or rather, is not) generating the api version field.

sdk/communication/AzureCommunicationChat/Package.swift
```diff
@@ -24,7 +24,7 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/Azure/azure-sdk-for-ios.git",
-            .branch("master")
+            .revision("3e3c80d60173613c8dd4cb6b219188cf5070e8e7")
```
bsiegel, line 27: Is this change intended to be merged?

tjprescott: Yes, though we may need to change it to something else like the release tag? I'm curious how either can really even matter since if you download AzureSDK, AzureCore is baked in.

sdk/communication/AzureCommunicationChat/Source/AzureCommunicationChatClient.swift
tjprescott: I have manually reverted this and updated the issue in the Autorest.Swift repo to fix this.

tjprescott:
There were numerous issues with the generated code that we have documented here (https://github.com/Azure/autorest.swift/issues/258) but this is now ready to merge and release. Many thanks to @sacheu for testing the SDK.

bsiegel (approved):
The manual fixes look reasonable to me. If the tests are passing, :shipit:

Issue #532, opened 2020-10-26
Support building and shipping CocoaPods


mitchdenny:
Here is what we need to do to support CocoaPods in our build/release system:

- [x] Convert *.podspec files to *.json (pod ipc spec)
- [x] Add analyze step which ensures all pod versions are aligned.
- [x] Add ```pod lib lint``` to the build
- [x] Add tagging support for repo
- [x] Add publish to trunk (API)
- [ ] Add auto version bumping

Open questions:

1. How do we decide when something is GA and beta?
2. How do we handle shipping beta releases with servicing?

mitchdenny:
Duplicate of https://github.com/Azure/azure-sdk-for-ios/issues/681

PR #564, opened 2020-11-18
Switch to *.podspec.json format for better automation.

mitchdenny:
As part of #532 we need to have the Azure SDK specs in ```*.podspec.json``` format so that we can run some automation to enforce versioning logic etc. This is the initial step (cutting over to ```*.podspec.json```.


eng/pipelines/templates/steps/verify-versions.yml
```diff
@@ -0,0 +1,8 @@
+steps:
+- task: PowerShell@2
```
weshaggard, line 2: Why make a template for this as opposed to adding the one step in the other template?

weshaggard (approved):
I don't have much context on podspec files but the other changes look reasonable.

mitchdenny:
Thanks Wes. I'm just waiting for @bsiegel to be ready to take these changes. I think we are clear of releases now. We need these in master so we can continue the rest of the automation.

bsiegel:
Need to regenerate podspecs based on the latest versions in master.

mitchdenny:
@bsiegel regenerated.

bsiegel (approved):
This looks good to me. From here we'd like to also get to the point where we can update the versions in the `.jazzy/*.yml` files, the READMEs, and perhaps even the xcodeproj files.

Issue #570, opened 2020-11-20
[BUG] `swift build` broken on master branch


tjprescott:
The following PR:
https://github.com/Azure/azure-sdk-for-ios/pull/503

has broken the ability to run `swift build` of any Targets (not just AzureCommunciation). 

We should:
1) Resolve the error (or revert the PR)
2) Add a test to CI to ensure that `swift build` always passes.

bsiegel:
I suspect the right thing to do here is probably to add a filter on the test targets in the Package.swift file, so that the targets only include *.swift files. That's where I'd start at least.

JoshuaLai:
Package.swift should be able to support excluding files with an exclude[]. So I am going to try to exclude the obj-c tests but if there is a super high urgency I created the PR to revert the change just in case here: https://github.com/Azure/azure-sdk-for-ios/pull/571

PR #572, opened 2020-11-20
[Communication] - [Package] - Exclude ObjC files 

JoshuaLai:
Exclude the obj-c in package, and include swift build command into the CI

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -72,6 +72,8 @@ let package = Package(
             dependencies: ["AzureCommunication"],
             path: "sdk/communication/AzureCommunication",
             exclude: [
+                "Tests/AzureCommunicationTests-Bridging-Header.h",
+                "Tests/ObjCCommunciationUserCredentialTests.m",
                 "Tests/ObjCCommunicationUserCredentialAsyncTests.m",
                 "Tests/ObjCTokenParserTests.m"
             ],
--- a/Package.swift
+++ b/Package.swift
@@ -72,8 +72,6 @@ let package = Package(
             dependencies: ["AzureCommunication"],
             path: "sdk/communication/AzureCommunication",
             exclude: [
-                "Tests/AzureCommunicationTests-Bridging-Header.h",
-                "Tests/ObjCCommunciationUserCredentialTests.m",
                 "Tests/ObjCCommunicationUserCredentialAsyncTests.m",
                 "Tests/ObjCTokenParserTests.m"
             ],
--- a/Package.swift
+++ b/Package.swift
@@ -71,6 +71,12 @@ let package = Package(
             name: "AzureCommunicationTests",
             dependencies: ["AzureCommunication"],
             path: "sdk/communication/AzureCommunication",
+            exclude: [
+                "Tests/AzureCommunicationTests-Bridging-Header.h",
+                "Tests/ObjCCommunciationUserCredentialTests.m",
+                "Tests/ObjCCommunicationUserCredentialAsyncTests.m",
+                "Tests/ObjCTokenParserTests.m"
+            ],
             sources: ["Tests"]
         )
     ],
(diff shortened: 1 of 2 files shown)


JoshuaLai:
https://github.com/Azure/azure-sdk-for-ios/issues/570
An idea to fix the swift build.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -37,7 +37,11 @@ jobs:
       - publish: $(Build.ArtifactStagingDirectory)/Build/Products/Release-iphonesimulator
         displayName: 'Publish release artifacts [iphonesimulator]'
         artifact: 'bundles-release-iphonesimulator'
-
+      
+      - script: |
+          swift build
+        displayName: "Use swift build for SPM"
+
```
JoshuaLai, line 44: Not sure if this script is enough or needs more details. Or needs to be in its own job, currently I just put it under the existing build job

tjprescott: It should be sufficient. If you exclude one of your exclusions does it fail the CI?

JoshuaLai: Created commit b469331 where I removed some files and you can see the CI failed for that case so looks like it is working as expected.

Package.swift
```diff
@@ -71,6 +71,12 @@ let package = Package(
             name: "AzureCommunicationTests",
             dependencies: ["AzureCommunication"],
             path: "sdk/communication/AzureCommunication",
+            exclude: [
+                "Tests/AzureCommunicationTests-Bridging-Header.h",
+                "Tests/ObjCCommunciationUserCredentialTests.m",
+                "Tests/ObjCCommunicationUserCredentialAsyncTests.m",
+                "Tests/ObjCTokenParserTests.m"
+            ],
```
tjprescott, line 79: Is there a way to wildcard exclude *.m and *.h?  If not, this is okay for now.

tjprescott (commented):
LGTM though I would prefer the exclude statement to be more generic. Would also prefer @bsiegel or @mitchdenny  approve.

Package.swift
JoshuaLai: @tjprescott Doing a bit more research on this. But according to this post from Jun 22 2020 https://forums.swift.org/t/support-for-wildcard-paths-in-package-swift/37814 it looks like it was pitched but not able to yet.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -37,7 +37,11 @@ jobs:
       - publish: $(Build.ArtifactStagingDirectory)/Build/Products/Release-iphonesimulator
         displayName: 'Publish release artifacts [iphonesimulator]'
         artifact: 'bundles-release-iphonesimulator'
-
+      
+      - script: |
```
bsiegel, line 41: I would prefer this be in its own job, that way we can run Xcode and SPM builds in parallel.

tjprescott (changes requested):
Can't merge this until I understand why `/azp run` fails.

tjprescott:
@bsiegel ADO is not running the pipeline on this PR. Any ideas why?

bsiegel:
> @bsiegel ADO is not running the pipeline on this PR. Any ideas why?

I'm not positive but I believe the job YAML for "Swift Build" is invalid. Let me try pushing a change to fix that.

JoshuaLai:
> > @bsiegel ADO is not running the pipeline on this PR. Any ideas why?
> 
> I'm not positive but I believe the job YAML for "Swift Build" is invalid. Let me try pushing a change to fix that.

Could it be a formatting issue? Spaces are off?

bsiegel:
> Could it be a formatting issue? Spaces are off?

A couple things:

1. Spacing was off for 'displayName'
2. The job name is not allowed to contain spaces

tjprescott (approved):

PR #574, opened 2020-11-23
Regenerate Chat SDK

knvsl:
- Regenerate Chat SDK using preview3 swagger
- Reapplied manual fixes for some of the issues documented here: https://github.com/Azure/autorest.swift/issues/258
- Manually fix generated code for setting the azure-Acs-IdempotencyKey header

- FYI rebased so we have the changes from today to exclude obj-c tests


knvsl:
After the compiler errors are fixed I notice successful build results in a bunch of linting(?) changes in unrelated files. Should those be included in this PR/is that expected?

sdk/communication/AzureCommunicationChat/Source/Generated/AzureCommunicationChatService.swift
```diff
@@ -1750,23 +1775,29 @@ public final class AzureCommunicationChatService {
...
 
         // Construct headers
         var headers = HTTPHeaders()
         headers["Content-Type"] = "application/json"
         headers["Accept"] = "application/json"
+        // Process endpoint options
+        // Query options
+
+        // Header options
+        if let azureAcsIdempotencyKey = options?.azureAcsIdempotencyKey {
```
knvsl, line 1798: I think this was pretty much the only new manual fix I had to add (rest were just re-applying previous manual fixes). Generated code was producing:

`if let azure - Acs - IdempotencyKey = options?.azure - Acs - IdempotencyKey {
            headers["azure-Acs-IdempotencyKey"] = azureAcsIdempotencyKey
        }`

sdk/communication/AzureCommunicationChat/Source/Custom/AzureCommuncationChatClient+Convenience.swift
```diff
@@ -38,7 +38,7 @@ extension AzureCommunicationChatClient {
     public func listChatReadReceipts(
```
tjprescott, line 38: I would just add that you could just move this file into `Source` since anything not in `Source/Generated` is considered to be custom.

sdk/communication/AzureCommunicationChat/Source/Custom/AzureCommuncationChatClient+Convenience.swift
```diff
@@ -192,11 +192,11 @@ extension AzureCommunicationChatClient {
     ///     success.
     public func removeChatThreadMember(
```
tjprescott, line 193: You probably also want to rename these, for example `removeChatParticipant`

sdk/communication/AzureCommunicationChat/Source/Custom/AzureCommuncationChatClient+Convenience.swift
```diff
@@ -161,10 +161,10 @@ extension AzureCommunicationChatClient {
     ///     success.
     public func listChatThreadMembers(
```
tjprescott, line 162: `listChatParticipants`.

This layer was created (rather hastily) so that customers don't have to say `client.azureCommandChatService.blah` and can just say `client.blah` the way they could with the first release. Since you are already writing a convenience layer, feel free to retain or remove as much of this as you like.

knvsl: Ok cool! I moved it out of custom folder & updated some of the names commented above with the API renaming, kept as is otherwise & will leave any changes for the refactor PR.

sdk/communication/AzureCommunicationChat/Source/Generated/AzureCommunicationChatService.swift
tjprescott: @sacheu can you take a look? I thought we'd squashed most of these.

sacheu: I have not tried using preview 3 chat sdk. I will take a look at this issue.

knvsl: Thanks @sacheu for the fixes, only new thing I notice is the api version causes linting error with the underscores:

Generated is producing:
        `case v2020_11_01_preview3 = "2020-11-01-preview3"`

Regenerated SDK has been tested with demo app and looks good otherwise.

tjprescott:
@knvsl your issue with the generated names `looking - like - this` has been fixed in https://github.com/Azure/autorest.swift/pull/274. Thanks @sacheu!

PR #561, opened 2020-11-17
Refactor Chat into two clients

knvsl:
Following pattern of other SDKs, split the chat client into two clients ChatClient and ChatThreadClient.

PR is in draft to gather initial feedback & still needs testing.

PR #578, opened 2020-11-25
Add ChatClient and ChatThreadClient convenience layer

knvsl:
- Addressing previous feedback from the old draft PR: https://github.com/Azure/azure-sdk-for-ios/pull/561
- ChatClient is responsible for managing ChatThreads and creating ChatThreadClients for them
- ChatThreadClient handles operations on the chat thread like sending messages, adding participants
- Updated README


sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+        )
+
+        self.service = client.azureCommunicationChatService
+    }
+
+    // MARK: Public Methods
+
+    /// Create a ChatThreadClient for the ChatThread with id threadId.
+    /// - Parameters:
+    ///   - forChatThread: The threadId.
```
tjprescott, line 74: If you use XCode's built-in documentation function, this would be populated `threadId`. You should follow that convention.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+        )
+    }
+
+    /// Create a new ChatThread and return the ChatThreadClient for it.
+    /// - Parameters:
+    ///   - chatThread: Request for creating a chat thread with the topic and members to add.
+    ///   - withOptions: Create chat thread options.
+    ///   - completionHandler: A completion handler that receives a ChatThreadClient on success.
+    public func create(
+        chatThread: CreateChatThreadRequest,
```
tjprescott, line 90: Is `CreateChatThreadRequest` just a proxy for a `ChatThread` object? If so, it would make sense to consider changing this.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+                        throw AzureError.service("Service response does not contain ThreadId.")
+                    }
+
+                    // Create and return a ChatThreadClient for the new ChatThread
+                    let chatThreadClient = try ChatThreadClient(
+                        endpoint: self.endpoint,
+                        credential: self.credential,
+                        threadId: threadId,
+                        withOptions: self.options
+                    )
```
tjprescott, line 108: I still disagree with this. By the principle of least surprise, I would not expect creating an object to return a client for that object. How does someone create a chat thread without creating a client?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+                completionHandler(.failure(azureError), httpResponse)
+            }
+        }
+    }
+
+    /// Get the ChatThread with given id.
+    /// - Parameters:
+    ///   - withId: The chat thread id.
+    ///   - withOptions: Get chat thread options.
+    ///   - completionHandler: A completion handler that receives the chat thread on success.
```
tjprescott, line 126: Again, the labels here should be the internal labels, not the external one, per XCode's convention.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+    public func getChatThread(
+        withId chatThreadId: String,
+        withOptions options: AzureCommunicationChatService.GetChatThreadOptions? = nil,
+        completionHandler: @escaping HTTPResultHandler<ChatThread>
```
tjprescott, line 130: Would like @bsiegel to weigh in on this, vs the alternative
`public func get(chatThread: String, withOptions:...)`

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -0,0 +1,182 @@
...
+    /// Delete the ChatThread with id chatThreadId.
+    /// - Parameters:
+    ///   - chatThreadId: The chat thread id.
+    ///   - withOptions: Delete chat thread options.
+    ///   - completionHandler: A completion handler.
+    public func delete(
+        chatThreadId: String,
+        withOptions options: AzureCommunicationChatService.DeleteChatThreadOptions? = nil,
+        completionHandler: @escaping HTTPResultHandler<Void>
+    ) {
```
tjprescott, line 171: This is now inconsistent with `getChatThread(withId: ...)`. Whatever naming convention is used, it should be consistent across the SDK.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+
+    // MARK: Public Methods
+
+    /// Updates the ChatThread's properties.
+    /// - Parameters:
+    ///   - chatThread: Request for updating a chat thread.
+    ///   - options: Update chat thread options.
+    ///   - completionHandler: A completion handler that receives a status code on success.
+    public func update(
+        chatThread: UpdateChatThreadRequest,
```
tjprescott, line 82: Consider accepting a `ChatThread` object unless the `UpdateChatThreadRequest` object has additional properties.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+        withOptions options: AzureCommunicationChatService.UpdateChatThreadOptions? = nil,
+        completionHandler: @escaping HTTPResultHandler<Void>
```
tjprescott, line 84: It's curious that the service doesn't return the updated object.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    /// Sends a read receipt.
+    /// - Parameters:
+    ///   - request: Request for sending a read receipt.
+    ///   - options: Send read receipt options.
+    ///   - completionHandler: A completion handler that receives a status code on success.
+    public func send(
+        readReceipt: SendReadReceiptRequest,
```
tjprescott, line 103: Consider using just `ReadReceipt` here.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    /// Sends a message to a ChatThread.
+    /// - Parameters:
+    ///    - chatMessage : Request for sending a message.
+    ///    - options: A list of options for the operation.
+    ///    - completionHandler: A completion handler that receives a status code on success.
+    public func send(
+        message: SendChatMessageRequest,
```
tjprescott, line 163: Consider accepting just `ChatMessage` here.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    /// Gets a message by id.
+    /// - Parameters:
+    ///    - withId : The message id.
+    ///    - options: Get chat message options
+    ///    - completionHandler: A completion handler that receives the chat message on success.
+    public func getMessage(
+        withId chatMessageId: String,
```
tjprescott, line 184: This is inconsistent with your update and delete methods. They should all be consistent.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    /// Adds thread participants to a ChatThread. If the participants already exist, no change occurs.
+    /// - Parameters:
+    ///    - request : The request with thread participants to be added to the thread.
+    ///    - options: Add chat participants options.
+    ///    - completionHandler: A completion handler that receives a status code on success.
+    public func add(
+        chatParticipants: AddChatParticipantsRequest,
```
tjprescott, line 284: Consider accepting `[ChatParticipant]` here.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,348 @@
...
+                    completionHandler(.failure(error), httpResponse)
+                }
+            }
+    }
+
+    /// Gets the participants of the thread.
+    /// - Parameters:
+    ///    - options: List chat participants options.
+    ///    - completionHandler: A completion handler that receives the list of members on success.
+    public func listParticipants(
```
tjprescott, line 334: in other places you refer to them as `chatParticipants`. It should stay consistent throughout--either `participants` or `chatParticipants`

tjprescott (changes requested):
Some comments, mostly on naming, but certainly on the right track. Would like @bisiegel to review as well.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: Thanks, I checked and this is actually supposed to be renamed to updateTopic() or similar. 

Regards to making everything consistent,

As an example for send methods we have:
`public func send(readReceipt: SendReadReceiptRequest, withOptions...)`
`public func send(message: SendChatMessageRequest, withOptions...)`
and 
`public func sendTypingNotification(withOptions...)` where there's no "typing notification" argument, just options, similar to the list methods - listMessages, listParticipants etc.

If we go with the format of `getChatThread(withId...)` do we still leave `send(message...)` or change to `sendMessage(message...)` ie should I make them ALL consistent or only where it isn't redundant?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
knvsl: I agree its not intuitive to return the client - I found out that there is a proposal to decouple the two but this is currently the implementation across the other SDKs so I was being consistent with that until its decided to make the change across the board.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
knvsl: Fixed the docs, thanks!

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: I've update `getChatThread(withId) -> get(chatThreadId)` and updated the getMessage to be consistent as well for now

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: Updated

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: Updated, thanks!

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: I will leave as AddChatParticipantsRequest, generated code doesn't accept [ChatParticipant], so still have to construct AddChatParticipantsRequest anyways.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: For this, SendReadReceiptRequest, and CreateChatThreadRequest will leave as is for similar reasons - ChatThread, ReadReceipt and ChatMessage contains extra properties

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: Commented on Chatmessage and sendreadreceipt, will leave as is

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
knvsl: I ended up changing this to `get(ChatThread...)` to be consistent with the delete, and similar in ChatThreadClient have updated getMessage to be consistent with `get(message...)`

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -0,0 +1,350 @@
...
+        }
+    }
+
+    /// Sends a message to a ChatThread.
+    /// - Parameters:
+    ///    - chatMessage : Request for sending a message.
+    ///    - options: A list of options for the operation.
+    ///    - completionHandler: A completion handler that receives a status code on success.
+    public func send(
+        chatMessage: SendChatMessageRequest,
```
knvsl, line 165: For send(chatMessage), send(readReceipt) will keep type as SendChatMessageRequest and SendReadReceiptRequest, as they aren't quite the same as ChatMessage and ReadReceipt, example is SendReadReceiptRequest is just the id of the message, not an actual ChatMessageReadReceipt

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -238,7 +237,7 @@ Use the `listChatThreads` method to retrieve a list of threads.
 ```swift
 import AzureCore
 let options = ListChatThreadsOptions(maxPageSize: 1)
-client.listChatThreads(withOptions: options) { result, _ in
+chatClient.listChatThreads(withOptions: options) { result, _ in
```
bsiegel, line 240: If this is a `ChatClient` which always operates on Chats, how about naming this e.g. `chatClient.listThreads()` instead?

bsiegel, line 240: Since this is a "ChatClient", would it be simpler to call this method `listThreads`?

bsiegel: Also just verifying that there's an overload where the customer can omit the `withOptions` parameter

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -209,12 +208,12 @@ client.create(chatThread: thread) { result, _ in
...
 #### Get a thread
 
-Use the `getChatThread` method to retrieve a thread.
+Use the `get` method of `ChatClient` to retrieve a thread.
 
 - `chatThreadId` is the unique ID of the thread.
 
 ```swift
-client.getChatThread(chatThreadId: threadId) { result, _ in
+chatClient.get(chatThreadId: threadId) { result, _ in
```
bsiegel, line 216: I'm a little torn on this. When we have a method whose name is only a verb, the first parameter must be the noun that the verb operates on. This method "gets" a thread, not a thread ID, so calling it `get(threadId:)` doesn't make sense. It could be called `get(thread:)`, if it were obvious to the customer that the `thread` parameter is in fact the thread ID. However I am a little dubious that this would be the case.

An alternative could be `getThread(withId:)` which moves the noun (back) into the method name, and provide some nice consistency with `listThreads` below. I am leaning towards this as the best option unless you feel that a customer would immediately understand what they are supposed to provide as the parameter to `get(thread:)`.

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -252,38 +251,15 @@ client.listChatThreads(withOptions: options) { result, _ in
...
 #### Delete a thread
 
-Use `deleteChatThread` method to delete a thread.
+Use `delete` method of `ChatClient` to delete a thread.
 
 - `chatThreadId` is the unique ID of the thread.
 
 ```swift
-client.deleteChatThread(chatThreadId: threadId) { result, httpResponse in
+chatClient.delete(chatThreadId: threadId) { result, httpResponse in
```
bsiegel, line 262: Same comment here. Either `delete(thread:)` or `deleteThread(withId:)`

sdk/communication/AzureCommunicationChat/README.md
bsiegel: Also just as a note, if we went with the former option, the internal name for the parameter could still denote that we're asking for the thread ID here (that is, the API signature in source code would be something like `get(thread threadId: String)`). I'm not sure if the customer would actually see the internal name though to help guide them.

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -315,7 +290,7 @@ let message = SendChatMessageRequest(
     senderDisplayName: "An Important person"
 )
 
-getClient().send(chatMessage: message, chatThreadId: threadId) { result, _ in
+chatThreadClient.send(chatMessage: message) { result, _ in
```
bsiegel, line 293: Similar comment here, could we simply call the parameter `message`?

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -328,15 +303,14 @@ getClient().send(chatMessage: message, chatThreadId: threadId) { result, _ in
...
+Use the `get` method of `ChatThreadClient` to retrieve a message in a thread.
 
-- `chatThreadId` is the unique ID of the thread.
 - `chatMessageId` is the unique ID of the message.
 
 `ChatMessage` is the response returned from getting a message.
 
 ```swift
-client.getChatMessage(chatThreadId: threadId, chatMessageId: messageId) { result, _ in
+chatThreadClient.getChatMessage(chatMessageId: messageId) { result, _ in
```
bsiegel, line 313: Same thing here, either `get(message:)` or `getMessage(withId:)`

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -369,7 +342,7 @@ if let date = dateFormatter.date(from: "2020-08-27T17:55:50Z") {
     )
 }
 
-client.listChatMessages(chatThreadId: threadId, withOptions: options) { result, _ in
+client.listChatMessages(withOptions: options) { result, _ in
```
bsiegel, line 345: Same, perhaps `listMessages()`?

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -385,20 +358,19 @@ client.listChatMessages(chatThreadId: threadId, withOptions: options) { result,
...
-- `chatThreadId` is the unique ID of the thread.
 - `chatMessageId` is the unique ID of the message.
 
 ```swift
 let message = UpdateChatMessageRequest(
     content: "A message content with update()"
 )
 
-getClient().update(chatMessage: message, chatThreadId: threadId, chatMessageId: messageId) { result, _ in
+chatThreadClient.update(chatMessage: message, chatMessageId: messageId) { result, _ in
```
bsiegel, line 373: Hmm, this one starts to get a little more interesting. Maybe something like `updateMessage(withId:changes:)`? I feel like the message ID should be the first parameter to match all of the other methods. I'm not sure what I feel the UpdateChatMessageRequest parameter should be called, but `changes` isn't terrible. Opinions?

Alternatively, using the verb-only naming scheme this would be something like `update(message:changes:)` (or maybe `update(message:withChanges:)`?)

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -411,13 +383,12 @@ getClient().update(chatMessage: message, chatThreadId: threadId, chatMessageId:
...
 
-Use the `deleteChatMessage` method to delete a message in a thread.
+Use the `delete` method of `ChatThreadClient` to delete a message in a thread.
 
-- `chatThreadId` is the unique ID of the thread.
 - `chatMessageId` is the unique ID of the message.
 
 ```swift
-getClient().deleteChatMessage(chatThreadId: threadId, chatMessageId: messageId) { result, _ in
+chatThreadClient.deleteChatMessage(chatMessageId: messageId) { result, _ in
```
bsiegel, line 391: Same comment here. Either `delete(message:)` or `deleteMessage(withId:)`

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -432,18 +403,17 @@ getClient().deleteChatMessage(chatThreadId: threadId, chatMessageId: messageId)
...
+Use the `listChatParticipants` of `ChatThreadClient` method to retrieve the participants of the thread.
 
-- `chatThreadId` is the unique ID of the thread.
 
-`PagedCollection<ChatThreadMember>` is the response returned from listing members.
+`PagedCollection<ChatParticipants>` is the response returned from listing members.
 
 ```swift
-client.listChatThreadMembers(chatThreadId: threadId) { result, _ in
+chatThreadClient.listChatParticipants() { result, _ in
```
bsiegel, line 412: Same here, perhaps just `listParticipants`?

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -457,20 +427,19 @@ client.listChatThreadMembers(chatThreadId: threadId) { result, _ in
...
-let threadMembers = AddChatThreadMembersRequest(
-    members: [ChatThreadMember(
+let threadParticipants = AddChatParticipantsRequest(
+    participants: [ChatParticipant(
         id: userId,
         displayName: "a new member"
     )]
 )
-client.add(chatThreadMembers: threadMembers, chatThreadId: threadId) { result, _ in
+chatThreadClient.add(chatParticipants: threadParticipants) { result, _ in
```
bsiegel, line 442: Same here, perhaps just `add(participants:)`

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -483,13 +452,12 @@ client.add(chatThreadMembers: threadMembers, chatThreadId: threadId) { result, _
...
-Use the `removeChatThreadMember` method to remove a member from a thread.
+Use the `remove` method of `ChatThreadClient` to remove a member from a thread.
 
-- `chatThreadId` is the unique ID of the thread.
-- `chatMemberId` is the user ID in the chat thread's member list.
+- `chatParticipantId` is the user ID in the chat thread's member list.
 
 ```swift
-client.removeChatThreadMember(chatThreadId: threadId, chatMemberId: memberId) { result, _ in
+chatThreadClient.removeChatThreadMember(chatParticipantId: memberId) { result, _ in
```
bsiegel, line 460: Either `remove(participant:)` or `removeParticipant(withId:)` - Note: not "removeChatThreadMemeber"

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -520,16 +488,15 @@ client.sendTypingNotification(chatThreadId: threadId) { result, _ in
...
 
 - `SendReadReceiptRequest` is the model to be passed to this method.
 - `chatMessageId` is the unique ID of the message.
-- `chatThreadId` is the unique ID of the thread.
 
 ```swift
 let readReceipt = SendReadReceiptRequest(chatMessageId: messageId)
 
-client.send(chatReadReceipt: readReceipt, chatThreadId: threadId) { result, _ in
+chatThreadClient.send(chatReadReceipt: readReceipt) { result, _ in
```
bsiegel, line 499: Maybe just `send(readReceipt:)`

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -542,14 +509,14 @@ client.send(chatReadReceipt: readReceipt, chatThreadId: threadId) { result, _ in
...
-Use the `listChatReadReceipts` method to retrieve read receipts for a thread.
+Use the `listChatReadReceipts` method of `ChatThreadClient` to retrieve read receipts for a thread.
 
 - `chatThreadId` is the unique ID of the thread.
 
 `PagedCollection<ReadReceipt>` is the response returned from listing read receipts.
 
 ```swift
-client.listChatReadReceipts(chatThreadId: threadId) { result, _ in
+chatThreadClient.listChatReadReceipts(chatThreadId: threadId) { result, _ in
```
bsiegel, line 519: Same, perhaps `listReadReceipts()` - Note: no threadId parameter

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -563,6 +530,29 @@ client.listChatReadReceipts(chatThreadId: threadId) { result, _ in
 }
 ```
 
+### Thread Update Operations
+
+#### Update the thread's topic
+
+Use the `update` method of `ChatThreadClient` to update a thread's topic.
+
+- `UpdateChatThreadRequest` is the model to pass to this method.
```
bsiegel, line 539: This doesn't seem to be true

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -432,18 +403,17 @@ getClient().deleteChatMessage(chatThreadId: threadId, chatMessageId: messageId)
 
 #### Get thread members
```
bsiegel, line 404:
```suggestion
#### Get thread participants
```

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -457,20 +427,19 @@ client.listChatThreadMembers(chatThreadId: threadId) { result, _ in
 
 Use the `add` method to add members to a thread.
```
bsiegel, line 428:
```suggestion
Use the `add` method to add participants to a thread.
```

sdk/communication/AzureCommunicationChat/README.md
bsiegel: Also, in this case can we just make the parameter an array of `ChatParticipant`s? I don't see any value in wrapping the array with an `AddChatParticipantsRequest`.

sdk/communication/AzureCommunicationChat/README.md
bsiegel: This is another case where the model only has one property. So what about something like `sendReadReceipt(forMessageWithId:)` (or `sendReadReceipt(forMessage:)`) ?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
bsiegel: Other Azure SDKs tend to have something like `myServiceClient.getItemClient()` or similar. Especially since you have a `get` method that returns a `ChatThread`, this should either also return a `ChatThread` or return nothing.

tjprescott: Most commonly I have seen create methods return the object they created.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
tjprescott: Isn't that the purpose of a convenience layer?

sdk/communication/AzureCommunicationChat/README.md
tjprescott: There wouldn't be an overload per se, but the signature should be `withOptions options: BlahOptions? = nil, ...`

sdk/communication/AzureCommunicationChat/README.md
tjprescott: I don't necessarily think `client.get(thread: String, ...)` is unintuitive. In Storage we have `get(container: String, ...)` which is a name. I don't think the customer would ever assume that the noun referred to the strong object type, since if you had that object, you wouldn't need to `get` it.

sdk/communication/AzureCommunicationChat/README.md
tjprescott: The more I see this, the more I prefer the first option. We don't use `getContainer(withName: String,...)` in Storage, so I'm not sure why we would here. In the case of Storage, the identifier of a container is a name--in the case of Chat, it's an id.

sdk/communication/AzureCommunicationChat/README.md
tjprescott: I agree that the identifying parameter should come first. So `update(message: String, withContent: ...)` I pulled use of content from lines 369-370. I think `update(message:with:)` would also be an acceptable format for updates.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
knvsl: Updated to return ChatThread instead of the client. I believe there is upcoming work for the SDKs to refactor this anyways, we can deal with any new changes then

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: Yah, on second look I just changed this as well as the SendReadReceipt scenario

sdk/communication/AzureCommunicationChat/README.md
knvsl: ^this is correct, and so the customer can call listThreads() without options.

sdk/communication/AzureCommunicationChat/README.md
knvsl: Updated, thanks :)

sdk/communication/AzureCommunicationChat/README.md
knvsl: As discussed offline I've update to follow `get(thread: String)`, using internal names where appropriate. There will be some exceptions in the case of the list methods as well as sendReadReceipt and sendTypingNotification

PR #586, opened 2020-12-07
Renaming fix for maxpagesize in list threads

knvsl:


knvsl:
failing due to known issue with obj-c tests

PR #596, opened 2020-12-10
AzureCore refactor to support Autorest.Swift

tjprescott:
Updates to AzureCore to support Autorest.Swift changes.


sdk/communication/AzureCommunicationChat/Package.swift
```diff
@@ -24,7 +24,7 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/Azure/azure-sdk-for-ios.git",
-            .revision("3e3c80d60173613c8dd4cb6b219188cf5070e8e7")
+            .branch("dev/AzureCore")
```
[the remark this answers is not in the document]
bsiegel: In fact, I didn't realize this file existed. Do we need it? It seems like the repo should have only a single Package.swift, located in the root.

sdk/communication/AzureCommunicationChat/Source/Custom.swift
```diff
@@ -0,0 +1,29 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
```
sacheu, line 28: What is the purpose of this generated file?

sdk/communication/AzureCommunicationChat/Source/Generated/AzureCommunicationChatClient.swift
```diff
@@ -70,53 +65,7 @@ public final class AzureCommunicationChatClient: PipelineClient, PageableClient
...
-            value: value?.addingPercentEncoding(withAllowedCharacters: .azureUrlQueryAllowed)
-        ) }
-        urlComps.percentEncodedQueryItems = queryItems
-        return urlComps.url
-    }
+    public lazy var azureCommunicationChatService = AzureCommunicationChatService(client: self)
 
-    public lazy var azureCommunicationChatService: AzureCommunicationChatService =
-        AzureCommunicationChatService(client: self)
+    // MARK: Public Client Methods
```
sacheu, line 70: Do we need this comment? Seem like the next link is "}" no method is followed by this line.

sacheu (approved):
Looks good. Just a few comments/questions.

sdk/communication/AzureCommunicationChat/Source/Generated/AzureCommunicationChatClient.swift
tjprescott: Good point. There would be public methods if there was no command group, since then the operations would live here. That would be a change in autorest.

sdk/communication/AzureCommunicationChat/Source/Custom.swift
tjprescott: It has to do with a change I made in the autorest repo that involves wiping out the Generated directory before regenerating the code (because old files could linger). It would result in there being no source files and the process would fail at some point. Creating this dummy stub file fixes the problem.

sdk/communication/AzureCommunicationChat/Package.swift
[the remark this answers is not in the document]
tjprescott: It is generated and would be used in the case where you generated an SDK to live in your own repo. So in that sense, I suppose it does make sense to retain this since in that instance, the SDK would not be packaged with AzureCore.

sdk/communication/AzureCommunicationChat/Source/Custom.swift
sacheu: If this dummy stub file prevents the process fails at some point, can this dummy stub files removed at the end of process?

tjprescott: Potentially. I would have some concerns about touching files outside the generated folder when the "contract" is that anything outside the generated folder is safe. I'll take a look at it on the autorest side.

PR #602, opened 2020-12-14
Add live tests to Chat SDK

knvsl:
- Added live tests for ChatClient and ChatThreadClient
- I have only added these to the AzureCommunicationChat scheme and not all "all" scheme yet. Will need to retrieve & set the credentials to run them on pipeline


Package.swift
```diff
@@ -78,6 +78,12 @@ let package = Package(
                 "Tests/ObjCTokenParserTests.m"
             ],
             sources: ["Tests"]
+        ),
+        .testTarget(
+            name: "AzureCommunicationChatTests",
+            dependencies: ["AzureCommunication", "AzureCommunicationChat", "OHHTTPStubsSwift"],
```
knvsl, line 84: Will remove OHHTTPStubs for now - I was using this to support record/playback mode, this PR just supports live mode for now.

tjprescott (changes requested):
You will not be able to merge these to master because they will fail. You'd either need some kind of recording/mocking framework or put these into an example app instead.

tjprescott:
Here's an example of a demo app I made for Storage. You could do something similar: https://github.com/tjprescott/azure-sdk-for-ios/tree/dev/AzureStorageBlob/examples/AzureStorageBlobDemo

knvsl:
> You will not be able to merge these to master because they will fail. You'd either need some kind of recording/mocking framework or put these into an example app instead.

I have a branch where I'm using OHTTPStubs to record/playback but was going to merge this first. The tests will fail without the credentials but I think the goal was to have the pipeline set those so the tests can run. If that is not possible then we could exclude them from the pipeline tests?

PR #604, opened 2020-12-15
Add unit tests for ACS chat service

Leoaqr:


sdk/communication/AzureCommunicationChat/Tests/ChatClientTests.swift
```diff
@@ -0,0 +1,175 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import XCTest
+import AzureCommunication
+import AzureCommunicationChat
+
+class ChatClientTests: XCTestCase {
```
knvsl, line 31: This file should be removed - not part of unit tests right?

Leoaqr: Yes, removed

JoshuaLai:
It looks like the `swift build` command fails because it can't find the new OHHTTPStubs.
I think you just need to add this to the package file under dependencies by being more explicit. 
https://forums.swift.org/t/unable-to-import-multiple-targets-from-external-dependency/39030
Try something like this

PR #607, opened 2020-12-17
Add integration tests with record/playback mode

knvsl:
- In TestSetup.swift we can specify mode to playback, record or live
- Defaults to playback mode, to run live or re-record we just swap out the mock credentials for real ones
- Note some of the recorded responses are empty, this is expected
- For now I'm using OHTTPStubs (same library in unit tests) to send back the recorded response, recording is just a simple file write. In the future can look at using UrlProtocol/UrlSession to remove the explicit calls to record


Leoaqr:
a question: do we need to sanitize id in the recorded files?

Leoaqr:
rest LG2M

Leoaqr (approved):

knvsl:
> a question: do we need to sanitize id in the recorded files?

Added sanitization to remove any ids, thank you for pointing this out!!

PR #611, opened 2020-12-22
[Communication] - Identifiers - Adding new Microsoft Teams User identifier

JoshuaLai:


sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -82,3 +82,19 @@ import Foundation
         self.value = phoneNumber
     }
 }
+/**
+ Communication identifier for Microsoft Teams Users
+ */
+public class MicrosoftTeamsUser: NSObject, CommunicationIdentifier {
```
DominikMe, line 88: In a follow up PR we should rename all identifiers to have the *Identifier suffix as name. Please sync up with Minnie.

DominikMe (approved):
Thanks Josh!

sdk/communication/AzureCommunication/Source/Identifiers.swift
JoshuaLai: @DominikMe @minnieliu Yes we have a task for that to apply arch board feedback. 
This will come in an upcoming PR.

PR #615, opened 2021-01-04
[Communication] - Auth - Introducing CommunicationTokenRefreshOptions

JoshuaLai:
As per arch review. We want to encapsulate CommunicationUserCredential params into TokenRefreshOptions.

### DIFF
--- a/sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
+++ b/sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
@@ -26,6 +26,7 @@
 
 import Foundation
 
+public typealias TokenRefresherClosure = (@escaping TokenRefreshOnCompletion) -> Void
 /**
  The Communication Token Refresh Options. Used to initialize a `CommunicationUserCredential`
  - SeeAlso: ` CommunicationUserCredential.init(...)`
@@ -33,7 +34,7 @@ import Foundation
 @objcMembers public class CommunicationTokenRefreshOptions: NSObject {
     var initialToken: String?
     var refreshProactively: Bool
-    var tokenRefresher: (@escaping TokenRefreshOnCompletion) -> Void
+    var tokenRefresher: TokenRefresherClosure
     /**
      Initializes a new instance of `CommunicationTokenRefreshOptions`
      The cached token is updated if `token(completionHandler: )` is called and if the difference between the current time
@@ -49,7 +50,7 @@ import Foundation
     public init(
         initialToken: String? = nil,
         refreshProactively: Bool = false,
-        tokenRefresher: @escaping (@escaping TokenRefreshOnCompletion) -> Void
+        tokenRefresher: @escaping TokenRefresherClosure
     ) {
         self.initialToken = initialToken
         self.refreshProactively = refreshProactively
--- a/sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
+++ b/sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
@@ -1,0 +1,58 @@
+// --------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// The MIT License (MIT)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the ""Software""), to
+// deal in the Software without restriction, including without limitation the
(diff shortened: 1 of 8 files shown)


sdk/communication/AzureCommunication/Source/Authentication/CommunicationUserCredential.swift
```diff
@@ -84,3 +80,20 @@ public typealias TokenRefreshOnCompletion = (String?, Error?) -> Void
         userTokenCredential.token(completionHandler: completionHandler)
     }
 }
+
+@objcMembers public class CommunicationTokenRefreshOptions: NSObject {
```
JoshuaLai, line 84: I put this object in `CommunicationUserCredential` but it should be in its own class. 
Should we create a new Utils folder? Cause right now the authentication folder is slowly turning into just a dumping ground for our code and not very well organized

minnieliu: Yea agreed, probably best to move this out to its own file.

sdk/communication/AzureCommunication/Tests/CommunicationUserCredentialTests.swift
```diff
@@ -101,11 +101,12 @@ class CommunicationUserCredentialTests: XCTestCase {
     func test_RefreshTokenProactively_TokenAlreadyExpired() throws {
         let expectation = XCTestExpectation()
 
-        let userCredential = try CommunicationUserCredential(
+        let option = CommunicationTokenRefreshOptions(
```
minnieliu, line 104: Might want to rename this to something a bit more explicit (e.g. `tokenRefreshOptions`)

sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
```diff
@@ -0,0 +1,58 @@
...
+     The cached token is updated if `token(completionHandler: )` is called and if the difference between the current time
+     and token expiry time is less than 120s.
+     If `refreshProactively` parameter  is `true`:
+        - The cached token will be updated in the background when the difference between the current time and token expiry time is less than 600s.
+        - The cached token will be updated immediately when the constructor is invoked and `initialToken` is expired
+     - Parameters:
+        - initialToken: The initial value of the token.
+        - refreshProactively: Whether the token should be proactively refreshed in the background.
+        - tokenRefresher: Closure to call when a new token value is needed.
+     */
```
tjprescott, line 48: So, this is not the correct format for Swift docstrings. You should "right click" the function name in Xcode and choose "Show Code Actions > Add Documentation" to get the correct format.

sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
```diff
@@ -0,0 +1,58 @@
...
+    public init(
+        initialToken: String? = nil,
+        refreshProactively: Bool = false,
+        tokenRefresher: @escaping (@escaping TokenRefreshOnCompletion) -> Void
```
tjprescott, line 52: This is not a syntax I've seen before. Why is the double @escaping needed?

tjprescott (changes requested):
One comment on docstring format and a question.

sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenRefreshOptions.swift
JoshuaLai: I agree this looks a little weird to read. 
`TokenRefreshOnCompletion` is a typealias which is closure: `public typealias TokenRefreshOnCompletion = (String?, Error?) -> Void` 
So `(@escaping TokenRefreshOnCompletion) -> Void` the inner @escaping refers to escaping inner closure. 
And the outer @escaping is escaping the closure `(@escaping TokenRefreshOnCompletion) -> Void`

JoshuaLai: Should I typealias `(@escaping TokenRefreshOnCompletion) -> Void` to SomeTokenRefresher
So then it will look @escaping SomeTokenRefresher and hide it all?

tjprescott: Hmmm... interesting. So the closure takes a closure. If the token refresh closure always needs to be escaping (which i presume it would) then I think it makes sense to move the inner escaping to the typealias---provided the compiler doesn't whine.

tjprescott (approved):

PR #633, opened 2021-01-11
[Communication] - Auth - Removing a duplicate policy and replacing it with bearer auth policy

JoshuaLai:


sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -56,7 +56,8 @@ public class ChatClient {
             throw AzureError.client("Unable to form base URL.")
         }
 
-        let authPolicy = CommunicationUserCredentialPolicy(credential: credential)
+        let communicationCredential = CommunicationPolicyTokenCredential(credential)
```
knvsl, line 59: Just to clarify - if this change is checked into the feature branch, and then the feature branch is updated with master, all that should change is the rename of CommunicationUserCredential -> CommunicationTokenCredential? 

Otherwise if it helps to update the feature branch first we can do that

JoshuaLai: Yes the only change we will need to make is the rename. There shouldn't be any other conflicts

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
```diff
@@ -28,45 +28,33 @@
 import AzureCore
 #endif
 import Foundation
-/**
- The Azure Communication Services authentication policy.
- */
-public class CommunicationUserCredentialPolicy: Authenticating {
-    public var next: PipelineStage?
 
+public class CommunicationPolicyTokenCredential: TokenCredential {
```
sacheu, line 32: Should the file renamed to CommunicationPolicyTokenCredential.swift?

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
```diff
@@ -28,45 +28,33 @@
...
-            completionHandler(request, nil)
+            
+            let accessToken = AccessToken(token: communicationAccessToken.token,
+                                          expiresOn: communicationAccessToken.expiresOn)
+            
+            completionHandler(accessToken, nil)
+        }
+    }
+    
+    public func validate() throws {
```
sacheu, line 55: When will validate() get called? From my understanding, where there is an error, it will call the completionHandler with the error (line 42-44). Why we need to have a property `error` and return it to caller when validate() is called.  This seems duplication of calling the ` completionHandler(nil, self.error)` in line 44

sacheu:
I think we can revert the changes in 

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift 
sdk/communication/AzureCommunicationChat/Source/Generated/ChatThreadOperation.swift

since they are generated code and not part of this PR

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
sacheu, line 32: Missed comment in the class CommunicationPolicyTokenCredential

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
```diff
@@ -28,45 +28,33 @@
...
-        credential.token { token, error in
-            if let error = error {
-                completionHandler(request, AzureError.client("Error while retrieving access token", error))
-                return
-            }
-
-            guard let token = token?.token else {
-                completionHandler(request, AzureError.client("Token cannot be empty"))
+    
+    public func token(forScopes scopes: [String], completionHandler: @escaping TokenCompletionHandler) {
```
sacheu, line 40: Miss comment for the method

JoshuaLai: Added comment block

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
JoshuaLai: I am not sure when validate is called. I see it is a required function in the protocol. I called the completionHandler with the error cause that should be convention so the client won't just hang there.

sdk/communication/AzureCommunication/Source/Authentication/CommunicationAuthenticationPolicy.swift
JoshuaLai: Added comment block

JoshuaLai: Yes I should update the name to the file good catch

JoshuaLai:
> I think we can revert the changes in
> 
> sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
> sdk/communication/AzureCommunicationChat/Source/Generated/ChatThreadOperation.swift
> 
> since they are generated code and not part of this PR

@sacheu I actually think we should keep these. I know they are just white spaces changes but I think fixing the format of the file to be consistent is important

AzureSDK.xcworkspace/xcshareddata/xcschemes/all.xcscheme
```diff
@@ -146,6 +146,20 @@
                ReferencedContainer = "container:sdk/storage/AzureStorageBlob/AzureStorageBlob.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
```
JoshuaLai, line 149: I noticed the Chat target was not added to `all` scheme so it wasn't building. 
I added it here

tjprescott (changes requested):
Please add a changelog entry and ensure this is documented as a breaking change.

tjprescott:
Added @raosanat as a review to be aware of a breaking change in AzureCommunication. I don't think it should impact Calling since that doesn't use the pipeline, but he should be aware.

JoshuaLai:
> Please add a changelog entry and ensure this is documented as a breaking change.

Yep that makes sense. I wasn't sure if we wanted to add this right away because this is a draft PR and going into a feature branch.

JoshuaLai:
> Please add a changelog entry and ensure this is documented as a breaking change.

@tjprescott Updated the change log. This feature branch is a little out of date so we will need to make sure it is cleaned up when the feature branch is merged into master.

knvsl:
> > I think we can revert the changes in
> > sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
> > sdk/communication/AzureCommunicationChat/Source/Generated/ChatThreadOperation.swift
> > since they are generated code and not part of this PR
> 
> @sacheu I actually think we should keep these. I know they are just white spaces changes but I think fixing the format of the file to be consistent is important

Did you need to manually change the spacing? Just noting if its not from linting then every time we regenerate Chat the spacing would get reverted again :D

JoshuaLai:
> > > I think we can revert the changes in
> > > sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
> > > sdk/communication/AzureCommunicationChat/Source/Generated/ChatThreadOperation.swift
> > > since they are generated code and not part of this PR
> > 
> > 
> > @sacheu I actually think we should keep these. I know they are just white spaces changes but I think fixing the format of the file to be consistent is important
> 
> Did you need to manually change the spacing? Just noting if its not from linting then every time we regenerate Chat the spacing would get reverted again :D

@knvsl No I didn't do it manually. I would have been from my linter.

PR #639, opened 2021-01-14
Add smoke test project

tjprescott:
Adds a simple smoke test project that will get the latest unreleased libraries and attempts to build an app that imports them all. In this way, we can identify incompatibilities before releasing.


examples/AzureSDKSmokeTest/AzureSDKSmokeTest/AppDelegate.swift
```diff
@@ -0,0 +1,59 @@
...
+// --------------------------------------------------------------------------
+
+import UIKit
+
+// Import all relevant, released SDKs here
+import AzureCore
+import AzureCommunication
+import AzureCommunicationChat
+import AzureCommunicationCalling
+
```
tjprescott, line 34: This is really the only relevant code.

.gitignore
```diff
@@ -69,4 +69,5 @@ fastlane/screenshots/**/*.png
 fastlane/test_output
 
 # Random Mac files.
-.DS_Store
\ No newline at end of file
+.DS_Store
+Podfile.lock
```
tjprescott, line 73: @bsiegel thoughts on this? I don't see any value in committing this file since everyone would pretty much need to run `pod install` locally anyhow

examples/AzureSDKSmokeTest/Podfile
```diff
@@ -0,0 +1,41 @@
...
+
+platform :ios, '12.0'
+use_frameworks!
+
+target 'AzureSDKSmokeTest' do
+  pod 'AzureCore', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+  pod 'AzureCommunication', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+  pod 'AzureCommunicationChat', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+  pod 'AzureCommunicationCalling', :podspec => 'https://raw.githubusercontent.com/Azure/azure-sdk-for-ios/master/AzureCommunicationCalling.podspec.json'
+
```
tjprescott, line 36: This is also relevant.

JoshuaLai:
So just a general question. Is the idea to now have the pipeline also build the new target "AzureSDKSmokeTest" and look for failures when building calling?

tjprescott:
> So just a general question. Is the idea to now have the pipeline also build the new target "AzureSDKSmokeTest" and look for failures when building calling?

The idea right now is that we'll run this before a release and then, when we have a release pipeline, it will be a required gate on proceeding with that pipeline.

tjprescott:
Starting point for release documentation is here: https://github.com/Azure/azure-sdk-for-ios/wiki/Release-Process

.gitignore
bsiegel: Agreed, especially since we want to install pods fresh each time the CI runs this.

examples/AzureSDKSmokeTest/Podfile
```diff
@@ -0,0 +1,36 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+#
+# --------------------------------------------------------------------------
+
+platform :ios, '12.0'
+use_frameworks!
```
bsiegel, line 29: I'd like to see if we can run two tests here, one with this line and one without (without means pods will be built as static libraries). It took me some messing about with the podspecs in order to ensure that our pods work in both modes, and it would be nice to validate that.

bsiegel (approved):
Looks good, added one request / piece of feedback.

examples/AzureSDKSmokeTest/Podfile
```diff
@@ -26,7 +26,10 @@
 # --------------------------------------------------------------------------
 
 platform :ios, '12.0'
-use_frameworks!
+
+if ENV['SWIFT_USE_FRAMEWORKS'] != nil
+  use_frameworks!
+end
```
tjprescott, line 33: @bsiegel this will allow us to run `pod install` with an environment variable set to test static and dynamic modes. I have tested both and they work. The other option would be to have two Podfiles which you'd have to swap names to use (since you can't configure the name of the podfile) or simply two separate projects. This seemed the simplest path. Let me know what you think.

examples/AzureSDKSmokeTest/Podfile
tjprescott: I use a conditional environment variable switch to control this. Let me know what you think.

examples/AzureSDKSmokeTest/Podfile
bsiegel: Yep, that appears to work just fine. 👍

bsiegel (approved):

PR #640, opened 2021-01-14
[Communication] - Identifiers - Teams property renamed to userId

JoshuaLai:
Renaming from `identifier` to `userId` in case `identifier` needs to be used for different purpose in the future. Specifically only for `MicrosoftTeamsUserIdentifier`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.0.0-beta.8 (Unreleased)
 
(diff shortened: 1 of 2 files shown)


minnieliu (approved):
LGTM!

JoshuaLai:
@raosanat I am pretty sure Calling isn't using this new Teams Identifier yet. But just an FYI.

PR #643, opened 2021-01-20
[Communication] Add CommunicationIdentifierSerializer to serialize/deserialize CommunicationIdentifierModel

sacheu:
1. Add CommunicationIdentifierSerializer 
* to deserialize CommunicationIdentifierModel to CommunicationIdentifier
* to serialize CommunicationIdentifier to CommunicationIdentifierModel

2. Enumerations.swift and CommunicationIdentifierModel.swift are generated code from AutoRest code from https://github.com/Azure/azure-rest-api-specs/blob/master/specification/communication/data-plane/Microsoft.CommunicationServicesCommon/preview/2020-11-19-preview1/common.json.

These generated code are put into /Users/sacheu/Documents/GitHub/azure-sdk-for-ios/sdk/communication/AzureCommunication/Source temporarily. They will be in chat preview 4 swagger. These files will replaced once chat team starts working on preivew 4 swagger and the generated chat sdk will contain these file

3. Rename "Rename PhoneNumberIdentifier.Value  to PhoneNumberIdentifier.PhoneNumber"

This is the corresponding change in dotnet sdk
 https://github.com/Azure/azure-sdk-for-net/pull/18052


sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
```diff
@@ -0,0 +1,82 @@
...
+        self.kind = kind
+        self.id = id
+        self.phoneNumber = phoneNumber
+        self.microsoftTeamsUserId = microsoftTeamsUserId
+        self.isAnonymous = isAnonymous
+    }
+
+    // MARK: Codable
+
+    enum CodingKeys: String, CodingKey {
```
JoshuaLai, line 55: Do you need to set this? 
It looks like you there is no change in coding keys names.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
```diff
@@ -0,0 +1,82 @@
...
+// swiftlint:disable cyclomatic_complexity
+
+/// Identifies a participant in Azure Communication services. A participant is, for example, a phone number or an Azure communication user.
+public struct CommunicationIdentifierModel: Codable {
+    // MARK: Properties
+
+    /// Kind of the communication identifier.
+    public let kind: CommunicationIdentifierKind
+    /// Full Id of the identifier.
+    public let id: String?
```
JoshuaLai, line 25: Although not a super big deal `id` is a reserved type in objective-c. I think we should change this to `identifier`. 
We do the same thing in the identifiers class.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
JoshuaLai: If we change `id` to `identifier` we will need to set coding keys.

sacheu: Enumerations.swift and CommunicationIdentifierModel.swift are generated code from AutoRest code.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
sacheu: Enumerations.swift and CommunicationIdentifierModel.swift are generated code from AutoRest swift.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
```diff
@@ -0,0 +1,82 @@
...
+    enum CodingKeys: String, CodingKey {
+        case kind = "kind"
+        case id = "id"
+        case phoneNumber = "phoneNumber"
+        case microsoftTeamsUserId = "microsoftTeamsUserId"
+        case isAnonymous = "isAnonymous"
+    }
+
+    /// Initialize a `CommunicationIdentifierModel` structure from decoder
+    public init(from decoder: Decoder) throws {
```
JoshuaLai, line 64: Since this is a struct and there doesn't look like there is complex nesting of objects. I don't think we need this `init`

sdk/communication/AzureCommunication/Source/CommunicationIdentifierSerializer.swift
```diff
@@ -0,0 +1,74 @@
...
+    import AzureCore
+#endif
+import Foundation
+
+public class CommunicationIdentifierSerializer {
+    static func deserialize(identifier: CommunicationIdentifierModel) throws -> CommunicationIdentifier {
+        let kind = identifier.kind
+
+        guard let id = identifier.id else {
+            throw AzureError.client("Can't serialize CommunicationIdentifierModel to CommunicationIdentifier.")
```
JoshuaLai, line 37: Do we want to be more specific with this error?
`"Can't serialize CommunicationIdentifierModel id of undefined."`

JoshuaLai (approved):
I just had one nit comment on error message but otherwise looks good!

PR #653, opened 2021-01-26
Build upgrade

bsiegel:
- Updated / standardized project configurations
  - This mostly involved applying Xcode recommended settings and a bunch of cleanups for the AzureCommunicationChat project since it was generated with many warnings disabled, was missing from various schemes, and was not correctly targeting iOS as the primary build target
- Updated build tools versions
- Updated MSAL dependency in Podfile
- Applied SwiftLint autofixes
- Significant improvements to build process
  - I added build steps to generate xcframeworks with the correct debugging symbols included and publish those as another artifact.
  - I broke out the demo project and the smoke test into their own jobs. The smoke tests seem to require their own xcworkspace and pod install, which makes sense since the idea is they are isolated and meant to model how a customer would acquire our libraries.


AzureSDK.xcworkspace/contents.xcworkspacedata
```diff
@@ -49,9 +49,6 @@
    <Group
       location = "group:examples"
       name = "Examples">
-      <FileRef
-         location = "group:AzureSDKSmokeTest/AzureSDKSmokeTest.xcodeproj">
-      </FileRef>
```
tjprescott, line 54: Why was this removed? It should appear under Examples in the AzureSDK workspace.

Package.swift
```diff
@@ -46,32 +46,41 @@ let package = Package(
...
         ),
         // Test targets
         .testTarget(
             name: "AzureCoreTests",
             dependencies: ["AzureCore"],
             path: "sdk/core/AzureCore",
+            exclude: [
+              "Tests/Info.plist",
+              "Tests/Data Files",
+              "Tests/test.playground"
```
tjprescott, line 74: Why do we have a `test.playground`?

Podfile.lock
```diff
@@ -1,18 +1,18 @@
 PODS:
```
tjprescott, line 1: I thought we added podfile.lock to the git ignore?

tjprescott (commented):
LGTM. A couple questions.

AzureSDK.xcworkspace/contents.xcworkspacedata
bsiegel: I removed it from the AzureSDK workspace because it didn't appear to build in that workspace. You have to go into the AzureSDKSmokeTest directory, run `pod install` there to install the pods, and then open `AzureSDKSmokeTest.xcworkspace` and build it from there, because the pod search directory is relative to the workspace.

Package.swift
bsiegel: That's a great question, I don't know :p When I updated to the Swift 5.3 toolchain, `swift build` started failing because it doesn't seem to ignore as many file types by default as the older toolchain, and these were files I needed to exclude to get the build working again. If we don't need it we could just delete the playground.

Podfile.lock
bsiegel: I think we discussed it but we didn't actually do it. If we want to we can, but we should also remove Podfile.lock from the repo too.

Package.swift
tjprescott: Let's delete the playground. We could also add that to the gitignore.

Podfile.lock
tjprescott: Let's do it. It may be buried in the Storage PR.

AzureSDK.xcworkspace/contents.xcworkspacedata
bsiegel: We can revisit whether the smoke test belongs in the examples/ directory in a future PR. For now I am going to keep it there.

Issue #644, opened 2021-01-20
[FEATURE REQ] PagedCollection supports ability to map from one type to another


knvsl:
**Is your feature request related to a problem? Please describe.**
Related to Chat SDK. 

The SDK wants to return custom models instead of the models generated by autorest. 
For example instead of returning the autorest generated ChatMessage, we would like to return a custom model, Message.

In the convenience layer we can map a generated model to an SDK defined model, but this is not easy to do when the generated code returns `PagedCollection<SomeGeneratedModel>`

**Describe the solution you'd like**
The goal is to be able to  map `PagedCollection<TypeA>` to `PagedCollection<TypeB>`, possibly by providing a mapping function to PagedCollection. Chat SDK would do the mapping in the convenience layer.

**Describe alternatives you've considered**

- Changes to autorest.swift to support customizing generated code.
- In the convenience layer - throw away the `PagedCollection<GeneratedModel>` returned by the generated code and construct a new `PagedCollection<CustomSDKModel>` from the data within the httpResponse, return this instead.

PR #655, opened 2021-01-26
Add cloud environment to MicrosoftTeamsIdentifier

sacheu:
* Add CommunicationCloudEnvironment class
* CommunicationCloudEnvironmentModel is a generated enum
* Add serialize/deserialize code to convert CommunicationCloudEnvironment from/to CommunicationCloudEnvironmentModel
* Add CommunicationCloudEnvironment to MicrosoftTeamsIdentifier

Corresponding .net implementation is  https://github.com/Azure/azure-sdk-for-net/pull/18120

PR #656, opened 2021-01-26
Chat SDK uses Communication identifiers

knvsl:
Chat convenience layer should return id's as CommunicationUserIdentifier's instead of strings.

- Introduced custom models that use the communication identifier
- Convenience layer calls the generated code and receives a response in the format of one of the generated models, it now maps the generated model to the custom model that uses identifiers and returns that instead
- For PagedCollection have gone with the workaround of throwing away the paged collection that the generated code returns and constructing a new one from the data within the httpResponse: https://github.com/Azure/azure-sdk-for-ios/issues/644


sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -276,30 +330,53 @@ public class ChatThreadClient {
...
     public func listMessages(
         withOptions options: ChatThreadOperation.ListChatMessagesOptions? = nil,
-        completionHandler: @escaping HTTPResultHandler<PagedCollection<ChatMessage>>
+        completionHandler: @escaping HTTPResultHandler<PagedCollection<Message>>
     ) {
         service.listChatMessages(chatThreadId: threadId, withOptions: options) { result, httpResponse in
             switch result {
-            case let .success(messages):
-                completionHandler(.success(messages), httpResponse)
+            case .success: https: // github.com/Azure/azure-sdk-for-ios/issues/644
```
sacheu, line 337: https: should be in a comment.

sdk/communication/AzureCommunication/Source/Identifiers.swift
```diff
@@ -33,7 +33,7 @@ import Foundation
 /**
  Communication identifier for Communication Services Users
  */
-@objcMembers public class CommunicationUserIdentifier: NSObject, CommunicationIdentifier {
+@objcMembers public class CommunicationUserIdentifier: NSObject, Codable, CommunicationIdentifier {
```
JoshuaLai, line 36: @sacheu this introduction of `Codable` careful need to take into consideration for https://github.com/Azure/azure-sdk-for-ios/pull/655

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
knvsl: oh weird, fixed. thanks!

sdk/communication/AzureCommunication/Source/Identifiers.swift
knvsl: I think I can actually revert this, when encoding/decoding the custom models not actually encoding/decoding CommunicationUserIdentifier

glorialimicrosoft (approved):

PR #658, opened 2021-01-27
[Communication] - Tests - Reenable the objc unit tests

JoshuaLai:


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -172,6 +172,7 @@ jobs:
         displayName: 'Install dependencies'
 
       - script: |
+          xcrun simctl delete unavailable
```
sacheu, line 175: what is the step for?

sdk/communication/AzureCommunication/Tests/CommunicationTokenCredentialTests.swift
```diff
@@ -45,7 +45,8 @@ class CommunicationTokenCredentialTests: XCTestCase {
         "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjEwMH0.1h_scYkNp-G98-O4cW6KvfJZwiz54uJMyeDACE4nypg"
 
     private var fetchTokenCallCount: Int = 0
-
+    private let timeout: TimeInterval = 18.0
```
sacheu, line 48: why 18? instead of a round number? why we need to bump up the time out to a such a high number , 18?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
```diff
@@ -73,18 +74,26 @@ - (void)xtest_ObjCRefreshTokenProactivelyTokenExpiringInOneMin {
         XCTAssertEqual(accessToken.token, weakSelf.sampleToken);
         XCTAssertEqual(weakSelf.fetchTokenCallCount, 1);
         
-        [expectation fulfill];
```
sacheu, line 76: if we don't call expection fulfill, are we removing expectation? I don't see the removal above line 76.

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
```diff
@@ -106,10 +115,18 @@ - (void)xtest_ObjCRefreshTokenProactivelyTokenExpiringInNineMin {
...
+        isComplete = YES;
     }];
 
-    [self waitForExpectations:@[expectation] timeout:2.0];
+    
+    NSDate *loopUntil = [NSDate dateWithTimeIntervalSinceNow: self.timeout];
+    while (isComplete == NO && [loopUntil timeIntervalSinceNow] > 0) {
```
sacheu, line 123: does this mean we will keep looping for 18 sec? if the flag isComplete is set the Yes, the test passed?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
```diff
@@ -106,10 +115,18 @@ - (void)xtest_ObjCRefreshTokenProactivelyTokenExpiringInNineMin {
         XCTAssertEqual(accessToken.token, weakSelf.sampleToken);
         XCTAssertEqual(weakSelf.fetchTokenCallCount, 1);
         
-        [expectation fulfill];
+        isComplete = YES;
```
sacheu, line 118: nit: should this TRUE or FALSE instead?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
JoshuaLai: This is used to close any simulators currently open on the CI virtual machine.

sdk/communication/AzureCommunication/Tests/CommunicationTokenCredentialTests.swift
JoshuaLai: I don't think it needs to be so high I was just trying large numbers to make sure the failures were not because of timeouts

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
JoshuaLai: Yes I changed up the implementation to not rely on `expectation`. Using some advice from Java and .net platform. I tried manipulating the run loop up until  the `loopUntil` similar to the custom clock in .net.

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
JoshuaLai: In ObjC it is `YES` `NO`. I think you can use `TRUE` `FALSE` but they are just aliases for YES and NO

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
JoshuaLai: Yes.

sacheu: inside the while loop, does any of the function is a blocking function? mean within this 18 sec will the CPU goes up to 100% inside this while loop?

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
sacheu: should there be code to remove the declaration of 'expectation'? something like this 
image

sdk/communication/AzureCommunication/Tests/ObjCCommunicationTokenCredentialAsyncTests.m
JoshuaLai: When `isComplete = YES` it will break out of the loop. Or if the `loopUntil` has exceeded the timeout

sdk/communication/AzureCommunication/Tests/CommunicationTokenCredentialTests.swift
sacheu: can we try with 10 instead of 18?

sacheu (approved):
Please try reducing the timeout to 10. Otherwise, looks good.

PR #673, opened 2021-02-02
[Communication] - Identifiers - Renamed id to be identifiers to be consistent

JoshuaLai:


JoshuaLai:
Misunderstanding on this PR so closing it. I will open a new one with the new name

Issue #309, opened 2020-05-18
Add SwiftUI example project


tjprescott:
We have a sample project for traditional UIKit development using storyboards. We should have a SwiftUI-based example project for those who prefer that approach.

osca2000:
No update yet?

tjprescott:
@osca2000 at the moment we are focused on readying and releasing our AzureCore library for preview, so no update on this right now.

tjprescott:
This was completed with #675.

PR #675, opened 2021-02-03
[Issues-309] - Example - Fleshing out swift ui demo 

JoshuaLai:
https://github.com/Azure/azure-sdk-for-ios/issues/309


Podfile
```diff
@@ -65,3 +65,8 @@ target 'AzureStorageBlobDemo' do
   project 'examples/AzureStorageBlobDemo/AzureStorageBlobDemo'
   pod 'MSAL', '~> 1.1.2'
 end
+
+target 'AzureSDKDemoSwiftUI' do
+  project 'examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI'
+  pod 'MSAL', '~> 1.1.2'
```
tjprescott, line 71: We can remove the MSAL dependency for this. When/if MSAL is Swift-UI friendly, we can reintroduce it.

Podfile.lock
```diff
@@ -0,0 +1,18 @@
+PODS:
```
tjprescott, line 1: This file shouldn't be committed. It should be in the gitignore on master and might have been undone in the rebase. Feel free to re-add  it.

examples/AzureSDKDemoSwift/Source/Base.lproj/Main.storyboard
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="16097" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="obV-HT-tcp">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="obV-HT-tcp">
```
tjprescott, line 2: I wouldn't expect any diff to this file since this is the non-SwiftUI demo. If inadvertent we can discard these changes.

examples/AzureSDKDemoSwift/Source/BlobDownloadViewController.swift
```diff
@@ -134,9 +134,9 @@ extension BlobDownloadViewController: UITableViewDelegate, UITableViewDataSource
     internal func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cellIdentifier = "CustomTableViewCell"
         guard let blobClient = blobClient,
-            let data = dataSource?.items,
-            let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier, for: indexPath)
-            as? CustomTableViewCell
+              let data = dataSource?.items,
```
tjprescott, line 137: Same. There shouldn't be any diff for the AzureSDKDemoSwift project.

examples/AzureSDKDemoSwift/Source/BlobUploadViewController.swift
```diff
@@ -168,8 +168,8 @@ extension BlobUploadViewController: UICollectionViewDelegate, UICollectionViewDa
     ) -> UICollectionViewCell {
         let cellIdentifier = "CustomCollectionViewCell"
         guard let blobClient = blobClient,
-            let cell = collectionView.dequeueReusableCell(withReuseIdentifier: cellIdentifier, for: indexPath)
-            as? CustomCollectionViewCell
+              let cell = collectionView.dequeueReusableCell(withReuseIdentifier: cellIdentifier, for: indexPath)
```
tjprescott, line 171: Discard this change as well.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  AppDelegate.swift
+//  AzureSDKDemoSwiftUI
+//
+//  Created by Joshua Lai on 2021-01-25.
+//
```
tjprescott, line 6: This should use the same license header as other files.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,131 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import SwiftUI
+
+import AzureCore
+import MSAL
```
tjprescott, line 30: We can get rid of the MSAL import.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,131 @@
...
+struct ContentView: View {
+    @State private var isLoggedIn = false
+    @State private var isShowingMedia = false
+    
+    @State private var isError = false
+    
+    @State private var userLabel: String = "Please log in"
+    @State private var errorMessage: String = ""
+    
+    var body: some View {
```
tjprescott, line 41: I would recommend commenting out the stuff that isn't exercised with a TODO comment about reenabling once MSAL is supported.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobDownloadView.swift
```diff
@@ -0,0 +1,150 @@
...
+//
+// --------------------------------------------------------------------------
+
+import SwiftUI
+import UIKit
+
+import AzureCore
+import AzureIdentity
+import AzureStorageBlob
+import MSAL
```
tjprescott, line 33: Scratch MSAL.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobDownloadView.swift
```diff
@@ -0,0 +1,150 @@
...
+
+import AVKit
+import Photos
+
+final class BlobDownloadTableViewController: UIViewControllerRepresentable, MSALInteractiveDelegate {
```
tjprescott, line 38: Let's get rid of the MSAL delegate stuff and make this like a regular SwiftUI View struct. That will yield the most "true" SwiftUI experience and we can evaluate future enhancements to MSAL against the desire to not deviate from this experience.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobListObservable.swift
```diff
@@ -0,0 +1,114 @@
...
+// --------------------------------------------------------------------------
+
+import Foundation
+
+import AzureCore
+import AzureIdentity
+import AzureStorageBlob
+import MSAL
+
+class BlobListViewModel {
```
tjprescott, line 34: Does this need to be a class? Does it work as a struct?

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobListObservable.swift
```diff
@@ -0,0 +1,114 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+import AzureCore
+import AzureIdentity
+import AzureStorageBlob
+import MSAL
```
tjprescott, line 32: Scratch MSAL import.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/MediaView.swift
```diff
@@ -0,0 +1,56 @@
...
+                        .tag(0)
+                    Text("Tab Content 2")
+                        .tabItem {
+                            Image(systemName: "square.and.arrow.up")
+                            Text("Upload")
+                        }
+                        .tag(1)
+                })
+    }
+}
```
tjprescott, line 50: Per your comment form the meeting, we can scratch the tab view.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/SceneDelegate.swift
```diff
@@ -0,0 +1,63 @@
+//
+//  SceneDelegate.swift
+//  AzureSDKDemoSwiftUI
+//
+//  Created by Joshua Lai on 2021-01-25.
+//
```
tjprescott, line 6: Should be the standard license header.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/AppDelegate.swift
```diff
@@ -0,0 +1,36 @@
...
+//  AppDelegate.swift
```
tjprescott, line 2: We we need an AppDelegate since we have a SceneDelegate, which is the key SwiftUI concept?

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Utils/Common.swift
```diff
@@ -0,0 +1,183 @@
...
+    static let storageAccountUrl = URL(string: "https://iosdemostorage1.blob.core.windows.net/")!
+    static let tenant = "7e6c9611-413e-47e4-a054-a389854dd732"
+    static let clientId = "6f2c62dd-d6b2-444a-8dff-c64380e7ac76"
+    static let redirectUri = "msauth.com.azure.examples.AzureSDKDemoSwift://auth"
+    static let authority = "https://login.microsoftonline.com/7e6c9611-413e-47e4-a054-a389854dd732"
+    static let uploadContainer: String! = "uploads"
+    static let videoContainer: String = "videos"
+    
+    // swiftlint:disable line_length
+    static let sasConnectionString =
```
tjprescott, line 43: Add a comment here that this string is a read-only SAS token and may need to be regenerated.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Utils/Common.swift
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    internal func showAlert(message: String) {
+        guard presentedViewController == nil else { return }
+        let alertController = UIAlertController(title: "Blob Contents", message: message, preferredStyle: .alert)
+        let defaultAction = UIAlertAction(title: "Close", style: .default, handler: nil)
+        alertController.addAction(defaultAction)
+        present(alertController, animated: true)
+    }
+}
```
tjprescott, line 159: Hopefully this can be repurposed into something easy to display a quick SwiftUI alert view.

sdk/communication/AzureCommunicationChat/Source/Generated/AzureCommunicationChatService.swift
```diff
@@ -45,7 +45,7 @@ public final class AzureCommunicationChatService {
         // Construct request
         let urlTemplate = "/chat/threads/{chatThreadId}/readreceipts"
         guard let requestUrl = client.url(host: "{endpoint}", template: urlTemplate, params: params),
-              let request = try? HTTPRequest(method: .get, url: requestUrl, headers: params.headers)
+            let request = try? HTTPRequest(method: .get, url: requestUrl, headers: params.headers)
```
tjprescott, line 48: This should be reverted.

sdk/storage/AzureStorageBlob/Source/DataStructures/TransferCollection.swift
```diff
@@ -121,18 +121,18 @@ public struct TransferCollection {
 
         if let localUrl = localUrl {
             guard let url = transfer.transferType == .download ? transfer.destination : transfer.source,
-                url == localUrl.rawUrl else { return false }
+                  url == localUrl.rawUrl else { return false }
```
tjprescott, line 124: This file should be reverted.

tjprescott (changes requested):
Looks good! Some comments and questions to get it ready to merge. Thanks for taking this on!

examples/AzureSDKDemoSwift/Source/Base.lproj/Main.storyboard
JoshuaLai: Yes. Changes were implemented when I open the storyboard file in Xcode. I will revert

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/AppDelegate.swift
JoshuaLai: This was automatically generated. Yes I think you are right we just need the scene

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobListObservable.swift
JoshuaLai: At the time being it is a class cause it made more sense for performance. 
The `downloadProgress` is called frequently and can hold onto state in the transfers dictionary. 
When I change this to a pure swift implementation I might be able to make this a struct using @Publish vars 
It will be on my radar

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Utils/Common.swift
JoshuaLai: I actually copied a lot of this from the original demo project. I will add it to my to do to clean this up and remove anything that is unused.

Podfile
JoshuaLai: So it turns out I actually need MSAL. 
https://github.com/Azure/azure-sdk-for-ios/pull/675/files#diff-8abc0fc0a497bc17bcf95e2e95626b721c7f3d3f52b7a670dcb1fb53e6333839R86
I need it to actually create the blob storage client

tjprescott: You can create the client using the `StorageBlobClient.init(connectionString:withOptions:)` initializer instead. That's why it has to be read-only connection string.

examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Utils/Common.swift
```diff
@@ -0,0 +1,106 @@
...
+        guard AppState.internalBlobClient != nil else {
+            throw error
+        }
+
+        let client = AppState.internalBlobClient!
+        client.delegate = delegate
+        StorageBlobClient.maxConcurrentTransfers = 4
+        return client
+    }
+}
```
tjprescott, line 106: Yes, this can pretty much all be removed and you can create the Blob client using the connection string initializer.

Podfile
JoshuaLai: I updated the code to use init from connection string but at run time looks like the `AzureStorageBlob` internally has a dependency on it. 
```
dyld: Library not loaded: @rpath/MSAL.framework/MSAL
  Referenced from: .../Developer/Xcode/DerivedData/AzureSDK-dwmootmmaabdspgmqqrzmsanobci/Build/Products/Debug-iphonesimulator/AzureStorageBlob.framework/AzureStorageBlob
  Reason: image not found
  ```

Comment (tjprescott) [reply Podfile:71]: Should only be this: 
image

However, I'm okay leaving it in and I can take a look later, since we do want the Storage library to be usable without MSAL.

File: examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobListObservable.swift (hunk 11 above)
Comment (JoshuaLai) [reply examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobListObservable.swift:33]: Tried a couple of things but looks like the best way is to use a class. Struct won't  have as nice a design  to easily access the blob transfers progress. Lets talk about if there is a cleaner way to implement it

File: Podfile (hunk 1 above)
Comment (JoshuaLai) [reply Podfile:71]: I have removed the MSAL implementation from the rest of the code.   But for the time being I am leaving this in the podfile.

File: examples/AzureSDKDemoSwiftUI/AzureSDKDemoSwiftUI/Pages/BlobDownloadView.swift [hunk 21]
```diff
@@ -0,0 +1,140 @@
...
+
+import AzureCore
+import AzureIdentity
+import AzureStorageBlob
+
+import AVKit
+import Photos
+
+final class BlobDownloadTableViewController: UIViewControllerRepresentable {
+    var viewController: UITableViewController?
```
tjprescott, line 38: So what is the reason for keeping these UIViewControllerRepresentable instead of a typical SwiftUI View struct?

tjprescott: If it needs to be this, that's fine. We can consider a baseline for improvement. But if it isn't needed, this should look as close to "pure" SwiftUI as possible.

JoshuaLai: Yes we can discuss this but as I started working that I noticed that the video player further down would require a `UIViewRepresentable` and I would eventually go down this path. IMO this is the cleaner implementation

tjprescott (approved):

PR #709, opened 2021-02-11
[Chat] - IdentifierModels - Moving the models into chat

JoshuaLai:


sdk/communication/AzureCommunication/Tests/CommunicationIdentifierSerializerTests.swift
```diff
@@ -132,69 +189,25 @@ class CommunicationIdentifierSerializerTests: XCTestCase {
...
-                rawId: expectedId
-            ))
-
-        XCTAssertEqual(model.kind, .microsoftTeamsUser)
-        XCTAssertEqual(model.microsoftTeamsUserId, "user id")
-        XCTAssertEqual(model.isAnonymous, isAnonymous)
-        XCTAssertEqual(model.id, expectedId)
-    }
-
-    func test_DeserializeMissingProperty() throws {
```
RezaJooyandeh, line 156: why is this test not needed anymore?

sdk/communication/AzureCommunicationChat/Source/Generated/Models/MicrosoftTeamsUserIdentifierModel.swift
```diff
@@ -0,0 +1,67 @@
...
+
+import AzureCore
+import Foundation
+// swiftlint:disable superfluous_disable_command
+// swiftlint:disable identifier_name
+// swiftlint:disable line_length
+// swiftlint:disable cyclomatic_complexity
+
+/// A Microsoft Teams user.
+public struct MicrosoftTeamsUserIdentifierModel: Codable, Equatable {
```
RezaJooyandeh, line 19: let's make sure the model types are not public

JoshuaLai: The `public` was automatically added in the generated file. Should this be removed (we would need to manually do it though)

sdk/communication/AzureCommunication/Tests/CommunicationIdentifierSerializerTests.swift
JoshuaLai: Screen Shot 2021-02-11 at 5 42 46 PM
Since the new init has a bunch of default parameters technically these tests for the model become less relevant/ not possible.

RezaJooyandeh: but some of the combinations that we want to prohibit are valid in the schema. For example, in deserialization we want to enforce `rawId` to be present. In this example `rawId` is nullable because it is not needed for serialization.

sdk/communication/AzureCommunicationChat/Source/Generated/Models/MicrosoftTeamsUserIdentifierModel.swift
RezaJooyandeh: In some other languages there are ways to inform the code generator to generate as non public. I am wondering if autorest.swift already has mechanisms for it.

sdk/communication/AzureCommunication/Tests/CommunicationIdentifierSerializerTests.swift
JoshuaLai: Oh I see your point. Ok yes I will need to update these tests!

sdk/communication/AzureCommunicationChat/Source/Generated/Models/MicrosoftTeamsUserIdentifierModel.swift
JoshuaLai: Had a quick chat with @tjprescott and looks like we don't have this mechanism for swift autorest

sdk/communication/AzureCommunication/Tests/CommunicationIdentifierSerializerTests.swift
JoshuaLai: @RezaJooyandeh re-introduced the these tests in the `Chat/IdentifierSerializerTests`

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ## 1.0.0-beta.9 (Unreleased)
 ### New Features
 
+### New Features
+- Azure Communication Common Library
+ - Introduction of  a new struct `CommunicationIdentifierModel` to repesent a union type that is either a `communicationUser`, `phoneNumber`, or `microsoftTeamsUser`.  Along side this new object we introduced `CommunicationIdentifierSerializer` to serialize and deserialize valid JSON.
```
RezaJooyandeh, line 8: Are not we moving this type to the Chat package? I am also not sure if we should add this here. The type is only public because autorest.swift can't make it internal at this stage; otherwise this type is for internal usage only. @tjprescott , do you have suggestions on if we should include this in the release notes or not.

tjprescott: I would say remove it. Is the intent that both the common and the Chat SDKs will have this model internally or is Chat supposed to leverage the implementation in Common? If it's the second case, the the model must be public.

JoshuaLai: The model is going to live in the Chat package. Would it be better instead of having it under `Azure Communication Common Library` to update it to be under the `Azure Communication Chat Library`? Because this change log still does technically apply @RezaJooyandeh @tjprescott

sdk/communication/AzureCommunication/Source/CommunicationCloudEnvironment.swift
```diff
@@ -39,22 +39,15 @@ import Foundation
...
 
-    public static func == (lhs: CommunicationCloudEnvironment, rhs: CommunicationCloudEnvironment) -> Bool {
+    public static func isEqual(lhs: CommunicationCloudEnvironment, rhs: CommunicationCloudEnvironment) -> Bool {
         if lhs.environmentValue == rhs.environmentValue {
             return true
         } else {
             return false
         }
     }
 }
```
tjprescott, line 53: Why this change?  The `==` operator is the idiomatic implementation.

sdk/communication/AzureCommunicationChat/Source/Generated/Models/ChatMessageReadReceipt.swift
```diff
@@ -16,7 +16,7 @@ import Foundation
 // swiftlint:disable cyclomatic_complexity
 
 /// A chat message read receipt indicates the time a chat message was read by a recipient.
-public struct ChatMessageReadReceipt: Codable {
+public struct ChatMessageReadReceipt: Codable, Equatable {
```
tjprescott, line 19: Is this regenerated or manually added?

sdk/core/AzureCore/Source/Pipeline/Policies/ContentDecodePolicy.swift
```diff
@@ -224,7 +224,8 @@ public class ContentDecodePolicy: PipelineStage {
             contentType: contentType,
```
tjprescott, line 224: This should be reverted.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ## 1.0.0-beta.9 (Unreleased)
 ### New Features
 
+### New Features
```
tjprescott, line 6: Duplicate line. Remove one

tjprescott (changes requested):
Please revert any changes to AzureCore.

sdk/communication/AzureCommunicationChat/Source/Generated/Models/ChatMessageReadReceipt.swift
JoshuaLai: This was generated from autorest

CHANGELOG.md
tjprescott: Well, it should be in both--removed from AzureCommunication and added to Chat. It would be a breaking change for AzureCommunication, yes?

tjprescott (approved):

PR #716, opened 2021-02-16
[DO NOT MERGE] EngSys Updates

tjprescott:
Work in progress.


eng/pipelines/templates/jobs/archetype-sdk-release.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    variables:
+      - template: ../variables/globals.yml
+
+    pool:
+      vmImage: '$(OSVmImage)'
+
+    steps:
+      - script: |
+          echo "TODO: Tag Azure/azure-sdk-for-ios repo!"
```
weshaggard, line 12: We should use the common https://github.com/Azure/azure-sdk-for-ios/blob/master/eng/common/pipelines/templates/steps/create-tags-and-git-release.yml template but we might need to make some tweaks based to handle the ios convention.

Scripts/check_versions.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python3
```
weshaggard, line 1: We should put these under eng/scripts similar to everything else.

PR #718, opened 2021-02-17
[DO NOT MERGE] AzureCommunicationCalling API Review

raosanat:
1. API samples written in Swift demonstrates how SDK will be used from a Swift application.
2. Swift 5 generated header files for the Objective-C public header files exposed in the framework.

**Missing features:**

1. Currently we are not setting nullability type specifier for properties and arguments to API's , we will be fixing those in the APIGen before GA.
2. Some API's in the SDK that are throwing we need to add `__attribute__((swift_error(nonnull_error)))` , currently those API's will make the application crash and there is no way for the Swift code to catch those exception. We will fix those as well before GA.

**Changes done since the previous review:**

1. Passing parameter labels to the API's are mandatory.
2. ACS prefix for all the classes, enums are dropped when used from Swift application.
3. default init where not applicable has been disabled with `NS_SWIFT_UNAVAILABLE`.
4. For events first parameter label is disabled and second is mandatory
   ` public func onCallEnded(_ incomingCall: IncomingCall!, args: PropertyChangedEventArgs!) `

### DIFF
--- a/sdk/communication/AzureCommunicationCalling/ApiSamples.md
+++ b/sdk/communication/AzureCommunicationCalling/ApiSamples.md
@@ -1,0 +1,754 @@
+# Azure Communication Calling iOS API Samples
+
+<!-- Note: We may want to add a section for terminology explanation. Eg. what's CallClient vs CallAgent etc. -->
+
+## Overview
+- ### [Start with CallClient and CallAgent](#start-with-callclient-and-callagent)
+- ### [Call Basics](#call-basics)
+- ### [Video Basics](#video-basics)
+- ### [Device Management](#device-management)
+- ### [Push Notification](#push-notification)
+- ### [Remote Participant Management](#remote-participant-management)
+
+## Start with CallClient and CallAgent
+<!--Start Start with CallClient and CallAgent-->
+### Initialization
+Create a CallClient instance and call createCallAgent api to get CallAgent instance. Make sure you have imported "AzureCommunication.framework" and "AzureCore.framework"
+
+* Create CommunicationTokenCredential object so that SDK can fetch token
+```swift
+    import AzureCommunication
+
+    let tokenString = "token_string";
+    var userCredential: CommunicationTokenCredential?
+    do {
+        userCredential = try CommunicationTokenCredential(with: CommunicationTokenRefreshOptions(initialToken: token, 
+                                                                    refreshProactively: true,
+                                                                    tokenRefresher: self.fetchTokenSync))
+    } catch {
+        print("Failed to create CommunicationTokenCredential object")
+    }
+
+    self.tokenProvider = ContosoTokenProvider()
+    /*self.tokenProvider is Contoso's internal implementation that fetches token from Contoso server*/
+    public func fetchTokenSync(then onCompletion: TokenRefreshOnCompletion) {
+        let newToken = self.tokenProvider.getToken()
+        onCompletion(newToken, nil)
+    }
+```
+
+* Pass CommunicationUserCredential object created above to CallClient
+```swift
+public class CallingApp : NSObject, CallAgentDelegate, IncomingCallDelegate
+{
+    var callClient: CallClient?
+    var callAgent: CallAgent?
+    var callObserver: CallObserver?
+    var deviceManager: DeviceManager?
+
+    func init()
+    {
+        self.callClient = CallClient()
+        let options = CallAgentOptions()!
+        options.displayName = displayName
+        self.callClient!.createCallAgent(
+            userCredential: userCredential!,
+            options: options,
+            completionHandler: { (callAgent, error) in
+                if error != nil {
+                    print("Create agent succeeded")
+                    self.callAgent = callAgent
+                    self.callClient!.getDeviceManager(
+                        completionHandler: { (deviceManager, error) in
+                            if (error == nil) {
+                                print("Got device manager instance")
+                                self.deviceManager = deviceManager
+                            } else {
+                                print("Failed to get device manager instance")
+                            }
+                } else {
+                    print("Create agent failed")
+                }
+            })
+        self.callObserver = CallObserver()
+    }
+
+    // Event raised to get notified when call collection on CallAgent is added/removed
+    public func onCallsUpdated(_ callAgent: CallAgent!, args: CallsUpdatedEventArgs!) {
+        // Add application logic when call is added
+        args.addedCalls?.forEach { call in }
+
+        // Add application logic when call is removed
+        args.removedCalls?.forEach { call in }
+    }
+
+    // Event raised when there is an incoming call
+    public func onIncomingCall(_ callAgent: CallAgent!, incomingcall: IncomingCall!) {
+        self.incomingCall = incomingcall
+        // Subscribe to get OnCallEnded event
+        self.incomingCall!.delegate = self
+    }
+
+    // Event raised when incoming call was not answered
+    public func onCallEnded(_ incomingCall: IncomingCall!, args: PropertyChangedEventArgs!) {
+        self.incomingCall = nil
+    }
+}
+```
+<!--End Start with CallClient and CallAgent-->
+
+## Call Basics
+<!--Start Call Basics-->
+### Call Properties
+Call object has various properties
+```swift
+// [String] caller identity
+self.call.id
+
+// CallStateNone = 0,CallStateEarlyMedia = 1,CallStateIncoming = 2,CallStateConnecting = 3,CallStateRinging = 4,CallStateConnected = 5,CallStateHold = 6,CallStateDisconnecting = 7,CallStateDisconnected = 8
+self.call.state
+
+// CallDirectionOutgoing = 1, CallDirectionIncoming = 2
+self.call.callDirection
+
+// [Bool] isMicrophoneMuted - is local audio muted
+self.call.isMicrophoneMuted
+
+// [Errror] callEndReason - containing code/subcode/message indicating how call ended
+self.call.callEndReason
+
+// String[] localVideoStreams - collection of video streams send to other participants in a call
+self.call.localVideoStreams
+
+// RemoteParticipant[] remoteParticipants - collection of remote participants participating in this call
+self.call.remoteParticipants
+
+// Indicates if the call is currently being recorded or not
+self.call.isRecordingActive;
+```
+
+### Call Operations
+* ### Make Outgoing Call
+
+To create and start a call you need to call one of the APIs on CallAgent
+```swift
+public class CallManager : NSObject, CallDelegate
+{
+    public func placeCall(names: [CommunicationIdentifier])
+    {
+        let callOptions = StartCallOptions()!
+        // Using the CallAgent received from the factory method
+        // and returns a Call object
+        let audioOptions = AudioOptions()
+        audioOptions!.muted = true // Start the call in mute state
+        callOptions.audioOptions = audioOptions
+        let call = self.callAgent!.startCall(participants: names,
+                                            options: callOptions)
+        call!.delegate = self.callObserver
+    }
+
+    public func placeVideoCall(names: [CommunicationIdentifier])
+    {
+        let camera = self.deviceManager!.cameras!.first
+        let localVideoStream = LocalVideoStream(camera: camera)
+        let videoOptions = VideoOptions(localVideoStream: localVideoStream)
+        
+        let callOptions = StartCallOptions()
+        callOptions?.videoOptions = videoOptions
+
+        let call = self.callAgent?.startCall(participants: names,
+                                                options: callOptions)
+        call!.delegate = self.callObserver
+    }
+
+    // Following is one event which is part of CallDelegate
+    public func OnStateChanged(_ call: Call!,
+                                args: PropertyChangedEventArgs!)
+    {
+        let state = CallObserver.callStateToString(state: call.state)
+        print("Call state changed to %@", state)
+    }
+}
+
+```
+
+* ### Accept Incoming Call
+
+To accept a call you need to call one of the APIs on CallAgent
+```swift
+public class CallManager : NSObject
+{
+    public func acceptCall()
+    {
+        let acceptCallOptions = AcceptCallOptions()!
+        let camera = self.deviceManager!.cameras!.first
+        let localVideoStream = LocalVideoStream(camera: camera)
+        let videoOptions = VideoOptions(localVideoStream: localVideoStream)
+        acceptCallOptions.videoOptions = videoOptions
+        self.incomingCall!.accept(options: acceptCallOptions) { (call, error) in
+            if(error == nil) {
+                self.call = call
+                self.callObserver = CallObserver(view:self)
+                self.call!.delegate = self.callObserver
+            } else {
+                print("[IncomingCall] Accepting call failed")
+            }
+        }
+    }
+
+    public func rejectCall()
+    {
+        s
+    }
+}
+```
+
+* ### Join Call
+
+To join a call you need to call one of the APIs on CallClient
+```swift
+public class CallManager : NSObject
+{
+    public func joinCall(threadId: String)
+    {
+        let options = JoinCallOptions()!
+        self.localVideoStream = LocalVideoStream(camera: self.deviceManager!.cameras!.first)
+        options.videoOptions = VideoOptions(localVideoStream: self.localVideoStream!)!
+        let audioOptions = AudioOptions()!
+        audioOptions.muted = true // Join group call muted
+        options.audioOptions = audioOptions
+        if (threadId.starts(with: "http")) {
+            // Join a meeting with link
+            let teamsMeetingLinkLocator = TeamsMeetingLinkLocator(meetingLink: threadId)!;
+            self.call = self.callAgent?.join(with: teamsMeetingLinkLocator, joinCallOptions: options)
+        } else {
+            let groupCallLocator = GroupCallLocator(groupId: UUID(uuidString: threadId))!
+            self.call = self.callAgent?.join(with: groupCallLocator, joinCallOptions: options)
+        }
+        // Get updates on the call
+        self.call.delegate = self.callObserver
+    }
+}
+```
+
+* ### Mute/Unmute
+
+[Asynchronous] Local mute
+
+```swift
+self.call?.mute(completionHandler: { (error) in
+    if error == nil {
+        print("Successfully muted")
+    } else {
+        print("Failed to mute")
+    }
+})
+
+```
+[Asynchronous] Local unmute
+
+```swift
+call.unmute(completionHandler:{ (error) in
+    if error == nil {
+        print("Successfully unmuted")
+    } else {
+        print("Failed to unmute")
+    }
+})
+```
+
+* ### HangUp Call
+To hangUp the call:
+```swift
+var hangUpOptions = HangUpOptions()!
+hangUpOptions.forEveryOne = true
+self.call.hangUp(hangUpOptions) { (error) in
+    if (error == nil) {
+        print("HangUp successfull")
+    } else {
+        print("HangUp failed, try again")
+    }
+}
+print(CallObserver.callStateToString(state:call.state)) // => Disconnecting -> Disconnected
+```
+
+### Call Event Handling
+
+#### Event model
+Most of properties and collections can change it's value.
+To subscribe to these changes you can use following:
+* ##### Properties
+Application should implement the delegate to get notfied for change in properties
+```swift
+    self.call.delegate = CallObserver()
+    // Get the property of the call state by doing get on the call's state member
+    public func OnStateChanged(_ call: Call!,
+                                args: PropertyChangedEventArgs!)
+    {
+        print("Callback from SDK when the call state changes, current state: " + CallObserver.callStateToString(state:call.state))
+    }
+```
+
+* ##### Collections
+To subscribe to collection updated event also delegate needs to be implemented
+to get notfified when there is a change in collection
+```swift
+    self.call.delegate = self
+    // Collection contains the streams that were added or removed only
+    public func onLocalVideoStreamsChanged(_ call: Call!,
+                                           args: LocalVideoStreamsUpdatedEventArgs!)
+    {
+        print(args.addedStreams.count);
+        print(args.removedStreams.count);
+    }
+```
+
+To unsubscribe:
+```swift
+self.call.delegate = nil // Will not recieve anymore call updates
+self.callClient.delegate = nil // Will not get incoming call notification
+self.incomingCall.delegate = nil // Will not get when incoming call end notification
+```
+
+#### Examples
+Call event handler for call object created when incoming call is answered or when placing an outgoing call succeeds
+
+```swift
+public class CallObserver : NSObject, CallDelegate
+{
+    public static func callStateToString(state: CallState) -> String {
+        switch state {
+            case .connected: return "Connected"
+            case .connecting: return "Connecting"
+            case .disconnected: return "Disconnected"
+            case .disconnecting: return "Disconnecting"
+            case .earlyMedia: return "EarlyMedia"
+            case .localHold: return "LocalHold"
+            case .remoteHold: return "RemoteHold"
+            case .none: return "None"
+            case .ringing: return "Ringing"
+            case .inLobby: return "InLobby"
+            default: return "Unknown"
+        }
+    }
+
+    public func OnStateChanged(_ call: Call!,
+                                args: PropertyChangedEventArgs!)
+    {
+        print("Call state changed to: " + callStateToString(state:call.state));
+    }
+
+    public func onRemoteParticipantsUpdated(_ call: Call!,
+                                                args: ParticipantsUpdatedEventArgs!)
+    {
+        print(args.addedParticipants.count);
+        print(args.removedParticipants.count);
+    }
+
+    public func onVideoStreamsUpdated(_ remoteParticipant: RemoteParticipant!,
+                                        args: RemoteVideoStreamsEventArgs!)
+    {
+        print(args.addedStreams.count);
+        print(args.removedStreams.count);
+    }
+
+    public func onIsRecordingActiveChanged(_ call: Call!, args: PropertyChangedEventArgs!) {
+        print("Call recording state changed to " + call.isRecordingActive)
+    }
+
+    public func onLocalVideoStreamsChanged(_ call: Call!, args: LocalVideoStreamsUpdatedEventArgs!) {
+        print("Added streams count: " + args.addedStreams)
+        print("Removed streams count: " + args.addedStreams)
+    }
+
+    public func onIdChanged:(_ call: Call!, args: PropertyChangedEventArgs!) {
+        print("Call id has changed: oldValue: \(self.call.id), newValue: \(call.id)")
+    }
+}
+```
+
+* list existing ongoing calls that local user is part of
+```swift
+for call in self.callClient.calls { print(call); } // [Call, Call, Call...]
+```
+* subscribe to added/removed call
+```swift
+self.callObserver = CallObserver()
+
+public func onCallsUpdated(_ callAgent: CallAgent!,
+                           args: CallsUpdatedEventArgs!)
+{
+    for addedCall in args.addedCalls { print(addedCall); }
+    for removedCall in args.removedCalls { print(removedCall); }
+}
+
+```
+
+### Call Options
+* Different call options, you can pass additional options when call is created and started to control 
+which audio/video streams that will be used to start a call. (Devices have to obtained by enumerating devices using deviceManager)
+```swift
+var startCallOptions = StartCallOptions()
+var joinCallOptions = JoinCallOptions()
+var acceptCallOptions = AcceptCallOptions()
+
+var videoOptions = VideoOptions()
+var audioOptions = AudioOptions()
+
+// start with selected camera
+videoOptions.camera = deviceManager?.cameras?.first
+// start call muted
+audioOptions.muted = true
+```
+
+* State transition when call starts
+```swift
+// => None -> Connecting -> Ringing -> Connected
+print(CallObserver.callStateToString(state:self.call?.state))
+// every remote participant state will transition idependently: Idle -> Connecting -> Connected
+for remoteParticipant in self.call?.remoteParticipants { print(CallObserver.callStateToString(state: remoteParticipant.state)) }
+
+```
+<!--End Call Basics-->
+
+## Video Basics
+<!--Start Video Basics-->
+### Start / Stop Video
+* [Asynchronous] Start/stop video using the call object
+```swift
+let firstCamera: VideoDeviceInfo? = self.deviceManager?.cameras?.first
+let localVideoStream = LocalVideoStream(camera: firstCamera)
+let videoOptions = VideoOptions(localVideoStream: localVideoStream)
+
+self.call?.startVideo(stream: localVideoStream) { (error) in
+    if (error == nil) {
+        print("Local video started successfully")
+    } else {
+        print("Local video failed to start")
+    }
+}
+
+self.call?.stopVideo(stream: localVideoStream!) { (error) in
+    if (error == nil) {
+        print("Local video stopped successfully")
+    } else {
+        print("Local video failed to stop")
+    }
+}
+```
+
+### Local Video Preview
+You can use Renderer to start render stream from your local camera, this stream won't be send to other participants, it's local preview feed
+* [Synchronous] Start local video preview
+```swift
+let firstCamera: VideoDeviceInfo? = self.deviceManager?.cameras?.first
+let localVideoStream = LocalVideoStream(camera: firstCamera)
+let previewRenderer = try! Renderer(localVideoStream: localVideoStream)
+let previewView = try! previewRenderer!.createView(with: RenderingOptions(scalingMode:scalingMode))
+
+rendererObserver = RendererObserver(view: self)
+previewView!.delegate = rendererObserver
+```
+
+```swift
+public class RendererObserver: NSObject, RendererDelegate {
+    let view: View
+    init(view: View) {
+        self.view = view
+    }
+
+    public func onFirstFrameRendered() {
+        print("Size of frame is " + view.renderer!.size)
+    }
+
+    public func rendererFailedToStart() {
+        print("Rederer failed to start")
+    }
+}
+```
+
+Renderer has set of properties and methods that allows you to control it:
+```swift
+// Constructor can take in LocalVideoStream or RemoteVideoStream
+let localRenderer = Renderer(localVideoStream:localVideoStream)
+let remoteRenderer = Renderer(remoteVideoStream:remoteVideoStream)
+
+// [StreamSize] size of the rendering view
+renderer.size
+
+// [RendererDelegate] an object you provide to receive events from this Renderer instance
+renderer.delegate
+
+// [Synchronous] create view with rendering options
+renderer!.createView(with: RenderingOptions(scalingMode:ScalingMode.crop));
+// [Synchronous] dispose rendering view
+renderer.dispose();
+```
+
+### Switch Video Source
+* [Asynchronous] Switch local video, pass localVideoStream you got from call.startVideo() API call
+```swift
+// get camera
+var firstCamera = self.deviceManager?.cameras?.first
+
+localVideoStream?.switchSource(firstCamera, completionHandler: { (error) in
+    if (error == nil) {
+        print("Successfully switched to new source")
+    } else {
+        print("Failed to switch to new source")
+    }
+})
+```
+<!--End Video Basics-->
+
+## Device Management
+<!--Start Device Management-->
+### Microphone / Speaker
+Device mananager allows you to set device that will be used when starting a call
+If not  will fallback to OS defaults
+
+#### Get / Set Devices
+```swift
+// get microphone
+var firstMicrophone = self.deviceManager?.microphones?.first
+
+// get speaker
+var firstSpeaker = self.deviceManager?.speakers?.first
+
+// [Synchronous] set microphone
+self.deviceManager?.setMicrophone(microphoneDevice: firstMicrophone)
+
+// [Synchronous] set speaker
+self.deviceManager?.setSpeaker(speakerDevice: firstSpeaker)
+```
+
+AudioDeviceInfo has a set of properties:
+```swift
+var audioDeviceInfo = self.deviceManager?.microphones?.first
+
+// [NSString] name of the audio device
+audioDeviceInfo?.name
+
+// [NSString] id of the audio device
+audioDeviceInfo?.id
+
+// [BOOL] true if device is a system default
+audioDeviceInfo?.isSystemDefault
+
+// [AudioDeviceType] type of the audio device
+audioDeviceInfo?.deviceType
+```
+
+#### Enumerate Local Devices
+Enumeration is synchronous
+```swift
+// enumerate local cameras
+var localCameras = self.CallingApp.deviceManager!.cameras // [VideoDeviceInfo, VideoDeviceInfo...]
+// enumerate local cameras
+var localMicrophones = self.CallingApp.deviceManager!.microphones // [AudioDeviceInfo, AudioDeviceInfo...]
+// enumerate local cameras
+var localSpeakers = self.CallingApp.deviceManager!.speakers // [AudioDeviceInfo, AudioDeviceInfo...]
+
+```
+
+<!--End Device Management-->
+
+## Push Notification
+<!--Start Push Notification-->
+
+### Push Notification APIs
+
+#### Register for Push Notification
+
+- In order to register for push notification, call registerPushNotification() on a *CallAgent* instance with a device registration token.
+
+```swift
+let deviceToken: Data = pushRegistry?.pushToken(for: PKPushType.voIP)
+self.callAgent?.registerPushNotifications(deviceToken: deviceToken) { (error) in
+                    if(error == nil) {
+                        print("Successfully register to push notification.")
+                    } else {
+                        print("Failed to register push notification.")
+                    }
+                })
+```
+
+#### Push Notification Handling
+- In order to receive incoming calls push notifications, call *handlePushNotification()* on a *CallAgent* instance with a dictionary payload.
+
+```swift
+
+let callNotification = IncomingCallPushNotification.from(payload: pushPayload?.dictionaryPayload)
+
+self.callAgent?.handlePush(notification: callNotification) { (error) in
+            if (error != nil) {
+                print("Handling of push notification failed")
+            } else {
+                print("Handling of push notification was successful")
+            }
+        }
+```
+
+#### Unregister Push Notification
+
+- Applications can unregister push notification at any time. Simply call the `unregisterPushNotification()` method on *CallAgent*.
+
+```swift
+self.callAgent?.unRegisterPushNotifications(completionHandler: { (error) in
+                if (error != nil) {
+                    print("Unregister of push notification failed, please try again")
+                } else {
+                    print("Unregister of push notification was successfull")
+                }
+            })
+```
+<!--End Push Notification-->
+
+## Remote Participant Management
+<!--Start Remote Participant Management-->
+
+### Remote Participants Properties
+Remote participant has set of properties
+
+```swift
+// [RemoteParticipantDelegate] delegate - an object you provide to receive events from this RemoteParticipant instance
+var remoteParticipantDelegate = remoteParticipant?.delegate
+
+// [CommunicationIdentifier] identity - same as the one used to provision token for another user
+var identity = remoteParticipant?.identity;
+
+// ParticipantStateIdle = 0, ParticipantStateEarlyMedia = 1, ParticipantStateConnecting = 2, ParticipantStateConnected = 3, ParticipantStateOnHold = 4, ParticipantStateInLobby = 5, ParticipantStateDisconnected = 6
+var state = remoteParticipant?.state;
+
+// [Error] callEndReason - reason why participant left the call, contains code/subcode/message
+var callEndReason = remoteParticipant?.callEndReason
+
+// [Bool] isMuted - indicating if participant is muted
+var isMuted = remoteParticipant?.isMuted;
+
+// [Bool] isSpeaking - indicating if participant is currently speaking
+var isSpeaking = remoteParticipant?.isSpeaking;
+
+// RemoteVideoStream[] - collection of video streams this participants has
+var videoStreams = remoteParticipant?.videoStreams; // [RemoteVideoStream, RemoteVideoStream, ...]
+
+// RemoteVideoStream[] - collection of screen sharing streams this participants has
+var screenSharingStreams = remoteParticipant?.screenSharingStreams; // [RemoteVideoStream, RemoteVideoStream, ...]
+```
+
+### Call Operations for Remote Participant
+#### Add Participant to a Call
+* [Synchronous] Add participant to a call
+```swift
+let remoteParticipantAdded: RemoteParticipant = self.call.add(participant: CommunicationUserIdentifier(identifier: "8:echo123"))
+```
+
+#### Add Phone Number to a Call
+* [Synchronous] Add phone number to a call
+```swift
+let remoteParticipantAdded: RemoteParticipant = self.call.add(participant: PhoneNumberIdentifier(phoneNumber: "+14243333"))
+```
+
+#### Remove Participant from a Call
+* [Asynchronous] Remove participant from a call
+```swift
+self.call?.remove(participant: remoteParticipantAdded) { (error) in
+    if (error == nil) {
+        print("Successfully removed participant")
+    } else {
+        print("Failed to remove participant")
+    }
+}
+```
+
+#### List Participants in a Call
+List all participants in a call
+```swift
+for remoteParticipant in self.call?.remoteParticipants { print(remoteParticipant); } // [RemoteParticipant1, RemoteParticipant2...]
+```
+
+```swift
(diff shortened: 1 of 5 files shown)


sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+}
+
+
+/// Describes a PropertyChanged event data
+open class PropertyChangedEventArgs : NSObject {
+
+    
+    /// Deallocates the memory occupied by this object.
+    open func dealloc()
+}
```
tjprescott, line 841: Does this have no info on what property changed?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open var code: Int32 { get }
+
+    
+    /// The subcode
+    open var subcode: Int32 { get }
+}
+
+
+/// Video stream on remote participant (NOT SUPPORTED)
+open class RemoteVideoStream : NSObject {
```
tjprescott, line 815: If not supported, why is it in the SDK?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    // Class extension begins for RemoteParticipant.
+    open var identity: CommunicationIdentifier
+}
+
+// Class extension ends for RemoteParticipant.
+
+
+/// Describes the reason for a call to end
+open class CallEndReason : NSObject {
+
```
tjprescott, line 799: Why is this not an enum?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    /// True if the remote participant is muted
+    open var isMuted: Bool { get }
+
+    
+    /// True if the remote participant is speaking
+    open var isSpeaking: Bool { get }
+
+    
+    /// Reason why participant left the call, contains code/subcode.
+    open var callEndReason: CallEndReason! { get }
```
tjprescott, line 773: If I participant hasn't left the meeting, what does this look like? In Swift I might expect it to be nil, but this isn't possible here. Is there a special code for "not ended"?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+}
+
+// Class extension ends for CallAgent.
+
+
+/// Describes a call
+open class Call : NSObject {
+
+    
+    /// Deallocates the memory occupied by this object.
```
tjprescott, line 663: Okay, I take it back... **this** seems like the `CallClient`, not what you are currently calling the `CallClient` or the `CallAgent`. How does this sequence occur?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+    
+    /// Specify the display name of the local participant for all new calls
+    open var displayName: String!
+}
+
+
+/// Call agent created by the CallClient factory method createCallAgent It bears the responsibility of managing calls on behalf of the authenticated user
+open class CallAgent : NSObject {
+
```
tjprescott, line 626: It seems like this functionality could be merged into the `Call` class

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSStreamSize.h
```diff
@@ -0,0 +1,18 @@
...
+
+open class StreamSize : NSObject {
+
+    
+    public init(width: Int32, height: Int32)
+
+    open var width: Int32 { get }
+
+    open var height: Int32 { get }
+}
```
tjprescott, line 18: Ideally this would be exposed to Swift as a struct rather than a class, if possible

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
```diff
@@ -0,0 +1,21 @@
...
+
+open class RendererView : UIView {
+
+    
+    open func update(scalingMode: ScalingMode)
+
+    open func dispose()
+
+    open func isRendering() -> Bool
+}
```
tjprescott, line 21: Why is this necessary?  This will only work with UIKit. What is someone wants to use the Calling library with SwiftUI?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRenderer.h
```diff
@@ -0,0 +1,33 @@
...
+
+    open func createView(with options: RenderingOptions?) throws -> RendererView
+
+    open func dispose()
+
+    
+    open var size: StreamSize { get }
+
+    unowned(unsafe) open var delegate: RendererDelegate?
+}
```
tjprescott, line 33: This implementation is tied to a particular UI paradigm. What are you targeting? Programmatic UIs, Storyboards, XIB, SwiftUI? Have you explored how well this works with these different approaches?

raosanat: We haven't tested ourselves with Storyboards, we have been using SwiftUI as of now. But story board has been tested by our customers and there were no issues they reported.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSStreamSize.h
raosanat: Yes good point. I will take a note of this.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: Not really `CallClient` is the main entry point which for each token/user can create multiple `CallAgent`s.
`CallAgent` object can now be used to place outgoing calls, join group calls and when there is incoming calls are recieved. They are all `Call` objects which gets created and destroyed as and when there is a call going on.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: CallClient is the main entry point which for each token/user can create multiple CallAgents.
CallAgent object can now be used to place outgoing calls, join group calls and when there is incoming calls are recieved. They are all Call objects which gets created and destroyed as and when there is a call going on.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        options.displayName = displayName
+        self.callClient!.createCallAgent(
+            userCredential: userCredential!,
+            options: options,
+            completionHandler: { (callAgent, error) in
+                if error != nil {
+                    print("Create agent succeeded")
+                    self.callAgent = callAgent
+                    self.callClient?.getDeviceManager(
+                        completionHandler: { (deviceManager, error) in
```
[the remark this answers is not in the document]
raosanat: I will take a note of this, i saw this as well. Will fix this.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+
+    func init()
+    {
+        self.callClient = CallClient()
+        let options = CallAgentOptions()!
+        options.displayName = displayName
+        self.callClient!.createCallAgent(
+            userCredential: userCredential!,
+            options: options,
+            completionHandler: { (callAgent, error) in
```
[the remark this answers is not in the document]
raosanat: I will take a note of this, i saw this as well. Will fix this.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        userCredential = try CommunicationTokenCredential(with: CommunicationTokenRefreshOptions(initialToken: token, 
+                                                                    refreshProactively: true,
+                                                                    tokenRefresher: self.fetchTokenSync))
+    } catch {
+        print("Failed to create CommunicationTokenCredential object")
+    }
+
+    self.tokenProvider = ContosoTokenProvider()
+    /*self.tokenProvider is Contoso's internal implementation that fetches token from Contoso server*/
+    public func fetchTokenSync(then onCompletion: TokenRefreshOnCompletion) {
```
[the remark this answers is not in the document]
raosanat: This is part of the AzureCommunication package. Not related to AzureCommunicationCalling SDK.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: We already support this will update the sample. Thanks

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: Updating the sample code, Thanks.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: I havent added the entire application logic but application needs to make sure that userCredential should not be null when callAgent is to be created.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,758 @@
...
+                                           args: LocalVideoStreamsUpdatedEventArgs!)
+    {
+        print(args.addedStreams.count);
+        print(args.removedStreams.count);
+    }
+```
+
+To unsubscribe:
+```swift
+self.call?.delegate = nil // Will not recieve anymore call updates
```
JoshuaLai, line 312: This feels very odd. Its not common practice to expect the developer to set the delegate to nil to unsubscribe to something

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,758 @@
...
+* ### Accept Incoming Call
+
+To accept a call you need to call one of the APIs on CallAgent
+```swift
+public class CallManager : NSObject
+{
+    public func acceptCall()
+    {
+        let acceptCallOptions = AcceptCallOptions()!
+        let camera = self.deviceManager?.cameras!.first
```
JoshuaLai, line 182: Does this make sense for `cameras` to be nullable? I would expect that if the device manager didn't have any cameras to just return an empty array, or throw

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,758 @@
...
+// [Synchronous] set microphone
+self.deviceManager?.setMicrophone(microphoneDevice: firstMicrophone)
+
+// [Synchronous] set speaker
+self.deviceManager?.setSpeaker(speakerDevice: firstSpeaker)
+```
+
+AudioDeviceInfo has a set of properties:
+```swift
+var audioDeviceInfo = self.deviceManager?.microphones?.first
```
JoshuaLai, line 531: Same comment here does this make sense for microphones to be nullable? I would expect that if the device manager didn't have any to just return an empty array, or throw

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
raosanat: We have been using this SDK with iOS application written in SwiftUI from day 1. We are not aware of any issues that the calling SDK does not work with SwiftUI.

raosanat: @tjprescott can you clarify more on this? maybe i am missing something.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: Well this could be a valid scenario in some cases, say application does not want to receive any incoming calls. Then they can set the delegate to nil on CallAgent,  I understand that this is not a common thing application users do, we are just providing a way when you want to subscribe to an event, how to unsubscribe to an event, its completely up the application developer to use this or not.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: cameras will never be null or empty array but app developer should still take care of edge cases when say  there is an operating system bug and cameras are not listed for some weird reason. 
```
        guard let camera = deviceManager.getCameraList()!.first else {
            print("")
            return
        }

        localVideoStream = LocalVideoStream(camera: camera)
        completionHandler(localVideoStream)
```
I havent added gaurd statements all over the sample to keep the length of the sample to minimum and this code will not be published to the users.

tjprescott: I suspect this is a result of the ObjC translation where pretty much anything can be nil.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
tjprescott: I don't think subscribe/unsubscribe is the right terminology for this pattern. Something like `NotificationCenter` would be a publish/subscribe type pattern. I think setting the `delegate` to nil here does make sense as part of the delegate pattern.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
tjprescott: I wonder if this is the result of ObjC->Swift bridging, but there should be annotations to control this

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRenderer.h
tjprescott: That's great. SwiftUI was what I was most concerned about

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
tjprescott: What's the link to your SwiftUI sample?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
tjprescott: This has already been vetted through archboard for other languages, so it should be fine here too.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
tjprescott: This was discussed during the review. There are too many codes and subcodes to create an effective enum. But, then how does the developer make sense of the code and subcode? Is there something that translates these codes into useful info?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    
+    // Class extension begins for CallClient.
+    open func createCallAgent(userCredential: CommunicationTokenCredential!, options callAgentOptions: CallAgentOptions!, completionHandler: ((CallAgent?, Error?) -> Void)!)
+
+    
+    open func createCallAgent(userCredential: CommunicationTokenCredential!, completionHandler: ((CallAgent?, Error?) -> Void)!)
+
+    
+    open var communicationCredential: CommunicationTokenCredential
+}
```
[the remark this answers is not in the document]
tjprescott: This was discussed and vetted previously through the archboard.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+    
+    optional func onMicrophonesUpdated(_ deviceManager: DeviceManager!, args: AudioDevicesUpdatedEventArgs!)
+
+    optional func onSpeakersUpdated(_ deviceManager: DeviceManager!, args: AudioDevicesUpdatedEventArgs!)
+
+    optional func onCamerasUpdated(_ deviceManager: DeviceManager!, args: VideoDevicesUpdatedEventArgs!)
+}
+
+
```
tjprescott, line 350: None of these delegate method names align with the guidelines or Apples own delegate methods. To the max extent possible, these should be renamed through annotations.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
JoshuaLai: I am curious to see / try this out. I think because it is a UIView and not a ViewController SwiftUI should be able to work.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
JoshuaLai: Unfortunate. Should be preface them on the ObjC side with Non_Null?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
raosanat: This is the link to our internal test-app sample written in SwiftUI is [here](https://skype.visualstudio.com/SCC/_git/client_calling_spool-native?path=%2Ftest_app%2FiOS%2FACSCallingTestApp%2FACSCallingTestApp%2FViews%2FCallInProgressView.swift&version=GBsanathr%2Fapi_review_2&line=89&lineEnd=94&lineStartColumn=1&lineEndColumn=22&lineStyle=plain&_a=contents).

Using Storyboard sample is [here](https://github.com/Azure-Samples/communication-services-ios-calling-hero-pr).

raosanat: Marking this as Resolved for now, please re-open if you have any more questions. Thanks.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: We still don't have support for nullability type specifiers. ApiGen at the moment does not allow us to specify type specifier for property types, this was already mentioned that we are working on this. Its a big implementation in ApiGen so we havent got around  that yet, but will be fixed before GA.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: We are trying to be consistent across platforms. Is this considered as a blocker ? I will discuss with the team if we are ok having different names for different platforms.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: This will be taken care when we support adding specifying of nullability specifier this will be taken care.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+- ### [Call Basics](#call-basics)
+- ### [Video Basics](#video-basics)
+- ### [Device Management](#device-management)
+- ### [Push Notification](#push-notification)
+- ### [Remote Participant Management](#remote-participant-management)
+
+## Start with CallClient and CallAgent
+<!--Start Start with CallClient and CallAgent-->
+### Initialization
+Create a CallClient instance and call createCallAgent api to get CallAgent instance. Make sure you have imported "AzureCommunication.framework" and "AzureCore.framework"
```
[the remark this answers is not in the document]
raosanat: Will update the documentation

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: What would be the suggested usage in this case ?
1. User of the application wants to get incoming call, so delegate is set and incoming call events is recvd.
2. Now when user of the application does not want to get incoming call notified, what would be the suggested approach that app developer needs to take ? If the CallAgent object is deleted and created and then not set the delegate it will work. But thats not a good solution , setting the delegate to nil then application will not get anymore incoming call notification.

Are you suggesting we should implement NotificationCenter support because we do not want app developers to set delegate to nil ?

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: Let me correct this, i investigated on this and ApiGen support for setting nullable type specifier will fix this issue. As mentioned previously this is WIP and will be fixed before GA.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: Will be fixed as part of nullability type specifier in ApiGen

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+    public func onCallsUpdated(_ callAgent: CallAgent!, args: CallsUpdatedEventArgs!) {
+        // Add application logic when call is added
+        args.addedCalls?.forEach { call in }
+
+        // Add application logic when call is removed
+        args.removedCalls?.forEach { call in }
+    }
+
+    // Event raised when there is an incoming call
+    public func onIncomingCall(_ callAgent: CallAgent!, incomingcall: IncomingCall!) {
```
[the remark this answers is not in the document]
raosanat: ApiGen support for setting nullable type specifier will fix this issue. As mentioned previously this is WIP and will be fixed before GA.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        self.incomingCall = incomingcall
+        // Subscribe to get OnCallEnded event
+        self.incomingCall?.delegate = self
+    }
+
+    // Event raised when incoming call was not answered
+    public func onCallEnded(_ incomingCall: IncomingCall!, args: PropertyChangedEventArgs!) {
```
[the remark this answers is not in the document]
raosanat: ApiGen support for setting nullable type specifier will fix this issue. As mentioned previously this is WIP and will be fixed before GA.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+to get notfified when there is a change in collection
+```swift
+    self.call.delegate = self
+    // Collection contains the streams that were added or removed only
+    public func onLocalVideoStreamsChanged(_ call: Call!,
+                                           args: LocalVideoStreamsUpdatedEventArgs!)
+    {
+        print(args.addedStreams.count);
+        print(args.removedStreams.count);
+    }
```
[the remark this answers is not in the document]
raosanat: Fixed

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+            case .disconnected: return "Disconnected"
+            case .disconnecting: return "Disconnecting"
+            case .earlyMedia: return "EarlyMedia"
+            case .localHold: return "LocalHold"
+            case .remoteHold: return "RemoteHold"
+            case .none: return "None"
+            case .ringing: return "Ringing"
+            case .inLobby: return "InLobby"
+            default: return "Unknown"
+        }
```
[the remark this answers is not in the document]
raosanat: Yes, will mark this for future improvement. Assuming this is not GA blocker, closing this. Please re-open if you think so.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    /// Push notification device registration token is invalid.
+    public static var invalidPNDeviceRegistrationToken: CommunicationErrors { get }
+
+    /// Cannot create multiple renderers for same device or stream.
+    public static var multipleRenderersNotSupported: CommunicationErrors { get }
+
+    /// Renderer doesn't support creating multiple views.
+    public static var multipleViewsNotSupported: CommunicationErrors { get }
+}
```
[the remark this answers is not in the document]
raosanat: Yes, agreed. I will create WI for this, have to see how much ApiGen change is required here.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        self.callObserver = CallObserver()
+    }
+
+    // Event raised to get notified when call collection on CallAgent is added/removed
+    public func onCallsUpdated(_ callAgent: CallAgent!, args: CallsUpdatedEventArgs!) {
+        // Add application logic when call is added
+        args.addedCalls?.forEach { call in }
+
+        // Add application logic when call is removed
+        args.removedCalls?.forEach { call in }
```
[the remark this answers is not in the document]
raosanat: as mentioned in the comment.
// Add application logic when call is removed
application needs to add the logic, if any UI needs to be updated.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+
+// String[] localVideoStreams - collection of video streams send to other participants in a call
+self.call?.localVideoStreams
+
+// RemoteParticipant[] remoteParticipants - collection of remote participants participating in this call
+self.call?.remoteParticipants
+
+// Indicates if the call is currently being recorded or not
+self.call?.isRecordingActive;
+```
```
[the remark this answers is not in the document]
raosanat: Yes we do already document the properties in reference documentation here. https://docs.microsoft.com/en-us/objectivec/communication-services/calling/acscall

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+
+/**
+ * A set of methods that are called by ACSInternalTokenProvider in response to important events.
+ */
+public protocol InternalTokenProviderDelegate : NSObjectProtocol {
+
+    
+    optional func onTokenRequested(_ internalTokenProvider: InternalTokenProvider!, sender: InternalTokenProvider!)
+}
```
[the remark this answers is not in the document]
raosanat: Yes we will fix this before GA. Currently ApiGen has a limitation that model file cannot have an internal class. But we should be able to work around his problem.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+/**
+ * A set of methods that are called by ACSCallAgent in response to important events.
+ */
+public protocol CallAgentDelegate : NSObjectProtocol {
+
+    
+    optional func onCallsUpdated(_ callAgent: CallAgent!, args: CallsUpdatedEventArgs!)
+
+    optional func onIncomingCall(_ callAgent: CallAgent!, incomingcall: IncomingCall!)
+}
```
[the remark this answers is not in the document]
raosanat: This would be a bit difficult to achieve in ApiGen, without major changes. But will investigate more.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    /// Sets to True when the local video stream is being sent on a call.
+    open var isSending: Bool { get }
+
+    
+    /// Video stream type being used for the current stream.
+    open var mediaStreamType: MediaStreamType { get }
+
+    
+    // Class extension begins for LocalVideoStream.
+    open func switchSource(camera: VideoDeviceInfo!, completionHandler: ((Error?) -> Void)!)
```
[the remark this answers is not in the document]
raosanat: this API is only meant for switching an active video stream with a different source. So the argument passed should be non-null. For stopping the stream its a different process , which is documented later.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+    
+    /// Get the Device Type of this video device.
+    open var deviceType: VideoDeviceType { get }
+}
+
+
+/// Internal Use Only. Should not be used publicly. Will be removed in the future.
+open class InternalTokenProvider : NSObject {
+
```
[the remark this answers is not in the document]
raosanat: This is part of the model file and at the moment ApiGen does not support private or internal class in model files.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open func dealloc()
+
+    
+    /// Get the name of this video device.
+    open var name: String! { get }
+
+    
+    /// Get Name of this audio device.
+    open var id: String! { get }
+
```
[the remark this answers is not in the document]
raosanat: yes , this is the correct one 

```
    /// Get the name of this audio device.
    open var name: String! { get }

    
    /// Get Id of this audio device.
    open var id: String! { get }
```

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    optional func onMicrophonesUpdated(_ deviceManager: DeviceManager!, args: AudioDevicesUpdatedEventArgs!)
+
+    optional func onSpeakersUpdated(_ deviceManager: DeviceManager!, args: AudioDevicesUpdatedEventArgs!)
+
+    optional func onCamerasUpdated(_ deviceManager: DeviceManager!, args: VideoDevicesUpdatedEventArgs!)
+}
+
+
+/// Property bag class for Video Options. Use this class to set video options required during a call (start/accept/join)
+open class VideoOptions : NSObject {
```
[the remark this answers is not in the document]
raosanat: I guess we mean the same things here , let me bring this up with the Team.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open func set(with token: String!, accountIdentity: String!, displayName: String!, resourceId: String!)
+
+    
+    /// Exclusively for Internal. Do not use publicly. Will be removed in the future.
+    open func set(error: String!)
+}
+
+
+/// Property bag class for Audio Options. Use this class to set audio settings required during a call (start/join)
+open class AudioOptions : NSObject {
```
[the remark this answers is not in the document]
raosanat: Resolving this, get the idea in general.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    /// Video options when starting a call
+    open var videoOptions: VideoOptions!
+
+    
+    /// Audio options when starting a call
+    open var audioOptions: AudioOptions!
+
+    
+    // Class extension begins for StartCallOptions.
+    open var alternateCallerID: PhoneNumberIdentifier
```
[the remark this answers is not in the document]
raosanat: Yes, makes sense

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open func dealloc()
+
+    
+    /// Video options when accepting a call
+    open var videoOptions: VideoOptions!
+}
+
+
+/// Options to be passed when starting a call
+open class StartCallOptions : NSObject {
```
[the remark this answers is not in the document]
raosanat: Resolving this, already have noted this down.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    
+    /// Deallocates the memory occupied by this object.
+    open func dealloc()
+}
+
+
+/// Options for joining a group call
+open class GroupCallLocator : AbstractJoinMeetingLocator {
+
+    public init!(groupId: UUID!)
```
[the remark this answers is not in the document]
raosanat: This needs to come to the application via link or a push notification message. Scenario is you are invited to a meeting and you are provided the meeting link , application is supposed to extract the guid and invoke the API.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+     */
+    unowned(unsafe) open var delegate: DeviceManagerDelegate!
+
+    
+    /// Set the microphone to be used for all active calls
+    open func setMicrophone(microphoneDevice: AudioDeviceInfo!)
+
+    
+    /// Set the speakers to be used for all active calls
+    open func setSpeaker(speakerDevice: AudioDeviceInfo!)
```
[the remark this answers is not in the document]
raosanat: Makes sense but probably would be a bit complicated to do this in ApiGen. Noting this down for now. But this isnt a blocker rite ?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open func accept(options: AcceptCallOptions!, completionHandler: ((Call?, Error?) -> Void)!)
+
+    
+    /// Reject this incoming call
+    open func reject(completionHandler: ((Error?) -> Void)!)
+}
+
+
+/// Property bag class as container for SDK initialization options.
+open class InitializationOptions : NSObject {
```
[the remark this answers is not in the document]
raosanat: This is also internal similar to `InternalTokenProvider`. Will fix this before GA.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+    open var addedCalls: [Call]! { get }
+
+    
+    /// Calls that are no longer tracked by the library
+    open var removedCalls: [Call]! { get }
+}
+
+
+/// Describes the reason for a call to end
+open class IncomingCall : NSObject {
```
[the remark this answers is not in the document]
raosanat: Yes, will fix that.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+     */
+    unowned(unsafe) open var delegate: IncomingCallDelegate!
+
+    
+    /// Accept an incoming call
+    open func accept(options: AcceptCallOptions!, completionHandler: ((Call?, Error?) -> Void)!)
+
+    
+    /// Reject this incoming call
+    open func reject(completionHandler: ((Error?) -> Void)!)
```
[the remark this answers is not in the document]
raosanat: No , `accept` and `reject` are operations you can call on a `Call` object previously. Now we have moved it to `IncomingCall` object. `CallAgent` class is an object which is obtained when you login with a token and all other call operations are done on the `CallAgent` object.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: Maybe in the wrong place, can you clarify ?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: As discussed at the moment this is how it is for other platforms also its supposed to be. Closing this as of now as it wont be possible, when we fix other platforms in the future we will fix this one also.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: This wont be nil anytime, success of failure will have non-null call end reason. With nullability type specifier support in the future we will fix the need to force un-wrapping.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        call?.delegate = self.callObserver
+    }
+
+    public func placeVideoCall(names: [CommunicationIdentifier])
+    {
+        let camera = self.deviceManager?.cameras!.first
+        let localVideoStream = LocalVideoStream(camera: camera)
+        let videoOptions = VideoOptions(localVideoStream: localVideoStream)
+        
+        let callOptions = StartCallOptions()
```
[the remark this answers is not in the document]
raosanat: i did not understand the question. Can you explain a bit more ?

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+        call?.delegate = self.callObserver
+    }
+
+    // Following is one event which is part of CallDelegate
+    public func OnStateChanged(_ call: Call!,
+                                args: PropertyChangedEventArgs!)
+    {
+        let state = CallObserver.callStateToString(state: call.state)
+        print("Call state changed to %@", state)
+    }
```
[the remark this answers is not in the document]
raosanat: Will check how we can implement this, will have to investigate. 

`OnStateChanged`  to `call` may not be possible really as it might break on other platforms and changing the logic only for iOS would be a bit complicated in ApiGen.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
```diff
@@ -0,0 +1,760 @@
...
+{
+    public func acceptCall()
+    {
+        let acceptCallOptions = AcceptCallOptions()!
+        let camera = self.deviceManager?.cameras!.first
+        let localVideoStream = LocalVideoStream(camera: camera)
+        let videoOptions = VideoOptions(localVideoStream: localVideoStream)
+        acceptCallOptions.videoOptions = videoOptions
+        self.incomingCall?.accept(options: acceptCallOptions) { (call, error) in
+            if(error == nil) {
```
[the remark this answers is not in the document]
raosanat: This is asynchronous call, need more info here.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+    
+    // Class extension begins for IncomingCallInformation.
+    open var from: CommunicationIdentifier! { get }
+
+    open var to: CommunicationIdentifier! { get }
+
+    open var callId: UUID { get }
+
+    open class func from(payload: [AnyHashable : Any]!) -> IncomingCallInformation!
```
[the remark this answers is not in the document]
raosanat: So the scenario is , 

1. App is in killed state.
2. App receives incoming push.
3. Static function will be called on IncomingCallInformation to parse payload information
4. payload info will be passed to CallKit  to notify incoming call info, if not done then app will be Killed by the iOS
5. App will initialize SDK and call handlePush API on the SDK in the background thread.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/ACSRendererView.h
raosanat: Swift UI sample in our repo is (here)[https://skype.visualstudio.com/SCC/_git/client_calling_spool-native?path=%2Ftest_app%2FiOS%2FACSCallingTestApp%2FACSCallingTestApp%2FViews%2FCallInProgressView.swift&version=GBmaster&line=729&lineEnd=730&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents]

Storyboard sample is (here)[https://github.com/Azure-Samples/communication-services-ios-calling-hero-pr/blob/main/ACSCall/Views/ParticipantView.swift]

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
tjprescott: Here's what the SDK Guidelines (general, not just iOS) say regarding consistency:
https://azure.github.io/azure-sdk/general_introduction.html#consistent
```
Client libraries should be consistent within the language, consistent with the service and consistent between all target languages. In cases of conflict, consistency within the language is the highest priority and consistency between all target languages is the lowest priority.
```

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
tjprescott: How is this resolved?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
tjprescott: Lines 836-841... has no info on what property changed. It seems to have no properties at all.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
tjprescott: Nothing for a preview SDK would be blocking

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: Reosolving this as its related to setting of nullability type specifiers.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: BLOCKER

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
raosanat: @tjprescott example ?

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
raosanat: Not a blcoker

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
```diff
@@ -0,0 +1,1124 @@
...
+
+/// Local and Remote Video scaling mode
+public enum ScalingMode : Int {
+
+    
+    /// Cropped
+    case crop = 1
+
+    /// Fitted
+    case fit = 2
```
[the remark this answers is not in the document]
raosanat: typedef NS_ENUM(NSInteger, ACSMediaStreamType)
{
    /// Video
    ACSMediaStreamTypeVideo = 0,
    /// Screen share
    ACSMediaStreamTypeScreenSharing = 1
} NS_SWIFT_NAME(MediaStreamType);



typedef NS_ENUM(NSInteger, ACSAudioDeviceType)
{
    /// Audio device is a microphone
    ACSAudioDeviceTypeMicrophone = 0,
    /// Audio device is a speaker
    ACSAudioDeviceTypeSpeaker = 1
} NS_SWIFT_NAME(AudioDeviceType);

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
raosanat: BLOCKER.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
raosanat: Already marked as a blocker.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
raosanat: already marked a a blocker

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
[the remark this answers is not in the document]
raosanat: Already noted down, not a blocker will discuss with the team.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
raosanat: Will sync up with @JoshuaLai offline.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: already marked as GA Blocker, closing this.

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: will fix the docstring

sdk/communication/AzureCommunicationCalling/Swift_Generated_Interfaces/AzureCommunicationCalling.h
raosanat: Add a string field here.

sdk/communication/AzureCommunicationCalling/ApiSamples.md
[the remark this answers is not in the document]
tjprescott:
```swift
 self.incomingCall?.accept(options: acceptCallOptions) { call, error in
  guard error == nil else {
    print("error")
    return
  }
  // otherwise it was successful
  ...
}
```

This would be the more idiomatic swift approach rather than if/else

PR #723, opened 2021-02-18
Update SmokeTest for use on CI

### LINKED ISSUE #695: SmokeTest does not work correctly in CI
The smoketest project is set up as a CI check. However, the podfile for the project only installs things from the master branch, meaning it will not pre-emptively catch a problem with a PR that would break the master branch. The problem is only caught once the PR is merged.

The smoketest Podfile should be parameterized to accept a branch name, which should be configured via CI to be the branch of the PR.

tjprescott:
Configures the SmokeTest to pull podspecs from the build branch. This should mean that changes which would break the build can be stopped before they are introduced rather than caught after the build has been broken.

Fixes #695.

Note that I'm note entirely sure this *should* run on PRs. My original intent was only to run it for releases. I can envision a scenario where, for example, `AzureCommunication` makes a breaking change and this check will fail until `AzureCommunicationCalling` updates their SDK accordingly. Thoughts?

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -247,49 +247,50 @@ jobs:
                      -configuration Debug | xcpretty -c
         displayName: 'Build examples [iphonesimulator]'
 
-#  - job: 'SmokeTest'
-#    strategy:
-#      matrix:
-#        # Build SDK libraries as frameworks, dynamically link to copies stored within the app bundle
-#        frameworks:
-#          PodsLinkage: frameworks
-#
-#        # Build SDK libraries as static libs (.a), statically link them into the app binary
-#        # This is currently broken, probably because Calling is pre-built as a framework
-#        # static:
-#        #   PodsLinkage: static
-#
-#        # Build SDK libraries as frameworks, but then statically link them into the app binary
-#        # Calling is still dynamically linked, probably because Calling is pre-built as a framework
-#        staticframeworks:
-#          PodsLinkage: staticframeworks
-#
-#    variables:
-#      - template: ../variables/globals.yml
-#
-#    pool:
-#      vmImage: '$(OSVmImage)'
-#
-#    steps:
-#      - script: sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app
-#        displayName: 'Use Xcode $(XcodeVersion)'
-#
-#      - script: |
-#          cd examples/AzureSDKSmokeTest
-#          pod install
-#        env:
-#          PODS_LINKAGE: $(PodsLinkage)
-#        displayName: 'Install dependencies'
-#
-#      - script: |
-#          set -o pipefail
-#          cd examples/AzureSDKSmokeTest
-#          xcodebuild build \
-#                     -workspace AzureSDKSmokeTest.xcworkspace \
-#                     -scheme AzureSDKSmokeTest \
-#                     -sdk iphonesimulator \
-#                     -configuration Debug | xcpretty -c
-#        displayName: 'Build smoke test [iphonesimulator]'
+  - job: 'SmokeTest'
+    strategy:
+      matrix:
+        # Build SDK libraries as frameworks, dynamically link to copies stored within the app bundle
+        frameworks:
+          PodsLinkage: frameworks
+
+        # Build SDK libraries as static libs (.a), statically link them into the app binary
+        # This is currently broken, probably because Calling is pre-built as a framework
+        # static:
+        #   PodsLinkage: static
+
+        # Build SDK libraries as frameworks, but then statically link them into the app binary
+        # Calling is still dynamically linked, probably because Calling is pre-built as a framework
+        staticframeworks:
+          PodsLinkage: staticframeworks
+
+    variables:
+      - template: ../variables/globals.yml
+
+    pool:
+      vmImage: '$(OSVmImage)'
+
+    steps:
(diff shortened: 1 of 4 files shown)


tjprescott:
@mitchdenny I'm not sure how to correctly wire up the repo and branch within Azure Pipelines.

image

I'm not sure why the repo is listed as ***, because in the queue time variables it shows up:
image

However, I don't actually think this is correct because the source repo is my fork, NOT the main repo. 

Any advice?

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -247,49 +247,51 @@ jobs:
...
+      - script: sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app
+        displayName: 'Use Xcode $(XcodeVersion)'
+
+      - script: |
+          cd examples/AzureSDKSmokeTest
+          pod install
+        env:
+          PODS_LINKAGE: $(PodsLinkage)
+          PODFILE_GIT_BRANCH: $(system.pullRequest.sourceBranch)
+          PODFILE_GIT_REPO: $(system.pullRequest.sourceRepositoryUri)
```
tjprescott, line 283: @mitchdenny this is where I'm currently stuck. I also tried `Build.SourceBranch` and `Build.SourceBranchName` but those were even more off than what I currently have.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -247,49 +247,50 @@ jobs:
...
+      vmImage: '$(OSVmImage)'
+
+    steps:
+      - script: sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app
+        displayName: 'Use Xcode $(XcodeVersion)'
+
+      - script: |
+          cd examples/AzureSDKSmokeTest
+          export PODFILE_GIT_REPO=$(curl -s https://api.github.com/repos/Azure/azure-sdk-for-ios/pulls/$(system.pullRequest.pullRequestNumber) | jq -r '.head.repo.full_name')
+          export PODFILE_GIT_BRANCH=$(curl -s https://api.github.com/repos/Azure/azure-sdk-for-ios/pulls/$(system.pullRequest.pullRequestNumber) | jq -r '.head.ref')
```
tjprescott, line 280: I don't love the fact that this calls a public API twice when I feel once would suffice, but bash is such a nightmare that I'm leaving it unless @mitchdenny or @bsiegel can suggest better syntax.

Also, any suggestions for how to avoid calling the Github API would be very welcome.

mitchdenny: The only thing I was thinking was that if one of the pre-defined variables in the pipeline already had this detail, but I think that they'll return the details for Azure/azure-sdk-for-ios rather than the fork that you are submitting the PR from.

mitchdenny:
Why is it returning null sometimes? That is surprising.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
tjprescott: I was able to omit calling the github API by detecting the CI condition and have the Podfile point to the local podspec files that are cloned from the branch/PR.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -247,49 +247,50 @@ jobs:
...
+    steps:
+      - script: sudo xcode-select --switch /Applications/Xcode_$(XcodeVersion).app
+        displayName: 'Use Xcode $(XcodeVersion)'
+
+      - script: |
+          cd examples/AzureSDKSmokeTest
+          pod install
+        env:
+          IS_CI: True
+          PODSPEC_DIR: $(Build.Repository.LocalPath)
```
tjprescott, line 282: @mitchdenny if there's a standard way of detecting CI, I would gladly replace this.

examples/AzureSDKSmokeTest/Podfile
```diff
@@ -37,9 +38,22 @@ else
...
+    pod 'AzureCommunication', :podspec => "#{local_path}/AzureCommunication.podspec.json"
+    pod 'AzureCommunicationChat', :podspec => "#{local_path}/AzureCommunicationChat.podspec.json"
+    pod 'AzureCommunicationCalling', :podspec => "#{local_path}/AzureCommunicationCalling.podspec.json"
+  else
+    # manual run targets AzureSDK master branch
+    pod 'AzureCore', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+    pod 'AzureCommunication', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+    pod 'AzureCommunicationChat', :git => 'https://github.com/Azure/azure-sdk-for-ios.git', :branch => 'master'
+    pod 'AzureCommunicationCalling', :podspec => 'https://raw.githubusercontent.com/Azure/azure-sdk-for-ios/master/AzureCommunicationCalling.podspec.json'
+  end
```
tjprescott, line 58: I could reinstate the env vars to target a specfic repo/branch, but this is less important for a manual test. The key scenario is testing the CI.

tjprescott:
> Why is it returning null sometimes? That is surprising.

@mitchdenny I have no idea, but calling out to an API felt dirty anyhow. By targeting the podspecs that get cloned when the pipeline starts, I accomplish my desired goal without calling the Github API directly. 

Please take a second look.

mitchdenny (approved):

PR #733, opened 2021-02-24
Chat preview4 - Communication identifier changes

knvsl:
- Includes Communications identifier changes
- Regenerated Chat using preview4 swagger
- Changed type on models returned by Chat convenience layer from CommunicationUserIdentifier to CommunicationIdentifier
- Using identifier serializer to convert between CommunicationIdentifierModels and CommunicationIdentifiers
- Updated tests + README


JoshuaLai (changes requested):

sdk/communication/AzureCommunication/Source/CommunicationCloudEnvironment.swift
```diff
@@ -29,7 +29,7 @@
 import Foundation
 
 @objcMembers public class CommunicationCloudEnvironment: NSObject {
-    private(set) var environmentValue: String
```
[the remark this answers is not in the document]
knvsl: looks like this change came from #709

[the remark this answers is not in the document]
JoshuaLai: 😱 I made this change. Odd I tried removing it and compiler complained

JoshuaLai (approved):

tjprescott (changes requested):
Preventing from being merged as this is not intended for March release.

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
```diff
@@ -37,8 +38,8 @@ public struct MessageContent: Codable {
     public let topic: String?
     /// Chat message content for messages of types participantAdded or participantRemoved.
     public let participants: [Participant]?
-    /// Chat message content for messages of types participantAdded or participantRemoved.
-    public let initiator: String?
+    /// Communication Identifier of chat message initiator
```
RezaJooyandeh, line 41: The initiator of the message.

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
```diff
@@ -47,25 +48,36 @@ public struct MessageContent: Codable {
...
+        }
     }
 
     /// Initialize a `MessageContent` structure.
     /// - Parameters:
     ///   - message: Message content for messages of types text or html.
     ///   - topic: Message content for messages of type topicUpdated.
     ///   - participants: Message content for messages of types participantAdded or participantRemoved.
-    ///   - initiator: Message content for messages of types participantAdded or participantRemoved.
+    ///   - initiator: The initiator of the message..
```
RezaJooyandeh, line 75: There is an extra "." in the end

sdk/communication/AzureCommunicationChat/Source/Models/Participant.swift
```diff
@@ -32,8 +32,8 @@ import Foundation
 public struct Participant: Codable {
     // MARK: Properties
 
-    /// The CommunicationUserIdentifier for the participant.
-    public let user: CommunicationUserIdentifier
+    /// The CommunicationIdentifier for the participant.
```
RezaJooyandeh, line 35: The identifier of the participant.

sdk/communication/AzureCommunicationChat/Swagger/swagger.json
```diff
@@ -0,0 +1,1711 @@
+{
```
RezaJooyandeh, line 1: We do not need this file here any more, right? I assume we are referencing the file in the spec repo.

RezaJooyandeh (approved):
Lgtm, I only left a few small comments.

JoshuaLai:
@tjprescott We updated with Camilo and we are going to merge preview 4 into `master` after all. We want to release this Chat and Communication together.

sdk/communication/AzureCommunicationChat/Swagger/swagger.json
knvsl: Correct - removed, thanks

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.0.0-beta.9 (Unreleased)
 ### New Features
+**Azure Communication Calling**
+ - Introduction of  a new struct `CommunicationIdentifierModel` to repesent a union type that is either a `communicationUser`, `phoneNumber`, or `microsoftTeamsUser`.  Along side this new object we introduced `CommunicationIdentifierSerializer` to serialize and deserialize valid JSON.
```
tjprescott, line 6: The second sentence isn't necessary, particularly if you don't envision customers using the serializer directly.

sdk/communication/AzureCommunication/Source/CommunicationCloudEnvironment.swift
```diff
@@ -39,8 +39,8 @@ import Foundation
         self.environmentValue = environmentValue
     }
 
-    static func fromModel(environmentModel: CommunicationCloudEnvironmentModel) -> CommunicationCloudEnvironment {
-        return CommunicationCloudEnvironment(environmentValue: environmentModel.requestString)
+    public func getEnvironmentValue() -> String {
+        return environmentValue
     }
```
tjprescott, line 44: It looks like the removal of `fromModel(...)` is a breaking change and should be annotated in the changelog.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
```diff
@@ -1,90 +0,0 @@
...
-import Foundation
-// swiftlint:disable superfluous_disable_command
-// swiftlint:disable identifier_name
-// swiftlint:disable line_length
-// swiftlint:disable cyclomatic_complexity
-
-/// Identifies a participant in Azure Communication services. A participant is, for example, a phone number or an Azure communication user.
-public struct CommunicationIdentifierModel: Codable {
-    // MARK: Properties
-
```
tjprescott, line 21: Where these models previously released in `AzureCommunication`? If so, we must call out the breaking change in the changelogs.

sdk/communication/AzureCommunicationChat/README.md
```diff
@@ -113,7 +113,7 @@ Using the APIs, users can also send typing indicators when typing a message and
 
 #### Initialization
 
-To instantiate ChatClient you will need the CommunicationServices resource endpoint, a CommunicationUserCredential created from a User Access Token, and optional options to create the client with.
+To instantiate a ChatClient you will need the CommunicationServices resource endpoint, a CommunicationTokenCredential created from a User Access Token, and optional options to create the client with.
```
tjprescott, line 116: Is this rename a breaking change that should be called out, or was the readme just wrong?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -91,22 +91,23 @@ public class ChatClient {
         thread: CreateThreadRequest,
         withOptions options: Chat.CreateChatThreadOptions? = nil,
         completionHandler: @escaping HTTPResultHandler<CreateThreadResult>
-    ) {
+    ) throws {
         // Set the repeatabilityRequestID if it is not provided
-        let requestOptions = ((options?.repeatabilityRequestID) != nil) ? options : Chat.CreateChatThreadOptions(
-            repeatabilityRequestID: UUID().uuidString,
+        let requestOptions = ((options?.repeatabilityRequestId) != nil) ? options : Chat.CreateChatThreadOptions(
```
tjprescott, line 96: The name change of `repeatabilityRequestID` to `repeatabilityRequestId` is a breaking change if it was previously released.

sdk/communication/AzureCommunicationChat/Source/Models/CreateThreadResult.swift
```diff
@@ -40,10 +40,12 @@ public struct CreateThreadResult: Codable {
 
     public init(
         from createChatThreadResult: CreateChatThreadResult
-    ) {
-        self
-            .thread = (createChatThreadResult.chatThread != nil) ? Thread(from: createChatThreadResult.chatThread!) :
-            nil
+    ) throws {
+        if let chatThread = createChatThreadResult.chatThread {
+            self.thread = try? Thread(from: chatThread)
```
tjprescott, line 45: Why is this marked `throws`?

sdk/communication/AzureCommunicationChat/Source/Models/Message.swift
```diff
@@ -60,17 +60,29 @@ public struct Message: Codable {
...
-        self
-            .sender = (chatMessage.senderId != nil) ? CommunicationUserIdentifier(identifier: chatMessage.senderId!) :
-            nil
+
+        // Deserialize the identifier model to CommunicationIdentifier
+        if let identifierModel = chatMessage.senderCommunicationIdentifier {
+            self.sender = try IdentifierSerializer.deserialize(identifier: identifierModel)
+        } else {
+            self.sender = nil
+        }
```
tjprescott, line 84: Why make this throw? If content and sender can both be `nil`, why not use `try?`, which would allow you to not throw?

sdk/communication/AzureCommunicationChat/Source/Models/Message.swift
```diff
@@ -136,15 +148,24 @@ public struct Message: Codable {
...
         // Convert ChatMessageContent to MessageContent
-        let content = try? container.decode(ChatMessageContent.self, forKey: .content)
-        self.content = MessageContent(from: content!)
+        let chatMessageContent = try? container.decode(ChatMessageContent.self, forKey: .content)
+
+        if let content = chatMessageContent {
+            self.content = try? MessageContent(from: content)
+        } else {
+            self.content = nil
+        }
```
tjprescott, line 157: If you let `from` accept a nil value, you could handle this in the initializer and not need this if/else logic

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
```diff
@@ -47,25 +48,36 @@ public struct MessageContent: Codable {
...
+        } else {
+            self.participants = nil
+        }
+
+        // Deserialize the identifier model to CommunicationIdentifier
+        if let identifierModel = chatMessageContent.initiatorCommunicationIdentifier {
+            self.initiator = try IdentifierSerializer.deserialize(identifier: identifierModel)
+        } else {
+            self.initiator = nil
+        }
```
tjprescott, line 67: It seems like a simple model type should not throw.

sdk/communication/AzureCommunicationChat/Tests/ChatClientTests.swift
```diff
@@ -59,7 +59,7 @@ class ChatClientTests: XCTestCase {
         completionHandler: @escaping (String) -> Void
     ) {
         let participant = Participant(
-            id: id,
+            user: CommunicationUserIdentifier(identifier: id),
```
tjprescott, line 62: This seems redundant. This would be a case where using `_` to omit the label would be acceptable. 
`init(_ identifier: String, ...)`

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -365,15 +371,17 @@ public class ChatThreadClient {
...
-                displayName: $0.displayName,
-                shareHistoryTime: $0.shareHistoryTime
+        let chatParticipants = try participants.map { (participant) -> ChatParticipant in
+            let identifierModel = try IdentifierSerializer.serialize(identifier: participant.user)
+            return ChatParticipant(
+                communicationIdentifier: identifierModel,
+                displayName: participant.displayName,
+                shareHistoryTime: participant.shareHistoryTime
             )
         }
```
tjprescott, line 383: If this didn't throw before, it probably shouldn't start now.

sdk/communication/AzureCommunicationChat/Source/ChatThreadClient.swift
```diff
@@ -404,11 +412,14 @@ public class ChatThreadClient {
...
         completionHandler: @escaping HTTPResultHandler<Void>
-    ) {
+    ) throws {
+        // Construct CommunicationIdentifierModel from id
+        let identifierModel = try IdentifierSerializer
+            .serialize(identifier: CommunicationUserIdentifier(identifier: participantId))
         service
-            .removeChatParticipant(
+            .remove(
+                chatParticipant: identifierModel,
```
tjprescott, line 421: Same. This should probably not throw.

tjprescott (changes requested):
It seems there are some undocumented breaking changes. Also, several models were converted to now throw. If this can sensibly be avoided, that would be preferable.

CHANGELOG.md
RezaJooyandeh: Agree, Serializer are not to be used by the developers. If Autorest would support, we should have even make the `CommunicationIdentifierModel` and it is nested types internal.

tjprescott: Yes, but if they were internal I don't think you'd be allowed to expose them on a public object, so I'm not sure Autorest is the problem here. I think CommunicationIdentifier will have to be public.

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
knvsl: The IdentifierSerializer is what could throw @JoshuaLai not sure if this can be avoided

I did not want to just swallow the error so any method where the convenience layer has to convert types using the serializer now throws

tjprescott: It's not convenient if the convenience layer forces the user to wrap calls in `do/catch`  that otherwise would not have. Since `nil` is an acceptable value for `initiator`, it would be okay to use `try?` to make them fail as nil, or perhaps you can catch the deserialization error here, log it, and still set `initiator` to `nil`. Then it doesn't silently fail since the issue will be in the logs, but the developer doesn't need more error handling.

tjprescott:
There are additional comments here: https://github.com/Azure/azure-sdk/pull/2462

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
knvsl: For Message and MessageContent where the identifier can be `nil` that sounds good. 

- `Add participant`, `remove participant` and `create thread` all use the serializer to convert the participants
- If the serializer fails it means we can't make the request to add/remove the participants
- In the case of creating a thread we could still technically make the request without the participants - but it might be confusing to the user as the response they receive would show that there were no errors adding participants

What about returning a failure to the completion handler? Error message could clarify that the request was never actually sent:
```
do {
  // Use the identifier serializer that throws in here
} catch {
    let azureError = AzureError.client("Failed to send create thread request.", error)
    completionHandler(.failure(azureError), nil)
}
```

sdk/communication/AzureCommunicationChat/README.md
knvsl: Was already called in changelog - chat's readme is out of date

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
knvsl: For `ChatClient` `create(thread)` and `ChatThreadClient``add(participant)`, `remove(participant` I've opted to return the failure to the user to handle instead of throwing, let me know if you have other thoughts.

For the initializers that now throw - `Participant`, `Message`, `ReadReceipt`, `Thread`, `MessageContent`, `CreateThreadResult` I've opted to keep them as throwing. Reasoning is that those initializers are used by the convenience layer but users of the SDK can use the non-throwing initializers.

sdk/communication/AzureCommunication/Source/CommunicationIdentifierModel.swift
knvsl: They were, they are now generated inside of Chat's generated models. I don't believe they are intended to be exposed to users though. I will update the changelog on their removal

tjprescott (commented):
All of the updates look good. One final comment and it should be ready to merge.

sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift
```diff
@@ -46,24 +46,28 @@ public struct MessageContent: Codable {
...
+            // Deserialize the identifier model to CommunicationIdentifier
+            if let identifierModel = content.initiatorCommunicationIdentifier {
+                self.initiator = try IdentifierSerializer.deserialize(identifier: identifierModel)
+            } else {
+                self.initiator = nil
+            }
         } else {
-            self.initiator = nil
+            return nil
         }
```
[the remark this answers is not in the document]
knvsl: Actually this file is outdated :D change has been made here: https://github.com/Azure/azure-sdk-for-ios/blob/feature/chat-preview4/sdk/communication/AzureCommunicationChat/Source/Models/MessageContent.swift

tjprescott (approved):

PR #741, opened 2021-03-04
Update Signalling App

knvsl:
Fresh branch based off: https://github.com/knvsl/azure-sdk-for-ios/tree/knvsl/update-signalling
- Signalling is moved into Chat so test app just imports AzureCommunicationChat
- A couple syntax changes in ViewController.swift: Use ChatEventId enum instead of strings, do/catch around start and stop because they can throw exceptions if signalling client was not succesfully initialized


examples/SignalingTestApp/SignalingTestApp/ViewController.swift
```diff
@@ -0,0 +1,721 @@
...
+
+// COMMENT EVERYTHING OUT UNTIL TROUTER INTEGRATION IS UNBLOCKED
+
+import UIKit
+import AzureCommunicationChat
+import AzureCommunication
+import AzureCore
+
+struct Constants {
+    static let endpoint =  "<ACS_RESOURCE_ENDPOINT>"
```
glorialimicrosoft, line 16: Let's get these values from info.plist

PR #742, opened 2021-03-04
Update Signalling 

knvsl:
Created fresh branch because too many rebase conflicts
- Added convenience initializer to replace getSignallingClient() method
- In listeners use ChatEventId as parameter instead of string
- Remove fatalError in CommunicationSignallingClient
- Throw exception in start/stop if signalling client doesn't exist, so user knows notifications are not working

After preview4 PR goes in I will rebase this branch again so we are up to date with master. Next I will update the signalling schema


glorialimicrosoft (approved):
Looks good :) Note that we won't need the trouter header files and module maps once we can successfully consume the framework that the trouter team provides.

PR #748, opened 2021-03-05
Update check version scripts

tjprescott:
The old PS Verify-Versions.ps script only checked podspecs. The new Python scripts verify Package.swift, podspecs, and Xcodeproj files all agree. 

- `version.py verify`: verifies that all versions are internally consistent. Used in CI.
- `version.py current`: returns the current version, assuming they have already been checked for consistency. This will likely be used in future shell scripts.
- `version.py update`: Takes `old` and `new` versions as a parameter and updates all Xcodeproj files, podspecs, Package.swift, jazzy files, etc. This will be run manually as part of the current release process.


eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -122,9 +122,12 @@ jobs:
 
       - script: |
           pod lib lint --quick
-        displayName: Lint PodSpecs 
+          pod spec lint --use-modular-headers
```
JoshuaLai, line 125: Great change

Scripts/check_versions.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/python3
```
weshaggard, line 1: Any reason these cannot go under eng/scripts? That is the convention we use in other places to try and keep the root of the repo a little cleaner.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -122,9 +122,12 @@ jobs:
...
+        displayName: Lint PodSpecs
 
-      - template: /eng/pipelines/templates/steps/verify-versions.yml
```
weshaggard, line 127: Probably worth just updating the verify-versions.yml file instead of removing it.

Scripts/get_version.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/python3
```
weshaggard, line 1: What context to you plan to use this? I know we talked about some helpers for the Prepare-Release scripts and I think we can use this there but I'm curious if you have other plans for it as well?

tjprescott: I had written this before then as preparation for a legitimate release stage in our pipeline to update the git tag for the release. I wanted to do that with a bash script (which I loathe) but trying to parse the results of `git describe` proved so frustrating and brittle that I moved to python. The idea here is that in my bash script, I can now simply say:
```bash
version=python3 version.py current
# do other bash-y things here
```

tjprescott:
I consolidated the different scripts into a single `version.py` script (similar to Django's `manage.py`) to take advantage of the shared code and minimize maintenance burden.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
tjprescott: We didn't have any other step templates so I tried to match the format in `archetype-sdk-client.yml` which has a bunch of small `script` steps.

Scripts/check_versions.py
tjprescott: Definitely should be able to.

tjprescott: Done.

Scripts/get_version.py
weshaggard: Cool before you go and write your own step of scripts/steps that do the git tagging please see my comment in your other PR as I'd like to use our common scripts as much as possible. We will need to light up some common infrastructure for both the tagging, github.io common publishing, and prepare-release. So the more we light up the more common infra you guys can leverage.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
weshaggard: Given this is likely the only place then doing it directly as a script step makes sense.

PR #755, opened 2021-03-10
Update CHANGELOG.md with ACS Calling SDK changes

kagbakpem:


CHANGELOG.md
```diff
@@ -16,7 +21,21 @@
 - `ChatThreadClient` `remove(participant)` method now accepts `CommunicationIdentifier` instead of a string
 - For `Participant` renamed `user` property to `id`
 
+**Azure Communication Calling**
+- `Renderer` renamed to `VideoStreamRenderer`.
```
JoshuaLai, line 25: Should this be considered a breaking change?

CHANGELOG.md
```diff
@@ -16,7 +21,21 @@
...
+- `AudioDeviceInfo` removed from `DeviceManager`, please use iOS system API's in your application to switch between audio devices.
```
JoshuaLai, line 26: Should this be considered a breaking change?

JoshuaLai (changes requested):
Looks good just to tiny questions

CHANGELOG.md
kagbakpem: Yes, as it requires code change from the end user perspective to keep using the SDK?

JoshuaLai: Ah sorry I misinterpreted looked like it was in the wrong section my mistake!

JoshuaLai (approved):

CHANGELOG.md
kagbakpem: Yes, as this drop will break previous contract. Code change is needed on User end in order to consume this if the previous API was used.

tjprescott (approved):

PR #763, opened 2021-03-12
[Communication] - Identifiers - Init for user and unknown dropping _

JoshuaLai:


JoshuaLai:
Related to https://github.com/Azure/Communication/issues/227

JoshuaLai:
Pending update to Changelog if we are going to re-introduce this change

tjprescott:
Generally, I would say the original code is correct from the Swift user perspective and we should no break that for the sake of ObjC. Perhaps instead of using `@objcMembers` we declare `@objc` on individual members and do not expose `init(_ identifier:)` with `@objc`, instead only exposing the `init(identifier:)` to ObjC using a technique described here (https://izziswift.com/how-to-expose-swift-functions-only-to-objc/).

JoshuaLai:
> Generally, I would say the original code is correct from the Swift user perspective and we should not break that for the sake of ObjC. Perhaps instead of using `@objcMembers` we declare `@objc` on individual members and do not expose `init(_ identifier:)` with `@objc`, instead only exposing the `init(identifier:)` to ObjC using a technique described here (https://izziswift.com/how-to-expose-swift-functions-only-to-objc/).

Are you suggesting we just change the object to 
```
public class CommunicationUserIdentifier: NSObject, CommunicationIdentifier {
    @objc public let identifier: String
    @objc(initWithIdentifier:)
    public init(_ identifier: String) {
        self.identifier = identifier
    }
}
```
I was actually wondering if it would make more sense to change the method to `init(with identifer:` I feel like that would at least create some consistency between Swift and ObjC.

tjprescott (changes requested):
Needs changelog entry. I am onboard with the ObjC naming if you think it's better for you customers.

JoshuaLai:
@tjprescott @raosanat Updated the change log with the new breaking change.

tjprescott (approved):

tjprescott (approved):

PR #772, opened 2021-03-18
[DONOTMERGE] - Communication - New swiftinterface

JoshuaLai:


examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,80 @@
...
+  @objc public static var Public: AzureCommunication.CommunicationCloudEnvironment
+  @objc public static var Dod: AzureCommunication.CommunicationCloudEnvironment
+  @objc public static var Gcch: AzureCommunication.CommunicationCloudEnvironment
+  @objc public init(environmentValue: Swift.String)
+  @objc public func getEnvironmentValue() -> Swift.String
+  public static func == (lhs: AzureCommunication.CommunicationCloudEnvironment, rhs: AzureCommunication.CommunicationCloudEnvironment) -> Swift.Bool
+  @objc deinit
+  @objc override dynamic public init()
+}
+public typealias TokenRefresherClosure = (@escaping AzureCommunication.TokenRefreshOnCompletion) -> Swift.Void
```
tjprescott, line 60: `TokenRefreshOnCompletion` implies this should be called `TokenRefreshHandler` to conform with the guidlines:
https://azure.github.io/azure-sdk/ios_design.html#ios-network-callback-suffix

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,80 @@
...
+@objc @objcMembers public class CommunicationTokenRefreshOptions : ObjectiveC.NSObject {
+  @objc public init(initialToken: Swift.String? = nil, refreshProactively: Swift.Bool = false, tokenRefresher: @escaping AzureCommunication.TokenRefresherClosure)
```
tjprescott, line 62: Is the `TokenRefreshClosure` something that does the magic of refreshing the token, or is it called when a token is refreshed?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,80 @@
...
+  @objc deinit
+  @objc override dynamic public init()
+}
+public class CommunicationPolicyTokenCredential : AzureCore.TokenCredential {
+  public init(_ credential: AzureCommunication.CommunicationTokenCredential)
+  public func token(forScopes _: [Swift.String], completionHandler: @escaping AzureCore.TokenCompletionHandler)
+  public func validate() throws
+  @objc deinit
+}
```
tjprescott, line 71: Is this intentionally not available with `@objc`?  Since it is based on AzureCore I would suspect that's the case.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,80 @@
...
+public typealias TokenRefresherClosure = (@escaping AzureCommunication.TokenRefreshOnCompletion) -> Swift.Void
+@objc @objcMembers public class CommunicationTokenRefreshOptions : ObjectiveC.NSObject {
+  @objc public init(initialToken: Swift.String? = nil, refreshProactively: Swift.Bool = false, tokenRefresher: @escaping AzureCommunication.TokenRefresherClosure)
+  @objc deinit
+  @objc override dynamic public init()
+}
+public class CommunicationPolicyTokenCredential : AzureCore.TokenCredential {
+  public init(_ credential: AzureCommunication.CommunicationTokenCredential)
+  public func token(forScopes _: [Swift.String], completionHandler: @escaping AzureCore.TokenCompletionHandler)
```
tjprescott, line 68: Scopes is not used? In that case, can you default it to an empty list so the customer doesn't have to submit one?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,80 @@
...
+  public func validate() throws
+  @objc deinit
+}
+public typealias CommunicationTokenCompletionHandler = (AzureCommunication.CommunicationAccessToken?, Swift.Error?) -> Swift.Void
+public typealias TokenRefreshOnCompletion = (Swift.String?, Swift.Error?) -> Swift.Void
+@objc @objcMembers public class CommunicationTokenCredential : ObjectiveC.NSObject {
+  @objc public init(token: Swift.String) throws
+  @objc public init(with option: AzureCommunication.CommunicationTokenRefreshOptions) throws
```
tjprescott, line 76: Looking at the guidelines, we always use `withOptions options:`. That's what I would recommend here. Also, is it possible to defer the `throwing` until the Credential object is used? With our Storage UX study, participants did not like when creating the credential could throw... and then creating the client with the credential could throw... so we consolidated toward making just the client creation fail, rather than the credential creation.

tjprescott (changes requested):

examples/x86_64.swiftinterface
JoshuaLai: This is called when the token is refreshed

examples/x86_64.swiftinterface
JoshuaLai: Yes we have the Core import wrapped
```swift
#if canImport(AzureCore)
    import AzureCore
#endif
```

examples/x86_64.swiftinterface
JoshuaLai: I don't think we can defer the throwing here. Its a little complex but the `CommunicationTokenCredential` when trying to create a `StaticTokenCredential` uses the `JWTTokenParser`. If the token passed in does not conform to the correct jwt format we want to throw. The throw comes from the JWT validation essentially.

JoshuaLai: I can update the method parameters, I will have to double check what the interface created for ObjC will be. It might look a little silly, so I will have and update it and make something specific for that case.

RezaJooyandeh: +1 on what Josh said, Sever issues or connectivity, etc. will not cause exceptions here. Throwing only happens if the developers passes the wrong argument. In which case delaying the throw to the clients using it would burry the issue down and I think it would get very hard for developers to figure out what they did wrong.

examples/x86_64.swiftinterface
JoshuaLai: This is contosos method that we will call to return a new token.

examples/x86_64.swiftinterface
JoshuaLai: `CommunicationPolicyTokenCredential` should not even be public so ideally we should make this internal to chat.

examples/x86_64.swiftinterface
RezaJooyandeh: We talked offline. Considering that the exception is thrown during the initialization of the credential not the client, we can keep it as is

examples/x86_64.swiftinterface
tjprescott: @JoshuaLai I do wonder if `TokenRefreshHandler` is too vague. Perhaps `TokenRefreshCompletionHandler`? I'm okay with either.

Also, I would recommend that since you have a method that handles the refreshing and a callback that happens when a token is refreshed that you consider adding something like a `TokenRefreshDelegate` for developers that prefer that pattern. I would suspect many ObjC developers would be more familiar with it and prefer it. 
https://azure.github.io/azure-sdk/ios_design.html#delegates

This is merely a suggestion for future work. It would be a non-breaking additive change.

tjprescott (approved):

JoshuaLai:
Making a note of your comments @tjprescott and going to add an item to our backlog 
Closing this PR

PR #775, opened 2021-03-19
[Communication] - Chat - Refactor of class name

JoshuaLai:


JoshuaLai:
Test was removed since  adapter is just a pass through and the test is already being performed in the communication common package.

RezaJooyandeh (approved):

PR #786, opened 2021-03-26
Pin MSAL version to avoid broken release of MSAL 1.1.16

tjprescott:
Workaround for:
https://github.com/AzureAD/microsoft-authentication-library-for-objc/issues/1280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@
   libraries.
 
 ### Breaking Changes
+
 **Azure Communication Chat Service**
   - The `baseUrl` parameter has been renamed to `endpoint` in the `AzureCommunicationChatClient` initializers.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -136,6 +136,7 @@
   libraries.
 
 ### Breaking Changes
+
```
JoshuaLai, line 139: Did you mean to make a change log entry here? @tjprescott

tjprescott: Originally yes, just to run the CI and verify it was failing.

JoshuaLai: Oh this I assume was just a mistake cause its all the way back in beta.5

knvsl (approved):

JoshuaLai:
@tjprescott Do we care if we pin it to `1.1.2` as opposed to `1.1.15`. Or is this just telling consumers `15` is the highest we are willing to go?

tjprescott:
The way it is written it will always grab 1.1.15

PR #789, opened 2021-03-29
Mechanism for Mono to Clone Swift PM Repos

tjprescott:
The goal here is for us to have fine-grained, independent version control for SwiftPM modules the way we do in other SDKs and with CocoaPods while maintaining the benefits of a mono-repo.

To facilitate this, we will have a per-module Package.swift and CHANGELOG.md. Each module will be represented in `manifest.yml` with its path in the mono-repo and its mirror repo name.

Whenever the mono-repo is updated, we can update the mirror repo with:

`python3 eng/scripts/sync_repo.py <MODULE>`

This will copy that branch to the root of the mirror, which can then be target by SwiftPM. In the current model, users target the "AzureSDK" package, and then specify individual modules, whereas with this change, the target will be the module.

The current example of this is AzureCore, which will mirror to `azure-sdk-for-ios-core`.


Package.swift
```diff
@@ -35,7 +35,6 @@ let package = Package(
         .macOS(.v10_15), .iOS(.v12), .tvOS(.v12)
     ],
     products: [
```
weshaggard, line 37: Don't we want to remove this root swift file completely?

manifest.yaml
```diff
@@ -0,0 +1,29 @@
+AzureCore:
```
weshaggard, line 1: What is this file used for? Ideally we should try to avoid yet another list of projects and versions to maintain if we can.

Package.swift
tjprescott: Yes, we will. At that point, no one would be able to target this repo with SPM. For this PR, I'm only separating out AzureCore to see what that would look like.

manifest.yaml
tjprescott: Right now the script only uses the source and mirror entries. Unless we have a deterministic way of mapping SDKs to repos, I don't see how we will avoid this. `Package.swift` files don't actually specify versions, so the idea behind version would be a simple place to acquire the correct version for tagging. We could potentially pull that from the podspec though.

Package.swift
weshaggard: OK once we do move over I expect folks can still use an older release tag for the preview packages we did ship from the repo still.

manifest.yaml
weshaggard: Ideally we would only have one place that defines the version. If swift doesn't need it perhaps we could use the pod. I know some languages also embed the version in source code so it can also be used by code if needed.

weshaggard: As for the other metadata here could we gather it from the swift files instead of explicitly listing it? For most languages we essentially traverse the sdk looking for some sort of project file and then read the metadata from there.

Package.swift
tjprescott: Yes, the old release tags will work

mitchdenny:
This PR seems to have two strategies in play. One strategy that has been discussed is having a copy of all the modules in a separate repository (synchronized from the mono-repo). But there also appears to be a reference to an ```azure-sdk-for-ios-core``` repository as well (which makes me think that this is just breaking out core separately, and will be managed as a separate repo (incl the engineering system etc).

tjprescott:
> This PR seems to have two strategies in play. One strategy that has been discussed is having a copy of all the modules in a separate repository (synchronized from the mono-repo). But there also appears to be a reference to an azure-sdk-for-ios-core repository as well (which makes me think that this is just breaking out core separately, and will be managed as a separate repo (incl the engineering system etc).

@mitchdenny thanks for the catch. I updated the reference to `SwiftPM-AzureCore`. Also, @weshaggard I updated the `sync_repo.py` script so it no longer requires any metadata file.

tjprescott:
@weshaggard @mitchdenny this PR should be ready for review.

mitchdenny (commented):
One comment around the copy logic. It would really need to delete-then-copy the contents of the directory otherwise you won't be catching renames etc.

Another thing that needs to be worked out is the tagging strategy when a release goes out. Is the plan just to replace the contents of the repo and then commit and tag? So we could end up with something like this:

commit1: 1.0.0-beta.11
commit2: 1.0.0
commit3: 1.1.0-beta.1
commit4: 1.0.1

... just based on blindly doing a copy/tag. It probably doesn't make much difference since folks should always be going from a tag in the swiftpm-* repos, but if anyone looked at the history of the repo it would be totally confusing. To try and mitigate that I think it might make sense to pre-pend something in the README.md file which provides a back-pointer to the actual commit in the mono-repo so folks know to navigate back there to find the actual state of the universe.

I think there will also need to be some content in the README that explains how/why these repos are done this way. Imagine a scenario where someone inherits an iOS code-base and they go and look at their dependencies and they see one of our swiftpm-* links. They may not have previously worked with the Azure SDK before so wouldn't know why we have these satellite repos.

eng/scripts/sync_repo.py
```diff
@@ -0,0 +1,75 @@
...
+            _log_error_and_quit(f'Source path does not exist: {source_path}')
+    except IndexError:
+        _log_error_and_quit(f'Could not find {target}.podspec.json')
+
+    dest_path = os.path.abspath(os.path.join(ROOT, '..', f'SwiftPM-{target}'))
+    if not os.path.exists(dest_path):
+        _log_error_and_quit(f'Destination path does not exist: {dest_path}')
+
+    # copy the subfolder to the root of the SwiftPM repo
+    _copy(source_path, dest_path)
```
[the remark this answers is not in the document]
tjprescott: Good catch. I have updated the script so it will clear the folder before copying.

tjprescott:
> Another thing that needs to be worked out is the tagging strategy when a release goes out. Is the plan just to replace the contents of the repo and then commit and tag? So we could end up with something like this:
> 
> commit1: 1.0.0-beta.11
> commit2: 1.0.0
> commit3: 1.1.0-beta.1
> commit4: 1.0.1

There would be corresponding tags on the monorepo like:

commit1: AzureCore_1.0.0-beta.11
commit2: AzureCore_1.0.0 (albeit not anytime soon)
commit3: AzureCore_1.1.0-beta.1
commit4: AzureCore_1.0.1

However, if that were the sequence in both repos, I don't see how either would be any less confusing. \

> I think there will also need to be some content in the README that explains how/why these repos are done this way. Imagine a scenario where someone inherits an iOS code-base and they go and look at their dependencies and they see one of our swiftpm-* links. They may not have previously worked with the Azure SDK before so wouldn't know why we have these satellite repos.

I think, most likely, they wouldn't think twice about the SwiftPM repo but would be surprised to learn of the existence of the mono-repo. There would be no reference to `azure-sdk-for-ios` in their files, but if they wanted to file a bug or contribute, the README would direct them to the mono-repo. But you are right, we should call this out in some fashion somewhere.

sdk/communication/AzureCommunication/Package.swift
```diff
@@ -0,0 +1,69 @@
...
+    ],
+    products: [
+        .library(name: "AzureCommunication", targets: ["AzureCommunication"])
+    ],
+    dependencies: [],
+    targets: [
+        // Build targets
+        .target(
+            name: "AzureCommunication",
+            dependencies: ["AzureCore"],
```
weshaggard, line 45: I don't understand swift packages that much but don't you need to list a version and/or repository? Does this essentially take a source level dependency the way it is currently configured?

sdk/communication/AzureCommunication/Package.swift
```diff
@@ -0,0 +1,69 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import PackageDescription
+
+let package = Package(
+    name: "AzureCommunication",
```
weshaggard, line 33: Do the project files not contain any sort of version?

eng/scripts/sync_repo.py
```diff
@@ -0,0 +1,86 @@
...
+        if os.path.isfile(obj_path):
+            os.unlink(obj_path)
+        else:
+            shutil.rmtree(obj_path)
+
+    # copy the subfolder to the root of the SwiftPM repo
+    _copy(source_path, dest_path)
+
+    # copy the LICENSE, CCONTRIBUTING.md, and .gitignore files
+    for fname in ["LICENSE", "CONTRIBUTING.md", ".gitignore"]:
```
weshaggard, line 74: On top of these did we plan to also mutate the readme in the root to point back to the mono repo? Also we should probably update the repo description with a pointer as well.

Package.swift
```diff
@@ -45,17 +44,6 @@ let package = Package(
...
-            dependencies: [],
-            path: "sdk/core/AzureCore",
-            exclude: [
-                "README.md",
-                "Tests",
-                "Source/Supporting Files"
-            ],
-            sources: ["Source"]
-        ),
         .target(
```
weshaggard, line 47: Shouldn't these also be removed now?

examples/AzureSDKSmokeTestSwiftPM/AzureSDKSmokeTestSwiftPM.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,360 @@
+// !$*UTF8*$!
```
weshaggard, line 1: Where does this get build and ran? Also how does it actually pull in the swift package references? Perhaps I'm missing something but I would expect to pull them in somewhere.

sdk/communication/AzureCommunication/Package.swift
mitchdenny: No the git reference is the version. Swift is 100% tied to Git as a version control mechanism.

sdk/communication/AzureCommunication/Package.swift
tjprescott: No they do not.

Package.swift
tjprescott: Removed the root Package.swift.

sdk/communication/AzureCommunication/Package.swift
tjprescott: The target dependencies reference the package dependencies by name. The package dependencies will reference the repos (in this case, the SwiftPM-* repos) and either a version, commit, or branch.

Package.swift
tjprescott: Removed.

examples/AzureSDKSmokeTestSwiftPM/AzureSDKSmokeTestSwiftPM.xcodeproj/project.pbxproj
tjprescott: I think I had that in there for testing. For now I will remove it and re-add it in a later PR, as it should be part of CI the way the CocoaPods smoke test is.

eng/scripts/sync_repo.py
tjprescott: So, I updated the READMEs so that they target the appropriate repos (either the SwiftPM-* ones or the mono repo), so I'm not sure we need to do any mutation. If we get a lot of people expressing confusion, I'll address the issue then.

sdk/communication/AzureCommunication/Package.swift
weshaggard: Where does that repo name and tag info get stored for these dependencies?

sdk/communication/AzureCommunicationChat/Package.swift
```diff
@@ -30,84 +30,49 @@
...
-        .library(name: "AzureCommunication", targets: ["AzureCommunication"]),
-        .library(name: "AzureCommunicationChat", targets: ["AzureCommunicationChat"]),
+        .library(name: "AzureCommunicationChat", targets: ["AzureCommunicationChat"])
     ],
     dependencies: [
+        .package(name: "AzureCore", url: "https://github.com/Azure/SwiftPM-AzureCore.git", .branch("master")),
+        .package(
+            name: "AzureCommunication",
+            url: "https://github.com/Azure/SwiftPM-AzureCommunication.git",
+            .branch("master")
```
weshaggard, line 45: I assume at some point these will turn into a release tags over a branch reference to master.

sdk/communication/AzureCommunication/Package.swift
weshaggard: I see that information listed under the dependencies section above now.

weshaggard (approved):
Overall this looks like a good starting point. Once you finish your testing lets get it in and start iterating.

sdk/communication/AzureCommunication/Package.swift
tjprescott: Here's the example from Chat

let package = Package(
    name: "AzureCommunicationChat",
    platforms: [
        .iOS(.v12)
    ],
    products: [
        .library(name: "AzureCommunicationChat", targets: ["AzureCommunicationChat"])
    ],
    dependencies: [
        **.package(name: "AzureCore", url: "https://github.com/Azure/SwiftPM-AzureCore.git", .branch("master")),**
    ],
    targets: [
        // Build targets
        .target(
            name: "AzureCommunicationChat",
            **dependencies: ["AzureCore"],**
            path: "Source",
            exclude: [
                "README.md",
                "Tests",
                "Source/Supporting Files",
                "LICENSE"
            ]
        ),
...

PR #803, opened 2021-04-08
Exclude Chat's swagger config file in Package.swift

knvsl:
Prevent the warnings

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -74,7 +74,8 @@ let package = Package(
             exclude: [
                 "README.md",
                 "Tests",
-                "Source/Supporting Files"
+                "Source/Supporting Files",
+                "Swagger/CONFIG.md"
             ],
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -74,7 +74,8 @@ let package = Package(
             exclude: [
                 "README.md",
                 "Tests",
-                "Source/Supporting Files"
+                "Source/Supporting Files",
+                "Swagger/CONFIG.md"
```
tjprescott, line 78: Really this file should be removed. The config should live in the the rest API specs repo's README.md

tjprescott (changes requested):

Package.swift
knvsl: closing this - will move any config to a readme in specs repo like you mentioned

PR #818, opened 2021-04-14
Make internal model & generated func to internal access level

Leoaqr:


sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
```diff
@@ -17,7 +17,7 @@ import Foundation
 // swiftlint:disable type_body_length
 
 public final class Chat {
-    public let client: AzureCommunicationChatClient
+    internal let client: AzureCommunicationChatClient
```
knvsl, line 20: Should AzureCommunicationChatClient also be internal?

knvsl (approved):
If live tests are passing we can manually change for now, but should document this for autorest.swift that autorest needs to match these changes.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
Leoaqr: Yup, all tests passed. We can't set AzureCommunicationChatClient to internal, because we pick up AzureCommunicationChatClient.apiVersion in AzureCommunicationChatClientOptions, where we call AzureCommunicationChatClientOptions as an input param in ChatThreadClient init()

tjprescott (changes requested):
Ensuring this does not get merged until reviewed by AzureSDK team.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
```diff
@@ -29,16 +29,16 @@ public final class Chat {
     ///    - options: A list of options for the operation
     ///    - completionHandler: A completion handler that receives a status code on
     ///     success.
-    public func create(
-        chatThread: CreateChatThreadRequest,
+    internal func create(
```
tjprescott, line 32: Why is the `Chat` class public if all its methods are `internal`?  I don't understand what you are supposed to do with this.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
```diff
@@ -29,16 +29,16 @@ public final class Chat {
...
         withOptions options: CreateChatThreadOptions? = nil,
-        completionHandler: @escaping HTTPResultHandler<CreateChatThreadResult>
+        completionHandler: @escaping HTTPResultHandler<CreateChatThreadResultInternal>
     ) {
         let dispatchQueue = options?.dispatchQueue ?? client.commonOptions.dispatchQueue ?? DispatchQueue.main
 
         // Create request parameters
         let params = RequestParameters(
-            (.header, "repeatability-request-id", options?.repeatabilityRequestId, .encode), (
+            (.header, "repeatability-Request-Id", options?.repeatabilityRequestId, .encode), (
```
tjprescott: Investigate if this is a Swagger issue or autorest.

tjprescott (changes requested):
Don't understand why your public classes have no public methods.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
Leoaqr: It's generated by autorest

sdk/communication/AzureCommunicationChat/Source/Generated/Models/ChatThreadPropertiesInternal.swift
```diff
@@ -16,19 +15,19 @@ import Foundation
 // swiftlint:disable cyclomatic_complexity
 
 /// Chat thread.
-public struct ChatThreadPropertiesInternal: Codable {
+internal struct ChatThreadPropertiesInternal: Codable {
```
tjprescott, line 18: This should be handled by autorest.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
tjprescott: This is a Swagger issue. 
https://github.com/Azure/azure-rest-api-specs/blob/d95c18e2d5fc678a1453c454d746fdce22d30122/specification/communication/data-plane/Microsoft.CommunicationServicesChat/stable/2021-03-07/communicationserviceschat.json#L853
Autorest is generating what the Swagger specifies, so it needs to be fixed there.

sdk/communication/AzureCommunicationChat/Source/Generated/Models/ChatError.swift
```diff
@@ -14,21 +14,20 @@ import Foundation
 // swiftlint:disable identifier_name
 // swiftlint:disable line_length
 // swiftlint:disable cyclomatic_complexity
-
 /// The Communication Services error.
-internal struct ChatError: Codable, Swift.Error {
+public final class ChatError: Codable, Swift.Error {
```
tjprescott, line 18: If you want this public, then just don't add it to remap-models.

sdk/communication/AzureCommunicationChat/Source/Generated/Chat.swift
Leoaqr: Updated, I reran the autorest, verified its in lowercase.

PR #820, opened 2021-04-15
[CommunicationTokenCredential] introduce the delegate pattern

JoshuaLai:
As per conversation during arch review: https://github.com/Azure/azure-sdk-for-ios/pull/772#discussion_r599747198
Introducing the delegate pattern.
Still to come need to introduce the ObjC unit tests


sdk/communication/AzureCommunication/Source/Authentication/CommunicationTokenCredential.swift
```diff
@@ -63,12 +87,38 @@ public typealias TokenRefreshHandler = (String?, Error?) -> Void
         )
     }
 
+    /**
+     Creates a CommunicationTokenCredential that automatically refreshes the token.
+     - Parameter delegate: Delegate class conforming to `TokenCredentialDelegate`
+     - Parameter option: Options for how the token will be refreshed
+     - Throws: `AzureError` if the provided token is not a valid user token.
+     */
+    public init(
```
JoshuaLai, line 96: @sacheu  @tjprescott I wonder does this make sense to have an init with with delegate when using options? 
It seems like an odd contradiction. The options object itself has a `tokenRefresher` closure we pass in, but if they are using the delegate pattern they would expect us to call the new `onTokenRetrieved` method from the delegate. 

So does it make more sense to create a new init looking like this 
```swift 
public init(withDelegate delegate: TokenCredentialDelegate, refreshProactively: Bool, initialToken: String) throws
```

tjprescott: Are there other options? I think the approach you suggest here makes sense, so long as any additional options are still available to the delegate.

JoshuaLai: One simple solution could be to make a change `var tokenRefresher: TokenRefresher?`

Screen Shot 2021-04-15 at 1 12 04 PM

Or we can create a new options object where there is no closure 
```swift
@objcMembers public class CommunicationTokenRefreshSimpleOptions: NSObject {
    var initialToken: String?
    var refreshProactively: Bool
}
```

RezaJooyandeh: This is very similar to what we did in C#:
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/communication/Azure.Communication.Common/src/CommunicationTokenRefreshOptions.cs

JoshuaLai:
@RezaJooyandeh  @sacheu Chatted with @tjprescott  offline on a few ideas on how to make the delegate a little more intuitive for developers to use. This is the final decision we came up with. 

Summary:
Introduction of a new method. And it will do everything the old one used to do, retrieve an access token from the credential but the only difference will be it will call the delegate method instead. If no delegate is assigned the user will be prompted with a warning in the logs.  (Logger class was copied from core over to communication)
```
public func token() 
```
Updated the old method to have a new check in the odd case the client has assigned a delegate but still wanted to call tis method we only call the completionHandler and log a warning saying we are ignoring the delegate. 
```
public func token(completionHandler: @escaping CommunicationTokenCompletionHandler)
```

JoshuaLai:
Having a chat with @RezaJooyandeh and @tjprescott about this. 
We might table this for now. If we find a compelling use case we can revisit this design.

PR #851, opened 2021-05-07
Add ACS Chat swiftinterface

Leoaqr:


examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,614 @@
...
+  public static func create(chatEvent chatEventId: AzureCommunicationChat.ChatEventId, from request: TrouterClientIos.TrouterRequest) throws -> Any
+}
+public struct ChatParticipantInternal : Swift.Codable {
+  public let communicationIdentifier: AzureCommunicationChat.CommunicationIdentifierModel
+  public let displayName: Swift.String?
+  public let shareHistoryTime: AzureCore.Iso8601Date?
+  public init(communicationIdentifier: AzureCommunicationChat.CommunicationIdentifierModel, displayName: Swift.String? = nil, shareHistoryTime: AzureCore.Iso8601Date? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
```
tjprescott, line 90: This should not be public.

tjprescott (changes requested):
There's still a ton of things in here marked `Internal` that are public. There should be no references to any properties that end in `Internal`

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public func add(participants: [AzureCommunicationChat.ChatParticipant], withOptions options: AzureCommunicationChat.AddChatParticipantsOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCommunicationChat.AddChatParticipantsResult>)
+  public func remove(participant participantIdentifier: AzureCommunicationCommon.CommunicationIdentifier, withOptions options: AzureCommunicationChat.RemoveChatParticipantOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<Swift.Void>)
+  public func listParticipants(withOptions options: AzureCommunicationChat.ListChatParticipantsOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCore.PagedCollection<AzureCommunicationChat.ChatParticipant>>)
+  @objc deinit
+}
+public typealias EventHandler = (Any, AzureCommunicationChat.ChatEventId) -> Swift.Void
+public enum TrouterEventUtil {
+  public static func getIdentifier(from mri: Swift.String) -> AzureCommunicationCommon.CommunicationIdentifier
+  public static func create(chatEvent chatEventId: AzureCommunicationChat.ChatEventId, from request: TrouterClientIos.TrouterRequest) throws -> Any
```
tjprescott, line 73: Looks like this still needs to be addressed.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public var id: AzureCommunicationCommon.CommunicationIdentifier?
+  public var displayName: Swift.String?
+  public var shareHistoryTime: AzureCore.Iso8601Date?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class SignalingChatThreadProperties {
+  public var topic: Swift.String
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class BaseChatEvent {
```
tjprescott, line 319: This class is undocumented.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public var senderDisplayName: Swift.String?
+  public var recipient: AzureCommunicationCommon.CommunicationIdentifier?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class BaseChatThreadEvent {
+  public var threadId: Swift.String
+  public var version: Swift.String
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class BaseChatMessageEvent : AzureCommunicationChat.BaseChatEvent {
```
tjprescott, line 331: Undocumented

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class BaseChatEvent {
+  public var threadId: Swift.String
+  public var sender: AzureCommunicationCommon.CommunicationIdentifier?
+  public var senderDisplayName: Swift.String?
+  public var recipient: AzureCommunicationCommon.CommunicationIdentifier?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class BaseChatThreadEvent {
```
tjprescott, line 326: Undocumented

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+}
+public enum IdentifierSerializer {
+  public static func deserialize(identifier: AzureCommunicationChat.CommunicationIdentifierModel) throws -> AzureCommunicationCommon.CommunicationIdentifier
+  public static func assertOneNestedModel(_ identifier: AzureCommunicationChat.CommunicationIdentifierModel) throws
+  public static func serialize(identifier: AzureCommunicationCommon.CommunicationIdentifier) throws -> AzureCommunicationChat.CommunicationIdentifierModel
+}
+extension CommunicationCloudEnvironmentModel {
+  public static var Public: AzureCommunicationChat.CommunicationCloudEnvironmentModel
+  public static var Dod: AzureCommunicationChat.CommunicationCloudEnvironmentModel
+  public static var Gcch: AzureCommunicationChat.CommunicationCloudEnvironmentModel
```
tjprescott, line 264: What is the purpose of these static variables if the `CommunicationCloudEnvironmentModel` enum already contains these values?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public let content: Swift.String?
+  public init(content: Swift.String? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public enum IdentifierSerializer {
+  public static func deserialize(identifier: AzureCommunicationChat.CommunicationIdentifierModel) throws -> AzureCommunicationCommon.CommunicationIdentifier
+  public static func assertOneNestedModel(_ identifier: AzureCommunicationChat.CommunicationIdentifierModel) throws
+  public static func serialize(identifier: AzureCommunicationCommon.CommunicationIdentifier) throws -> AzureCommunicationChat.CommunicationIdentifierModel
+}
```
tjprescott, line 260: This seems like it should be internal?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public func add(participants: [AzureCommunicationChat.ChatParticipant], withOptions options: AzureCommunicationChat.AddChatParticipantsOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCommunicationChat.AddChatParticipantsResult>)
+  public func remove(participant participantIdentifier: AzureCommunicationCommon.CommunicationIdentifier, withOptions options: AzureCommunicationChat.RemoveChatParticipantOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<Swift.Void>)
+  public func listParticipants(withOptions options: AzureCommunicationChat.ListChatParticipantsOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCore.PagedCollection<AzureCommunicationChat.ChatParticipant>>)
+  @objc deinit
+}
+public typealias EventHandler = (Any, AzureCommunicationChat.ChatEventId) -> Swift.Void
+public enum TrouterEventUtil {
+  public static func getIdentifier(from mri: Swift.String) -> AzureCommunicationCommon.CommunicationIdentifier
```
tjprescott, line 72: it's not clear from the documentation what an "mri" is.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+@_hasMissingDesignatedInitializers public class SignalingChatParticipant {
+  public var id: AzureCommunicationCommon.CommunicationIdentifier?
+  public var displayName: Swift.String?
+  public var shareHistoryTime: AzureCore.Iso8601Date?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class SignalingChatThreadProperties {
```
tjprescott, line 315: Why is this a class instead of a struct?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct CreateChatThreadRequest : Swift.Codable {
+  public let topic: Swift.String
+  public let participants: [AzureCommunicationChat.ChatParticipant]?
+  public init(topic: Swift.String, participants: [AzureCommunicationChat.ChatParticipant]? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+@_hasMissingDesignatedInitializers public class SignalingChatParticipant {
```
tjprescott, line 309: Why is this a class instead of a struct?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+@_hasMissingDesignatedInitializers public class ChatMessageEditedEvent : AzureCommunicationChat.BaseChatMessageEvent {
+  public var message: Swift.String
+  public var editedOn: AzureCore.Iso8601Date?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class ChatMessageDeletedEvent : AzureCommunicationChat.BaseChatMessageEvent {
+  public var deletedOn: AzureCore.Iso8601Date?
+  @objc deinit
+}
+@_hasMissingDesignatedInitializers public class TypingIndicatorReceivedEvent : AzureCommunicationChat.BaseChatEvent {
```
tjprescott, line 351: It seems all of these Event classes should also be `struct`s, unless there is some specific reason they are `class`es.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct UpdateChatThreadRequest : Swift.Codable {
+  public let topic: Swift.String?
+  public init(topic: Swift.String? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct AzureCommunicationChatClientOptions : AzureCore.ClientOptions {
```
tjprescott, line 210: Would you prefer to alias this to something shorter?  ChatClientOptions?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+      get
+    }
+    public var requestString: Swift.String {
+      get
+    }
+    public init(_ val: Swift.String)
+  }
+  public init(apiVersion: AzureCommunicationChat.AzureCommunicationChatClientOptions.ApiVersion = .latest, logger: AzureCore.ClientLogger = ClientLoggers.default(tag: "AzureCommunicationChat"), telemetryOptions: AzureCore.TelemetryOptions = TelemetryOptions(), transportOptions: AzureCore.TransportOptions? = nil, dispatchQueue: Dispatch.DispatchQueue? = nil)
+}
+public struct ChatThreadItem : Swift.Codable {
```
tjprescott, line 229: It's confusing from the documentation what the difference between `ChatThreadItem` and `ChatThreadProperties` is. Can you explain?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public let id: Swift.String
+  public let topic: Swift.String
+  public let deletedOn: AzureCore.Iso8601Date?
+  public let lastMessageReceivedOn: AzureCore.Iso8601Date?
+  public init(id: Swift.String, topic: Swift.String, deletedOn: AzureCore.Iso8601Date? = nil, lastMessageReceivedOn: AzureCore.Iso8601Date? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct PhoneNumberIdentifierModel : Swift.Codable {
```
tjprescott, line 238: Is there a reason for "model" at the end here? Seems like it doesn't add anything. Ditto for the other `FooIdentifierModel`s.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+  public init(clientRequestId: Swift.String? = nil, cancellationToken: AzureCore.CancellationToken? = nil, dispatchQueue: Dispatch.DispatchQueue? = nil, context: AzureCore.PipelineContext? = nil)
+}
+public struct GetChatMessageOptions : AzureCore.RequestOptions {
+  public let clientRequestId: Swift.String?
+  public let cancellationToken: AzureCore.CancellationToken?
+  public var dispatchQueue: Dispatch.DispatchQueue?
+  public var context: AzureCore.PipelineContext?
+  public init(clientRequestId: Swift.String? = nil, cancellationToken: AzureCore.CancellationToken? = nil, dispatchQueue: Dispatch.DispatchQueue? = nil, context: AzureCore.PipelineContext? = nil)
+}
+public struct SendChatReadReceiptOptions : AzureCore.RequestOptions {
```
tjprescott, line 442: Should this be consistent with `SendReadReceiptRequest`, i.e. `SendReadReceiptOptions`?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
...
+public struct ListChatThreadsOptions : AzureCore.RequestOptions {
+  public let maxPageSize: Swift.Int32?
+  public let startTime: AzureCore.Iso8601Date?
+  public let clientRequestId: Swift.String?
+  public let cancellationToken: AzureCore.CancellationToken?
+  public var dispatchQueue: Dispatch.DispatchQueue?
+  public var context: AzureCore.PipelineContext?
+  public init(maxPageSize: Swift.Int32? = nil, startTime: AzureCore.Iso8601Date? = nil, clientRequestId: Swift.String? = nil, cancellationToken: AzureCore.CancellationToken? = nil, dispatchQueue: Dispatch.DispatchQueue? = nil, context: AzureCore.PipelineContext? = nil)
+}
+public struct RemoveChatParticipantOptions : AzureCore.RequestOptions {
```
tjprescott, line 45: Why is there no `RemoveChatParticipantsRequest`?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,511 @@
+// swift-interface-format-version: 1.0
```
tjprescott, line 1: Generally, it seems curious that each endpoint inconsistently features the following:
`FooOptions`
`FooRequest`
`FooResult`

Options always exists, but the request exists infrequently and response even less frequently. You might want to look at this inconsistency and question whether it is what you want.

tjprescott (commented):
Comments and questions.

examples/x86_64.swiftinterface
knvsl: When adding participants we send ChatParticipant model(s), for removing we just send the string id of the participant

examples/x86_64.swiftinterface
knvsl: Looks like this needs to be updated with Leo's change. I'm going to address your other comments first

examples/x86_64.swiftinterface
juancamilor: ChatThreadProperties has all the properties of the thread and may grow. ChatThreadItem is a subset for enumerating all threads for a user, and it includes when the last message was received on that particular thread. Similar but different.

examples/x86_64.swiftinterface
JoshuaLai: Yah that is correct. It was originally in common and it has been named that consistently that way with the other language platforms.

examples/x86_64.swiftinterface
JoshuaLai: No I don't think we need both. Are these static vars being used at all anywhere?

knvsl: they were being referenced inside the IdentifierSerializer but I just updated it to use the generated enum instead

examples/x86_64.swiftinterface
knvsl: @JoshuaLai I think we could make it internal the SDK already handles using the serializer to do the conversions - trying to think is there scenario though where dev would want to use it themselves? If they're already using the SDK they shouldn't need it right?

examples/x86_64.swiftinterface
knvsl: ChatClientOptions makes sense, is this a small change on autorest side?

knvsl:
Closing & opening the updated PR to address these comments

examples/x86_64.swiftinterface
tjprescott: If the SDK uses objects with identifiers from AzureCommunicationCommon, then why do these `FooModel` types need to be publicly exposed? It sounds like they are internally generated and could be removed from the public API.

PR #855, opened 2021-05-12
Stubbing our release stages

mitchdenny:
This PR stubs out the release stages for the iOS pipelines. We'll need to place the CocoaPods trunk publishing logic and SPM mirroring logic in here, but this is to just get the environments/approval gates configured.


eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -0,0 +1,78 @@
...
+    type: object
+    default: []
+
+stages:
+
+  # We generate two interdepdent stages for each artifact listed in the ci.yml file, creates the release
+  # in GitHub. The Release stage publishes to Maven Central. Both stages require approval since they
+  # effectively burn the version number. For testing of packages prior to burning the version number -
+  # the Validation step below publishes a package to a "burner" feed which is cleaned up after the
+  # pipeline completes.
```
tjprescott, line 16:
```suggestion
  # We generate two interdepdent stages for each artifact listed in the ci.yml file, creates the release
  # in GitHub. The Release stage publishes to CocoaPods trunk. Both stages require approval since they
  # effectively burn the version number. For testing of packages prior to burning the version number -
  # the Validation step below publishes a package to a "burner" feed which is cleaned up after the
  # pipeline completes.
```

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -0,0 +1,78 @@
...
+                deploy:
+                  steps:
+                    - checkout: self
+                    - template: /eng/common/pipelines/templates/steps/retain-run.yml
+                    # TODO: Will need to be adapted to work for Cocoapods (we'll read versions from the Cocoapod version specs)
+                    # - template: /eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
+                    #   parameters:
+                    #     ArtifactLocation: $(Pipeline.Workspace)/${{parameters.ArtifactName}}-signed/${{artifact.groupId}}/${{artifact.name}}
+                    #     PackageRepository: Maven
+                    #     ReleaseSha: $(Build.SourceVersion)
```
tjprescott, line 44: To get a package version, you can use:
`python3 eng/scripts/version.py check AzureFoo` and it will output the version to STDOUT. We run version verify in CI to ensure that the Xcode projects, podspecs, etc are consistent for each service. 

We probably DO need to verify that a version we intend to push isn't already on CocoaPods, since in theory one could trigger a release build without incrementing the version in the Azure SDK repo.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -0,0 +1,78 @@
...
+            pool:
+              name: Azure Pipelines
+              vmImage: $(OSVmImage)
+            strategy:
+              runOnce:
+                deploy:
+                  steps:
+                    - script: |
+                        echo Pretend we are publishing to CocoaPods truck.
+
```
tjprescott, line 62: Here's the manual process we use:
https://github.com/Azure/azure-sdk-for-ios/wiki/Publishing-Cocoapods

Two potentially tricky bits:
1. Registering the CI with a `pod trunk` session
2. Retry logic in case a pushed dependency isn't yet "visible" on CocoaPods.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -0,0 +1,78 @@
...
+              - template: ../variables/globals.yml
+            pool:
+              name: Azure Pipelines
+              vmImage: $(OSVmImage)
+            strategy:
+              runOnce:
+                deploy:
+                  steps:
+                    - script: |
+                        echo Pretend we are publishing to Swift Package Manager mirrors
```
tjprescott, line 78: The command used here is:
`python3 eng/scripts/sync_repo.py AzureFoo`

Followed by a git push. We should probably decide whether we want to mirror each push to master that affects a service directory, or just sync the repos on release.

tjprescott: Of course, we also need to tag the SwiftPM-AzureFoo repo with the version and the `azure-sdk-for-ios` repo accordingly. I believe we are following the correct convention:
- SwiftPM-AzureFoo: tagged with `1.0.0-beta.x`
- azure-sdk-for-ios: tagged with `AzureFoo_1.0.0-beta.x`

mitchdenny:
@tjprescott  thanks for the suggestions on the various scripts. My intention is the merge this without those so that I can setup all the approval gates etc so I've got some circuit breakers as the automation is being built. But I will incorporate these suggestions on a subsequent PR (probably one job at a time).

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: Thanks. I will be implementing this in an upcoming PR.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: Thanks. I will be implementing this in an upcoming PR.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: Will look at this in an upcoming PR.

sdk/storage/ci.yml
```diff
@@ -23,6 +23,8 @@ pr:
     - sdk/eng/
 
 extends:
-  template: templates/stages/archetype-sdk-client.yml
+  template: ../../eng/pipelines/templates/stages/archetype-sdk-client.yml
```
weshaggard, line 26: nit:
```suggestion
  template: /eng/pipelines/templates/stages/archetype-sdk-client.yml
```

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -0,0 +1,78 @@
...
```
weshaggard, line 78: Note that these echo commands will not work work as expected on non-windows. Echo on linux needs quotes but on windows doesn't. I realize this is just temporarily but just wanted to call it out if you hit an error.

mitchdenny: Ah thanks. I'll fix that up now.

tjprescott (approved):

PR #864, opened 2021-05-25
CocoaPods push support

mitchdenny:
This PR adds CocoaPods push support. To authenticate we are making use of the ```COCOAPODS_TRUNK_TOKEN``` environment variable which we'll pull in via a variable group (and ultimately our KeyVault). At this point I don't have the token but once that is available I'll do an end-to-end test, but the error output is what I expect at this point in time.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-ios-release.yml
+++ b/eng/pipelines/templates/stages/archetype-ios-release.yml
@@ -51,7 +51,7 @@ stages:
               - template: ../variables/globals.yml
             pool:
               name: Azure Pipelines
-              vmImage: $(OSVmImage)
+              vmImage: macOS-10.15
             strategy:
               runOnce:
                 deploy:
--- a/eng/pipelines/templates/stages/archetype-ios-release.yml
+++ b/eng/pipelines/templates/stages/archetype-ios-release.yml
@@ -35,12 +35,12 @@ stages:
                 deploy:
                   steps:
                     - checkout: self
-                    - template: /eng/common/pipelines/templates/steps/retain-run.yml
-                    - template: /eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
-                      parameters:
-                        ArtifactLocation: $(System.DefaultWorkingDirectory)/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}
-                        PackageRepository: CocoaPods
-                        ReleaseSha: $(Build.SourceVersion)
+                    # - template: /eng/common/pipelines/templates/steps/retain-run.yml
+                    # - template: /eng/common/pipelines/templates/steps/create-tags-and-git-release.yml
+                    #   parameters:
+                    #     ArtifactLocation: $(System.DefaultWorkingDirectory)/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}
+                    #     PackageRepository: CocoaPods
+                    #     ReleaseSha: $(Build.SourceVersion)
 
           - deployment: PublishPackageToCocoaPodsTrunk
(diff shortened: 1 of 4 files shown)


eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -51,13 +51,17 @@ stages:
...
               name: Azure Pipelines
-              vmImage: $(OSVmImage)
+              vmImage: macOS-10.15
             strategy:
               runOnce:
                 deploy:
                   steps:
                     - script: |
-                        echo "Pretend we are publishing to CocoaPods trunk.""
+                        pod trunk me
```
weshaggard, line 60: I'm not familiar with the pod tool but is it generally available on mac machines by default?

tjprescott: No, you need to install it with a ruby command.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -51,13 +51,17 @@ stages:
...
+                        pod trunk push $(System.DefaultWorkingDirectory)/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}/${{artifact.name}}.podspec.json --synchronous --use-modular-headers
```
tjprescott, line 61: Are we checking elsewhere that the podspec in question has not already been published?

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: It appears to be there by default.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: Not explicitly.

mitchdenny:
@tjprescott @weshaggard publishing to CocoaPods is working:

image

eng/pipelines/templates/stages/archetype-ios-release.yml
tjprescott: I don't know what the sequence for tagging/pushing to trunk is, but we should probably do that lest we tag a release and then find out we need additional commits because the version was not bumped. Then again, I suppose we could check that the tag doesn't already exist.

weshaggard (approved):
Looks reasonable to me.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -51,13 +51,18 @@ stages:
               - template: ../variables/globals.yml
```
weshaggard, line 51: FYI @tjprescott based on the release changes I've opened https://github.com/Azure/azure-sdk/issues/2894 and assigned to you to also update the guidelines and automation based on all these changes.

weshaggard (approved):
Looks reasonable to me.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: The tag release process has a mechanism which triggers a failure if you try to tag a release again using a different commit sha (this is consistent across the engineering system). As part of that mechanism we actually check to see if that particular package has also been published (but its not quite explicitly what you are asking).

If we tag and fail to publish, and have to add more commits we have control of our GitHub releases and can delete the release/tag and then run the release process again (we do it all the time). Once things hit something we don't control (e.g. CocoaPods) then the version number is burnt.

For this reason, when we get around to doing the SPM sync (next) it might pay to do that before the CocoaPods publish.

PR #875, opened 2021-06-01
Fix build issues with enum CodingKeys

tjprescott:
Fixes issue with `StorageError` and `MappedThing` where Xcode can no longer automatically synthesize these coding keys.

Also cleans up some other project files.


jairmyree (approved):
Everything looks good :)

PR #876, opened 2021-06-02
Add SPM publishing to release pipeline

mitchdenny:
An initial pass at publishing SPM packages via the release pipeline. It requires that we add a repository resource (or two) to each individual pipeline. Then during this job we clone the mono-repo, and the ```SwiftPM-[packagename]``` repo. We then checkout an orphan branch and copy the files over, commit, tag and then push (tag only).

In cases where a package doesn't support SPM (e.g. ```AzureCommunicationCalling```), we just don't emit that job.

### DIFF
--- a/eng/pipelines/templates/stages/archetype-ios-release.yml
+++ b/eng/pipelines/templates/stages/archetype-ios-release.yml
@@ -81,5 +81,5 @@ stages:
                     - checkout: self
                     - checkout: SwiftPM-AzureTemplate
                     - pwsh: |
-                        Get-ChildItem $(Pipeline.Workspace)
+                        Get-ChildItem $(Pipeline.Workspace)\s
                     
--- a/eng/pipelines/templates/stages/archetype-ios-release.yml
+++ b/eng/pipelines/templates/stages/archetype-ios-release.yml
@@ -85,7 +85,7 @@ stages:
                         Get-ChildItem $(Pipeline.Workspace)/s
                         Push-Location -Path $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}}
                         git checkout --orphan run-$(Build.BuildID)
-                        Copy-Item -Path $(Pipeline.Workspace)/s/azure-sdk-for-ios/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}} -Destination $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}} -Recurse
+                        Copy-Item -Path $(Pipeline.Workspace)/s/azure-sdk-for-ios/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}/* -Destination $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}} -Recurse
                         Get-ChildItem -Path $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}} -Recurse
                       displayName: Copy and tag sources
 
--- a/eng/pipelines/templates/stages/archetype-ios-release.yml
+++ b/eng/pipelines/templates/stages/archetype-ios-release.yml
@@ -66,7 +66,7 @@ stages:
 
           - deployment: PublishPackageToSwiftPackageManagerMirrors
             displayName: "Publish to SPM Mirrors"
-            condition: and(succeeded(), ne(variables['Skip.PublishPackage'], 'true'))
+            condition: and(succeeded(), ne(variables['Skip.PublishPackage'], 'true'), ne('${{artifact.skipSwiftPackageManager}}', true))
             environment: cocoapodstrunk
             dependsOn: TagRepository
             variables:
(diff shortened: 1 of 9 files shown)


sdk/storage/ci.yml
```diff
@@ -22,6 +22,13 @@ pr:
     - sdk/storage/
     - sdk/eng/
 
+resources:
+  repositories:
+  - repository: SwiftPM-AzureStorageBlob
```
tjprescott, line 27: I don't think these exist yet since these aren't even in preview release, in case that matters.

sdk/template/ci.yml
```diff
@@ -22,6 +22,13 @@ pr:
     - sdk/template/
     - sdk/eng/
 
+resources:
+  repositories:
+  - repository: SwiftPM-AzureTemplate
+    type: github
+    name: Azure/SwiftPM-AzureTemplate
+    endpoint: Azure
+
```
tjprescott, line 31: Out of curiosity, what happens if a new SDK is added to the service directory but NOT to this resources section? Does it fail silently, or cause an error?

tjprescott (commented):
I'm not sure I fully understand the git magic, but if it works I'm happy. Have a couple comments/questions.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -66,17 +66,32 @@ stages:
...
               runOnce:
                 deploy:
                   steps:
-                    - script: |
-                        echo "Pretend we are publishing to Swift Package Manager mirrors"
+                    - checkout: self
+                    - checkout: SwiftPM-${{artifact.name}}
+                      persistCredentials: true
+                    - pwsh: |
+                        $podspec = Get-Content -Raw -Path $(Pipeline.Workspace)/s/azure-sdk-for-ios/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}/${{artifact.name}}.podspec.json | ConvertFrom-Json
```
weshaggard, line 85: There are a lot of steps here which might be difficult to debug if something goes wrong. We should add some logging and perhaps refactoring some of this into a script.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -66,17 +66,32 @@ stages:
...
+                        $version = $podspec.version
+                        Push-Location -Path $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}}
+                        git checkout --orphan run-$(Build.BuildID)-temp-branch
+                        git reset --hard
```
weshaggard, line 89: We should explicitly put a ref to reset here instead of relying on defaults.

eng/pipelines/templates/stages/archetype-ios-release.yml
```diff
@@ -66,17 +66,32 @@ stages:
...
+                        Copy-Item -Path $(Pipeline.Workspace)/s/azure-sdk-for-ios/sdk/${{parameters.ServiceDirectory}}/${{artifact.name}}/* -Destination $(Pipeline.Workspace)/s/SwiftPM-${{artifact.name}} -Recurse
+                        git add .
```
weshaggard, line 91: We should try  to use our existing templates https://github.com/Azure/azure-sdk-tools/blob/master/eng/common/pipelines/templates/steps/git-push-changes.yml.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: I was debating a script on this one, I think you are right, I'll extract it into a script.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: The preceding command basically creates a new ref pointing to an empty state. The reset is just an easy way of getting rid of the files left over from master/main.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: I think that these are very different scenarios. In this case I'm pushing the tag, not the branch, I'm not sure what this script will provide to this code, but adding my scenario to the script would probably just introduce a lot of unnecessary complexity.

sdk/storage/ci.yml
mitchdenny: I'll remove them and put a skip in for the SPM package publish job.

sdk/template/ci.yml
mitchdenny: When you go to trigger the release pipeline it would fail because it will try to checkout a repo that doesn't exist. To handle publishing packages to CocoaPods and not SPM, you can add a skip variable, and in those cases the absence of the resources section won't be a problem because in the final emitted pipeline it is not used.

eng/pipelines/templates/stages/archetype-ios-release.yml
weshaggard: Fair if you are only pushing the tag then I agree we probably don't need to add that complexity to that script. If we are simply pushing a branch and then tagging separately then we could consider using the common stuff but I agree this is a different scenario.

mitchdenny: Yeah the branch never goes to the server, I'm only using it as a way of creating an orphan parent commit.

eng/pipelines/templates/stages/archetype-ios-release.yml
mitchdenny: Its now in a script.

sdk/template/ci.yml
tjprescott: Perfect. The silent fail would be the scenario I feared.

tjprescott (commented):
LGTM. Will defer to @weshaggard for approval

weshaggard (approved):
Looks reasonable.

PR #877, opened 2021-06-02
Update test schemes

tjprescott:
Update to ensure that the various schemes run all of the correct tests.


jairmyree (approved):
Everything looks good

PR #880, opened 2021-06-03
Streamline Chat live testing

tjprescott:
This PR commits a script to aid in setting up AzureCommunicationChat tests for re-recording.

### DIFF
--- a/eng/scripts/prepare_chat_tests.py
+++ b/eng/scripts/prepare_chat_tests.py
@@ -1,0 +1,40 @@
+#!/usr/bin/python3
+
+"""
+prepare_chat_tests.py
+
+1) Run `pip install azure.communication.administration` prior to running.
+
+2) Set the `AZURE_COMMUNICATION_CS` environment variable
+
+3) Run the script.
+
+4) Copy the values to the AzureCommunicationChat scheme's testing environment variables and set `TEST_MODE` to
+   "record".
+"""
+
+import os
+from azure.communication.administration import CommunicationIdentityClient
+
+connection_string = os.environ['AZURE_COMMUNICATION_CS']
+identity_client = CommunicationIdentityClient.from_connection_string(connection_string)
+
+items = { key:val for (key, val) in (x.split('=', 1) for x in connection_string.split(';')) }
+endpoint = items['endpoint']
+
+user1 = identity_client.create_user()
+user2 = identity_client.create_user()
+
+token2 = identity_client.issue_token(user2, scopes=["chat"]).token
+
+print('\n== AZURE_COMMUNICATION_ENDPOINT ==')
+print(endpoint)
+
+print('\n== AZURE_COMMUNICATION_USER_ID_1 ==')
+print(user1.identifier)
+
+print('\n== AZURE_COMMUNICATION_USER_ID_2 ==')
+print(user2.identifier)
+
+print('\n== AZURE_COMMUNICATION_TOKEN ==')
+print(token2)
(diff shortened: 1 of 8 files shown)


eng/scripts/prepare_chat_tests.py
```diff
@@ -0,0 +1,40 @@
...
+print(endpoint)
+
+print('\n== AZURE_COMMUNICATION_USER_ID_1 ==')
+print(user1.identifier)
+
+print('\n== AZURE_COMMUNICATION_USER_ID_2 ==')
+print(user2.identifier)
+
+print('\n== AZURE_COMMUNICATION_TOKEN ==')
+print(token2)
```
knvsl, line 40: I think we should return the token for user1 instead of user2

In the tests the initial user (whose token we use to create the ChatClient) is meant to be user1, and for adding participants we use user2. If we provide the token for user2 it means the initial user is actually user2 & the add participants tests need to be swapped to add user1 instead

sdk/communication/AzureCommunicationChat/Tests/ChatClientTests.swift
```diff
@@ -33,19 +33,19 @@ class ChatClientTests: XCTestCase {
     /// ChatClient initialized in setup.
     private var chatClient: ChatClient!
     /// Test mode.
-    private var mode = ProcessInfo.processInfo.environment["TEST_MODE"] ?? "playback"
+    private var mode = getEnvironmentVariable(withKey: "TEST_MODE", default: "playback")
```
knvsl, line 36: We could define the environment variable names in one place like an enum or struct instead of hardcoding them multiple times as I did, since only certain ones are valid anyways

eng/scripts/prepare_chat_tests.py
tjprescott: I used token2 because I didn't actually see any Chat tests that used USER_ID_1. I'll update it.

sdk/communication/AzureCommunicationChat/Tests/ChatClientTests.swift
tjprescott: True. I made this change mainly because I wanted the environment variable names to remain in the scheme settings, which meant the previous way would return empty string instead of nil. As we look for ways to make this more generic, we would ideally be able to entirely do away with environment variables.

eng/scripts/prepare_chat_tests.py
tjprescott: Updated to use token1.

knvsl (approved):

eng/scripts/prepare_chat_tests.py
```diff
@@ -0,0 +1,40 @@
...
+connection_string = os.environ['AZURE_COMMUNICATION_CS']
+identity_client = CommunicationIdentityClient.from_connection_string(connection_string)
+
+items = { key:val for (key, val) in (x.split('=', 1) for x in connection_string.split(';')) }
+endpoint = items['endpoint']
+
+user1 = identity_client.create_user()
+user2 = identity_client.create_user()
+
+token2 = identity_client.issue_token(user2, scopes=["chat"]).token
```
knvsl, line 28: Just noticed should be updated to token1 & user1 here too:
```
token1 = identity_client.issue_token(user1, scopes=["chat"]).token
```

tjprescott: Thanks for that catch. I updated and reran the script to make sure it all works as expected.

jairmyree (approved):
Everything looks good :)

PR #888, opened 2021-06-04
[Communication] - Common - Introduce new error object local to common

JoshuaLai:
Removing the Core dependency and having local copy of error


tjprescott (commented):
This LGTM. One of the things we mentioned was beefing up the unit test that currently just ensures an error was thrown to do more inspection of that error. The goal would be to ensure the new NSError thrown would be seen as functionally identical (i.e. passes the old test without modification). Did we do that?

JoshuaLai:
@tjprescott Yep I updated the test a bit but it looks like on the swift side its a bit harder to get the user info in the NSError. `localizedDescription` just outputs: `// The operation couldn’t be completed. (AzureCommunicationCommon.JwtTokenParser.decodeJwtPayload error 0.)` so I added a string contains.

JoshuaLai:
Oh actually improved on it a bit more. In swift unit test casted it to an NSError for testing to check against the userInfo. I should update the documentation to reflect this change as well.

tjprescott (approved):

jairmyree (approved):

tjprescott:
Now the only question is why the smoketest is not using your PR's changes... ?

JoshuaLai:
> Now the only question is why the smoketest is not using your PR's changes... ?

I am trying to run the smoke test locally but I am running into some issues with trouter

JoshuaLai:
closing this PR 
https://github.com/Azure/azure-sdk-for-ios/pull/905 has my changes included

Issue #897, opened 2021-06-10
RPF: Scrub information from recordings


tjprescott:
Based on the content in https://github.com/Azure/azure-sdk-for-python/blob/master/tools/azure-devtools/src/azure_devtools/scenario_tests/recording_processors.py:

**General Sanitizing** (PR #932)
- [x] Subscription ID 
  - [x] Request URI
  - [x] Request body
  - [x] Response body
  - [x] Response location header
  - [x] Response azure-asyncoperation header
  - [x] Response URL
- [x] Location header
- [x] Operation-location header
- [x] Azure-asyncoperation header
- [x] www-authenticate header

**Special Situations**
1. OAuth Requests and Responses should not be recorded
2. Authentication metadata requests should be dropped
   - Request URI contains "/.well-known/openid-configuration" or "/common/discovery/instance"
3. Large request bodies (truncate)
4. Large response bodies (truncate)

**Not Applicable for Mobile**
- Deployment names (replace random name with fixed mock name)
- Access tokens for service principal auth in response body
- Replace randomized live names with deterministic names

tjprescott:
As part of this, we will need to see if DVR supports a hook into the request and response before saving. If not, we may need to create some.

tjprescott:
This issue has been completed.

PR #907, opened 2021-06-14
Feature/recording location

### LINKED ISSUE #901: RPF: Save recordings in code path instead of simulator
The current default behavior in DVR is to save the recording files to the simulator storage. The desired behavior would be to have the recordings saved to a `recordings` folder in the appropriate test directory. An acceptable workaround would be to move the files from the simulator to the appropriate directory, but hopefully DVR provides the capability to specify this.

jairmyree:
This PR places recordings inside of the recording folder in chat instead of in the simulator. 
Closes #901

### DIFF
--- a/sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
+++ b/sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
@@ -530,9 +530,8 @@
 				D1CD93E7258D6F0600409613 /* updateMessage.json */,
 				D1CD93EB258D6F1300409613 /* updateTopic.json */,
 			);
-			name = Recordings;
-			path = "/Users/jairmyree/Desktop/azure-sdk-for-ios/sdk/communication/AzureCommunicationChat/Tests/Util/Recordings";
-			sourceTree = "<absolute>";
+			path = Recordings;
+			sourceTree = "<group>";
 		};
 		OBJ_113 /* Products */ = {
 			isa = PBXGroup;
--- a/sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
+++ b/sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
@@ -530,8 +530,9 @@
 				D1CD93E7258D6F0600409613 /* updateMessage.json */,
 				D1CD93EB258D6F1300409613 /* updateTopic.json */,
 			);
-			path = Recordings;
-			sourceTree = "<group>";
+			name = Recordings;
+			path = "/Users/jairmyree/Desktop/azure-sdk-for-ios/sdk/communication/AzureCommunicationChat/Tests/Util/Recordings";
+			sourceTree = "<absolute>";
 		};
 		OBJ_113 /* Products */ = {
 			isa = PBXGroup;
--- a/sdk/test/AzureTest/DVRSessionTransport.swift
+++ b/sdk/test/AzureTest/DVRSessionTransport.swift
@@ -59,8 +59,9 @@ public class DVRSessionTransport: TransportStage {
     public func open() {
         guard session == nil else { return }
         let sdkPath = environmentVariable(forKey: "SDK_REPO_ROOT", default: "~")
-        
-        let outputDirectory = sdkPath + ((sdkPath.last == "/") ? "sdk/communication/AzureCommunicationChat/Tests/Recordings" : "/sdk/communication/AzureCommunicationChat/Tests/Recordings")
+        guard let outputDirectory = URL(string: sdkPath)?.appendingPathComponent("sdk/communication/AzureCommunicationChat/Tests/Recordings").absoluteString else {
+            fatalError("SDK Path Invalid")
+        }
         session = Session(outputDirectory: outputDirectory, cassetteName: cassetteName)
         if environmentVariable(forKey: "TEST_MODE", default: "playback") == "record" {
             session?.recordingEnabled = true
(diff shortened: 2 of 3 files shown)


eng/common/scripts/ChangeLog-Operations.ps1
```diff
@@ -37,12 +37,14 @@ function Get-ChangeLogEntriesFromContent {
     return $null
```
tjprescott, line 37: I don't think this change should be in your PR.

eng/common/scripts/Verify-ChangeLog.ps1
```diff
@@ -6,11 +6,12 @@ param (
   [string]$ServiceDirectory,
```
tjprescott, line 6: Ditto here. I think you can remove these two files from your PR.

sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
```diff
@@ -530,8 +536,9 @@
 				D1CD93E7258D6F0600409613 /* updateMessage.json */,
 				D1CD93EB258D6F1300409613 /* updateTopic.json */,
 			);
-			path = Recordings;
-			sourceTree = "<group>";
+			name = Recordings;
+			path = "/Users/jairmyree/Desktop/azure-sdk-for-ios/sdk/communication/AzureCommunicationChat/Tests/Util/Recordings";
```
tjprescott, line 540: Is there a way to make this path relative? I suspect it won't work for me when I pull this change.

sdk/communication/AzureCommunicationChat/Tests/Util/Recordings/test_CreateThread_WithoutParticipants.json
```diff
@@ -0,0 +1,52 @@
+{
```
tjprescott, line 1: We need to remove the recording files until #897 is complete. This contains sensitive info.

sdk/test/AzureTest/DVRSessionTransport.swift
```diff
@@ -58,7 +58,9 @@ public class DVRSessionTransport: TransportStage {
 
     public func open() {
         guard session == nil else { return }
-        session = Session(cassetteName: cassetteName)
+        let sdkPath = environmentVariable(forKey: "SDK_REPO_ROOT", default: "~")
+        let outputDirectory = sdkPath + ((sdkPath.last == "/") ? "sdk/communication/AzureCommunicationChat/Tests/Util/Recordings" : "/sdk/communication/AzureCommunicationChat/Tests/Util/Recordings")
```
tjprescott, line 62: `<MOD>/Tests/Recordings` is probably a sufficient location.

tjprescott (changes requested):

sdk/communication/AzureCommunicationChat/Tests/Util/Recordings/test_CreateThread_WithoutParticipants.json
jairmyree: I've removed the JSON recording files

sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
jairmyree: I believe that's possible, but this was not meant to be commited

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: Also, instead of this, you should be able to use `sdkPath.appendingPathComponent` and not have to do this special checking of separators.

sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/project.pbxproj
tjprescott: You can revert this file and I can see what happens on my end

eng/common/scripts/Verify-ChangeLog.ps1
jairmyree: This is also from commits to master. Should I still remove it from the pr?

sdk/test/AzureTest/DVRSessionTransport.swift
jairmyree: When you say <MOD>, do you mean the location pointed to by the env variable?

jairmyree (commented):
@tjprescott please respond when you get a chance.

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: Well, it would be:
`<SDK_PATH>/sdk/communication/<MODULE>/Tests/Recordings`

In other words, recordings don't need to be in Util. :)

jairmyree: sdkPath is a string, so I can't use that function. I did not convert it into a URL because outputDirectory takes a string and the environment variable gives out a string as well, I thought it'd be more tedious to convert to a URL just to go back to a string.

jairmyree (commented):
@tjprescott When you get a chance please take another look

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: It's better to go to URL and manipulate that way and then go back to string. Duplicating the path is a red flag. If somewhere else you forget to do this, you have bugs. If you temporarily convert to URL, manipulate and then go back, you are good. 
```swift
guard let outputDirectory = URL(sdkPath)?.appendingPathComponent("sdk/blah/")?.absoluteString else { error("SDK path invalid") }
```

tjprescott (approved):

PR #909, opened 2021-06-15
Env Variable TEST_MODE works as expected

### LINKED ISSUE #906: RPF: Configure DVR to re-record existing tests
DVR concurrently looks to see if a recording exists, and if it does, it will consider itself in playback mode. We would prefer to explicitly call out whether we want DVR to record.

Modes:
- record: record tests or, if they already exist, re-record them
- playback: playback tests if they exist; if not, fail

This should honor the TEST_MODE environment variable we are setting.

jairmyree:
Enforce that DVR record-playback honors environment variables
Live will ignore recordings and always run live
Record will ignore recordings, run live, and create a new recording
Playback will only play recordings and fail if the recording isn't valid or does not exist.
Closes #906

### DIFF
--- a/sdk/test/AzureTest/DVRSessionTransport.swift
+++ b/sdk/test/AzureTest/DVRSessionTransport.swift
@@ -58,7 +58,11 @@ public class DVRSessionTransport: TransportStage {
 
     public func open() {
         guard session == nil else { return }
-        session = Session(cassetteName: cassetteName)
+        let sdkPath = environmentVariable(forKey: "SDK_REPO_ROOT", default: "~")
+        guard let outputDirectory = URL(string: sdkPath)?.appendingPathComponent("sdk/communication/AzureCommunicationChat/Tests/Recordings").absoluteString else {
+            fatalError("SDK Path Invalid")
+        }
+        session = Session(outputDirectory: outputDirectory, cassetteName: cassetteName)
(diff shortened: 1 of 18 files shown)


tjprescott (commented):
This looks good. I'm assuming it depends on a yet-to-be-opened PR against DVR?

tjprescott:
Closes #906.

tjprescott:
Depends on: https://github.com/tjprescott/DVR/pull/1

jairmyree (commented):
@tjprescott Please take a look when you get a chance.

sdk/test/AzureTest/DVRSessionTransport.swift
```diff
@@ -64,8 +64,14 @@ public class DVRSessionTransport: TransportStage {
         }
         session = Session(outputDirectory: outputDirectory, cassetteName: cassetteName)
         if environmentVariable(forKey: "TEST_MODE", default: "playback") == "record" {
+            session?.recordMode = .newEpisodes
```
tjprescott, line 67: If in record mode, this should be `.all`. I don't think we'll use `.once` or `.newEpisodes` at all.

tjprescott (changes requested):

tjprescott (approved):

Issue #882, opened 2021-06-03
Chat Record/Playback Enhancements


tjprescott:
- The test setup should construct a resource group and the resources needed to support the test (in this case, an ACS resource, a couple users, and a token). Currently that must be done manually using a script. The test should automatically delete that resource group when the test completes. Complicating this is the fact that there is no management plane SDK for iOS. See issue #892.
- The scrubbing mechanism needs to be more robust.
- The need to manually register stubs is problematic and results in a lot of boilerplate code. Ideally, all of that metadata like status code, path, etc should be contained in the recording, and all of that can be registered by simply processing the recordings.

knvsl:
Covered most of it above, just following up notes from the meeting:
- We need to keep the ability to scrub out sensitive data from the recorded responses
- Ideally we remove the boilerplate code - registering stubs & also recording responses, so devs writing tests don't have to think about it
- Avoid conflicts between tests, ie GET message vs GET message on some updated message - same call but different tests might want different responses, avoid conflict between unit test mocks and live test recordings

tjprescott:
With the recent merging of AzureTest, these broad goals have been achieved. We can create specific issues for follow-on work.

PR #912, opened 2021-06-17
Automated means of creating test resources

### LINKED ISSUE #911: AzureTest: data-plane operations
There are certain operations supported by the service that, for whatever reason, aren't made available via the iOS SDK. For testing, we may need to invoke these operations in order to set up the test. We should investigate the feasibility of supporting something like Python's low-level client to facilitate making these calls.

tjprescott:
Closes #911. Partially addresses #882.

Leverages existing Engineering System work to create resources with `New-TestResources.ps1 communication`. It's not pretty, but it works. 

**old way**
1. Create resources manually via portal. Copy connection string and export to AZURE_COMMUNICATION_CS
2. Run `python eng/scripts/prepare_chat_tests.py`
3. Copy output of Python script to environment variables in XCode

**new way**
1. With Powershell (🤕 ) run: `eng/common/TestResources/New-TestResources.ps1 communication`
2. Copy output of Python script to environment variables in XCode

So there is still a very onerous step of linking the script output to the test, but I will focus on that in a different PR. Also, I suspect changes will need to be made to actually run this on a live test pipeline, but it works locally (for me).


eng/scripts/prepare_chat_tests.py
```diff
@@ -5,36 +5,31 @@
...
-print(user1.identifier)
-
-print('\n== AZURE_COMMUNICATION_USER_ID_2 ==')
-print(user2.identifier)
+results = {
+   'endpoint': endpoint,
+   'user1': user1.identifier,
+   'user2': user2.identifier,
+   'token': token
+}
```
tjprescott, line 33: I originally did this because Azure deploymentScripts required the output to be JSON... now it doesn't matter. The original might be preferable.

jairmyree (approved):
Is the use of a local repo for DVR in the Podfile meant to be temporary?

eng/scripts/prepare_chat_tests.py
tjprescott: Reverted this.

PR #924, opened 2021-06-23
Injectable test settings 

### LINKED ISSUE #908: AzureTest: Replace Environment Variable with Configuration File
Currently we are using environment variables defined in a particular test scheme to control the recording mechanism. This is problematic since one could accidentally commit real values for these variables, which we do _not_ want. Instead, we should load a configuration file and access necessary settings to control the recording in that way.

tjprescott:
Closes #908. This PR eliminates environment variables in the scheme files and populates a test-settings.plist (which is part of the gitignore) to facilitate recording and playing back tests.


jairmyree (approved):
I would like to talk you during our sync-up about something I noticed in the Podfile. It's not directly related to this PR but it did provoke a few questions. :)

PR #932, opened 2021-06-25
Feature/scrub subscription ids

jairmyree:
This PR is to provide a filter that properly scrubs the subscription IDs and sensitive headers from end-to-end test records.


sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,66 @@
...
+// --------------------------------------------------------------------------
+
+import XCTest
+import Foundation
+import DVR
+import AzureTest
+
+class AzureTestTests: XCTestCase {
+
+    let fakeData = try! String(contentsOfFile: "/Users/jairmyree/Downloads/output-onlineyamltools.json")
```
tjprescott, line 34: You won't be able to use a hard-coded string like this. You'll need to include it as a test resource.

sdk/test/AzureTest/DVRSessionTransport.swift
```diff
@@ -64,6 +64,9 @@ public class DVRSessionTransport: TransportStage {
             .appendingPathComponent("sdk/communication/AzureCommunicationChat/Tests/Recordings").absoluteString else {
             fatalError("SDK Path Invalid")
         }
+        let replacements = ["authorization": "", "client-request-id" : "", "retry-after" : "","x-ms-client-request-id": "","x-ms-correlation-request-id": "", "x-ms-ratelimit-remaining-subscription-reads": "",
+                            "x-ms-request-id":"","x-ms-routing-request-id":"","x-ms-gateway-service-instanceid":"","x-ms-ratelimit-remaining-tenant-reads":"","x-ms-served-by":"","x-ms-authorization-auxiliary":""]
+
```
tjprescott, line 69: So in python, this would simply be a list of keys that would be passed in as a header filter and handled by DVR.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+public class Scrubbing {
```
tjprescott, line 29: This probably doesn't need to be a class. You can just put helper methods in here:
```swift
func scrubSubscriptionIDs(from: URLRequest) -> URLRequest {
  let regex = ...
  // rest of logic
}
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        let dirtyHeaders = request.allHTTPHeaderFields ?? [:]
+        var cleanHeaders : [String:String] = [:]
+        for key in dirtyHeaders.keys {
+            cleanHeaders[key] = regexSubscriptionReplace(dirtyHeaders[key])
+        }
+        if cleanHeaders.isEmpty {
+            cleanRequest.allHTTPHeaderFields = nil
+        } else {
+            cleanRequest.allHTTPHeaderFields = cleanHeaders
+        }
```
tjprescott, line 48:
```suggestion
        cleanRequest.allHTTPHeaderFields = cleanHeaders.isEmpty ? nil : cleanHeaders
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+    
+    static func scrubSubscriptionIDs(request: URLRequest) -> URLRequest {
+        
+        var cleanRequest = request
+        
+        let dirtyHeaders = request.allHTTPHeaderFields ?? [:]
+        var cleanHeaders : [String:String] = [:]
+        for key in dirtyHeaders.keys {
+            cleanHeaders[key] = regexSubscriptionReplace(dirtyHeaders[key])
+        }
```
tjprescott, line 43:
```suggestion
    let dirtyHeaders = request.allHTTPHeaderFields ?? [:]
    let cleanHeaders = dirtyHeaders.map { key, val in
        (key, regexSubscriptionReplace(val))
    }
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+    func stringForKey(_ key: Key) throws -> String {
+        if let string = self[key] as? String {
+            return string
+        }
+        throw "Not a string"
+    }
+}
+
+extension String {
+    public func dictionaryFromString() throws -> Dictionary<String, Any> {
```
tjprescott, line 155: It's okay if we use this internally, but it should not be made public.
```suggestion
    func convertToDictionary() throws -> [String: Any] {
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        do {
+            let data = self.data(using: .utf8)
+            let dictionary = try JSONSerialization.jsonObject(with: data!, options: []) as? [String: Any]
+            return dictionary!
+        } catch {
+            throw "Not a dictionary"
+        }
```
tjprescott, line 162: It's a little weird to throw but also use forced-unwrapping here. If dictionary is `nil` you should also throw "Not a dictionary" instead of crashing due to the force-unwrap.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        throw "Not an array"
+    }
+    
+    func stringForKey(_ key: Key) throws -> String {
+        if let string = self[key] as? String {
+            return string
+        }
+        throw "Not a string"
+    }
```
tjprescott, line 151: These names don't conform to Swift guidelines. They should be `fooType(forKey key: Key)`. Also, it would probably make more sense to simply have these return optionals instead of throwing.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+    
+}
+
+ extension Dictionary where Key == String, Value == Any {
+    
+    static func findValue(key: String, dictionary: Dictionary<String,Any>) -> Any? {
```
tjprescott, line 116: Don't understand why this would be static.

```suggestion
    func value(forKey key: String) -> Any? {
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+       }
+        
+       for testKey in dictionary.keys {
+           if let innerDictionary = dictionary[testKey] as? Dictionary<String , Any> {
+               let returnedValue = findValue(key: key, dictionary: innerDictionary)
+               if returnedValue != nil {
+                   return returnedValue
+               }
+           }
+       }
```
tjprescott, line 128: It looks like you are recursively looking for a key and stop once the first match is found. This seems like it may be error-prone. For example, if both request and response have a uri field, and you search for uri, you'll get back only one, and the other may not be reachable at all with this logic.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        throw "Not an array"
+    }
+    
+    func stringForIndex(_ index: Int) throws -> String {
+        if let string = self[index] as? String {
+            return string
+        }
+        throw "Not a string"
+    }
```
tjprescott, line 110: Same comment as below regarding naming and return types.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+    
+    static func regexSubscriptionReplace(_ string: String?)->String?{
+        guard string != nil else {
+            return nil
+        }
+        var editedString = string!
+        var range = NSRange(location: 0, length: editedString.utf8.count)
+        if subscriptionIDRegex.numberOfMatches(in: editedString, options: [], range: range) > 0 {
+            editedString = subscriptionIDRegex.stringByReplacingMatches(in: editedString, options: [], range: range, withTemplate: "$1/99999999-9999-9999-9999-999999999999")
```
tjprescott, line 64: Is there a reason you are using all 9's instead of all 0's?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        cleanRequest.httpBody = regexSubscriptionReplace(String(data: request.httpBody, encoding: .utf8))?.data(using: .utf8)
+    
+        return cleanRequest
+    }
+    
+    static func regexSubscriptionReplace(_ string: String?)->String?{
+        guard string != nil else {
+            return nil
+        }
+        var editedString = string!
```
tjprescott, line 61:
```suggestion
        guard var editedString = string else {
            return nil
        }
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+        
+        cleanRequest.url = URL(string: regexSubscriptionReplace(request.url?.absoluteString))
+        cleanRequest.mainDocumentURL = URL(string: regexSubscriptionReplace(request.mainDocumentURL?.absoluteString))
+        cleanRequest.httpBody = regexSubscriptionReplace(String(data: request.httpBody, encoding: .utf8))?.data(using: .utf8)
+    
+        return cleanRequest
+    }
+    
+    static func regexSubscriptionReplace(_ string: String?)->String?{
```
tjprescott, line 57: This should probably not be static, should be more descriptive, and avoid using `_`:
```suggestion
    func scrubSubscriptionId(from string: String?) -> String? {
```

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        let dataDictionary = try! JSONSerialization.jsonObject(with: fakeData.data(using: .utf8)!, options: []) as! [String:Any]
+        fakeRequest = URLRequest(url: URL(string: try! dataDictionary.arrayForKey("interactions").dictionaryForIndex(0).dicionaryForKey("request").stringForKey("uri"))!)
+        fakeRequest?.httpBody = try! dataDictionary.arrayForKey("interactions").dictionaryForIndex(0).dicionaryForKey("request").stringForKey("body").data(using: .utf8)
+    }
+
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+
+    func testCleanRequestURL() throws {
```
tjprescott, line 54: Take a look at tests in other modules to see how to name this. Something like:
```swift
func test_scrubbingRequest_removesSubscriptionId
```
Or something like that. Generally we don't use underscore in Swift, except for these test names.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    var fakeResponse : URLResponse?
+    
+    var fakeResponseData : Data?
+    
+    
+    override func setUpWithError() throws {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+        let dataDictionary = try! JSONSerialization.jsonObject(with: fakeData.data(using: .utf8)!, options: []) as! [String:Any]
+        fakeRequest = URLRequest(url: URL(string: try! dataDictionary.arrayForKey("interactions").dictionaryForIndex(0).dicionaryForKey("request").stringForKey("uri"))!)
+        fakeRequest?.httpBody = try! dataDictionary.arrayForKey("interactions").dictionaryForIndex(0).dicionaryForKey("request").stringForKey("body").data(using: .utf8)
```
tjprescott, line 47: Seems to be a typo in here: dicionary -> dictionary

tjprescott (changes requested):

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: With this PR (https://github.com/tjprescott/DVR/pull/5) you'd be able to omit these with something like:
```swift
let filters = ["authorization": .remove, "client-request-id": .remove, ...]
```

For now, instead of leveraging `replacements`, I think you should just drop these headers in the `beforeRecordRequest` closure.

tjprescott:
@jairmyree you also should rebase and resolve the conflicts--otherwise the CI won't kick off.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
jairmyree: I've made the change :)

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I've made this change :)

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I've made the changes :)

sdk/test/AzureTest/Scrubbing.swift
jairmyree: These functions are currently just for testing, so I had them throw because I wanted them to break the test if I used them incorrectly.

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I've made the changes

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I made this edit

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
jairmyree: I fixed this

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: This was completed in #935.

sdk/test/AzureTest/AzureTest.xcodeproj/xcshareddata/xcschemes/AzureTest.xcscheme
```diff
@@ -20,26 +20,33 @@
...
-            value = ""
-            isEnabled = "YES">
-         </EnvironmentVariable>
-      </EnvironmentVariables>
       <Testables>
          <TestableReference
             skipped = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
-               BlueprintIdentifier = "AzureTest::AzureTestTests"
```
tjprescott, line 42: Why is this a name and your commit is a GUID?

sdk/test/AzureTest/DVRSessionTransport.swift
```diff
@@ -64,6 +64,9 @@ public class DVRSessionTransport: TransportStage {
             .appendingPathComponent("sdk/communication/AzureCommunicationChat/Tests/Recordings").absoluteString else {
             fatalError("SDK Path Invalid")
         }
+        let replacements = ["authorization": "", "client-request-id" : "", "retry-after" : "","x-ms-client-request-id": "","x-ms-correlation-request-id": "", "x-ms-ratelimit-remaining-subscription-reads": "",
+                            "x-ms-request-id":"","x-ms-routing-request-id":"","x-ms-gateway-service-instanceid":"","x-ms-ratelimit-remaining-tenant-reads":"","x-ms-served-by":"","x-ms-authorization-auxiliary":""]
```
tjprescott, line 68: This might be a bad merge, because after #932, these should read like:
```swift
"authorization": .remove,
"client-request-id": .remove
...
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+import DVR
+
+public extension Filter {
+    static let subscriptionIDRegex = try! NSRegularExpression(pattern: "(/(subscriptions))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: [.caseInsensitive])
+    static let graphSubscriptionIDRegex = try! NSRegularExpression(pattern: "(https://(graph.windows.net))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: [.caseInsensitive])
```
tjprescott, line 32: Why is this an extension, instead of just being a part of Filter?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+    
+    static let standard: Filter = {
+        var standardFilter = Filter()
+        let headers = ["Authorization" : FilterBehavior.remove]
```
tjprescott, line 36: If all of those headers referenced earlier should be removed, it would be more concise to do:
```swift
let headers: [String: FilterBehavior] = [
  "Authorization": .remove,
  "Foo": .remove,
  ...
]
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+// --------------------------------------------------------------------------
+
+import Foundation
+import DVR
+
+public extension Filter {
+    static let subscriptionIDRegex = try! NSRegularExpression(pattern: "(/(subscriptions))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: [.caseInsensitive])
+    static let graphSubscriptionIDRegex = try! NSRegularExpression(pattern: "(https://(graph.windows.net))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: [.caseInsensitive])
+    
+    static let standard: Filter = {
```
tjprescott, line 34: I'm not sure that it makes sense to do this. It might make more sense to instantiate a Filter object and then just use it:
```swift
let standard = Filter()
... do stuff
DVR(withFilter: standard)
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        var standardFilter = Filter()
+        let headers = ["Authorization" : FilterBehavior.remove]
+        standardFilter.beforeRecordRequest = scrubSubscriptionIDs
+        standardFilter.beforeRecordResponse = scrubSubscriptionIDs
+        return standardFilter
+    }()
+    
+    static func scrubSubscriptionId(from string: String?, embeddedStringKeys: [String] = ["tenantID","objectId"]) -> String? {
```
tjprescott, line 42: Why is this static?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        
+        
+        
+        let embeddedIDRegex = { (_ key : String) in
+            try! NSRegularExpression(pattern: "\"\(key)\":\"/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\"", options: [.caseInsensitive])
+        }
```
tjprescott, line 48: Why is this a closure?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        
+        guard var editedString = string else {
+            return nil
+        }
+        var range = NSRange(location: 0, length: editedString.utf8.count)
+        if Filter.subscriptionIDRegex.numberOfMatches(in: editedString, options: [], range: range) > 0 {
+            editedString = Filter.subscriptionIDRegex.stringByReplacingMatches(in: editedString, options: [], range: range, withTemplate: "$1/99999999-9999-9999-9999-999999999999")
+        }
```
tjprescott, line 56: If there are no matches, can't we just call line 55?

tjprescott: Same for below.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        if Filter.graphSubscriptionIDRegex.numberOfMatches(in: editedString, options: [], range: range) > 0 {
+            editedString = Filter.graphSubscriptionIDRegex.stringByReplacingMatches(in: editedString, options: [], range: range, withTemplate: "$1/99999999-9999-9999-9999-999999999999")
+        }
+        
+        for key in embeddedStringKeys {
+            let embeddedRegex = embeddedIDRegex(key)
+            range = NSRange(location: 0, length: editedString.utf8.count)
+            if embeddedRegex.numberOfMatches(in: editedString, options: [], range: range) > 0 {
+                editedString = embeddedRegex.stringByReplacingMatches(in: editedString, options: [], range: range, withTemplate: "\"\(key)\":\"/99999999-9999-9999-9999-999999999999\"")
+            }
```
tjprescott, line 67: Since this code is repetitive, this might be a good place for a regex extension.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return editedString
+        
+        
+    }
+    
+    static func scrubSubscriptionIDs(request: URLRequest) -> URLRequest {
```
tjprescott, line 74: Per swift naming guidelines, this should probably be something like `scrubSubscriptionsIds(for request: URLRequest)` or `scrubSubscriptionsIds(forRequest request: URLRequest)`

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+        
+        var cleanRequest = request
+        
+        let dirtyHeaders = request.allHTTPHeaderFields ?? [:]
+        let cleanHeaders = dirtyHeaders.mapValues { val -> String in
```
tjprescott, line 79: Do you have to specify `-> String`? Typically Swift can infer this

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+            guard let newValue = scrubSubscriptionId(from: val) else {
+                return val
+            }
+            return newValue
```
tjprescott, line 83:
```suggestion
            return scrubSubscriptionId(from: val) ?? val
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+    
+    static func scrubSubscriptionIDs(response: Foundation.URLResponse, data: Data?) -> (Foundation.URLResponse, Data?) {
+        guard let dirtyData = data else {
+            return (response, data)
+        }
+        
+        do {
+            guard let responseDataString = try JSONSerialization.jsonObject(with: dirtyData, options: .allowFragments) as? String else {
+                return (response, data)
+            }
```
tjprescott, line 102: If you are cleaning this as a string, why transform it to JSON?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,120 @@
...
+            let cleanedDataString = scrubSubscriptionId(from: responseDataString)
+            let cleanedData = cleanedDataString?.data(using: .utf8)
+            return (response, cleanedData)
```
tjprescott, line 105: You could make this more concise. However, I would question whether it makes sense to use `?` here. If `scrubSubscriptionId` returns nil for some weird reason, it will `nil` out the data that wouldn't otherwise be `nil`.

```suggestion
            let cleanedData = scrubSubscriptionId(from: responseDataString)?.data(using: .utf8)
            return (response, cleanedData)
```

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+import DVR
+
+public extension Filter {
```
tjprescott, line 30: Rather than do a public extension, why not simply create `class SubscriptionIdFilter: Filter`?  Then you don't need to declare a bunch of static variables.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+    
+    static let subscriptionIDRegex = "(/(subscriptions))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
+    static let graphSubscriptionIDRegex = "(https://(graph.windows.net))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
```
tjprescott, line 33: I believe that, in general, creating an `NSRegularExpression` is a relatively expensive operation. Thus, it would be better to do that here instead of just having the string pattern as a constant. Otherwise you are recreating the regex object every time you make a call to `replaceAllMatches`.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+    static var subcriptionIDReplacement = "99999999-9999-9999-9999-999999999999"
+    
+    static var standardHeaderScrubbing: [String : FilterBehavior] = [
+        "location": .closure({ key, val in
+            scrubSubscriptionId(from: val ?? "")
+        }),
+        "operation-location": .replace(""),
+        "azure-asyncoperation": .replace(""),
+        "www-authenticate": .replace(""),
+        "access_token": .replace("")
```
tjprescott, line 44: Should these be replaced with empty string or removed? For the other headers, they were simply removed. Can you find these headers in any of the Python recordings?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+        
+    ]
+    
+    static func scrubSubscriptionId(from string: String?) -> String? {
```
tjprescott, line 48: Even though this is an internal method, doc comments would still be helpful for intellisense purposes.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+    
+    
+    
+    func scrubSubscriptionIDs(from request: URLRequest) -> URLRequest {
+        
+        var cleanRequest = request
+        
+        let dirtyHeaders = request.allHTTPHeaderFields ?? [:]
+        let cleanHeaders = dirtyHeaders.mapValues { val in
+            return Filter.scrubSubscriptionId(from: val) ?? val
```
tjprescott, line 69: This is potentially dangerous. If something contains a subscription ID but the scrubbing fails (possibly due to a logic error) then this would use the original string, containing the subscription ID!

If a string doesn't contain a subscription ID, then it should return the original string, but this leaves it unclear why this would return `nil` and thus makes this suspect.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,111 @@
...
+extension String {
+    
+    mutating func replaceAllMatches(usingRegex: String, withReplacement: String, options: NSRegularExpression.Options = [], matchingOptions: NSRegularExpression.MatchingOptions = []) {
+        do {
+            let regularExpression = try NSRegularExpression(pattern: usingRegex, options: options)
+            let range = NSRange(location: 0, length: self.utf8.count)
+            self = regularExpression.stringByReplacingMatches(in: self, options: matchingOptions, range: range, withTemplate: withReplacement)
+        } catch {
+            return
+        }
```
tjprescott, line 102: I'd get rid of the `do-catch` by moving the `NSRegularExpression` initialization out of here. Otherwise, if an error is thrown, then the string stays unchanged, which isn't the intention.

sdk/test/AzureTest/Scrubbing.swift
tjprescott: I think this would define a much clearer contract if the signature were:
```swift
/// Returns the original string with subscription ID info (if any) replaced.
func scrubSubscriptionId(from string: String) -> String {...}
```

Many of my comments stem from seeing this invocation in the wild with no clear understanding of why it could return `nil`.

tjprescott (changes requested):

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,171 @@
...
+        let path = testBundle.path(forResource: "TestData", ofType: "json")
+        fakeData = try! Data(contentsOf: URL(fileURLWithPath: path!))
+        chooseRecordedInteraction(number: 8)
+        
+    }
+
+    func test_scrubbingRequest_removeSubscriptionIDs() throws {
+        let cleanedRequest = Filter().scrubSubscriptionIDs(from: fakeRequest!)
+        guard let shouldPass = cleanedRequest.url?.absoluteString.contains(regex: Filter.subcriptionIDReplacement) else {
+            XCTFail(); return
```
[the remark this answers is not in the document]
jairmyree: I changed this so there's no need for the return

sdk/test/AzureTest/Scrubbing.swift
jairmyree: It can return nil because URLRequest's url and a few others are optional strings. It would return nil in the if it's passed something nil.

tjprescott: Yep, I get that. If the only reason it will return nil is if it is passed nil, it should simply not be allowed to pass nil. While it can be a convenience (which we used in Swift APIView, for instance) in this case it is for scrubbing sensitive information. Allowing the method to return nil invites real or perceived vulnerabilities.

jairmyree: Would you prefer I overload this function with strictly a String parameter and return type?

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I think this is connected to my above explanation. If it isn't explained by that please let me know.

sdk/test/AzureTest/Scrubbing.swift
jairmyree: I wanted to ask you about this. After checking the recording we've been using, none of the above are being included. I'll look into it more but I believe that they should all just be removed so I'll change it.

jairmyree: This file is meant to implement the scrubbing things that absolutely should be scrubbed. However it's not enforced in any capacity and someone could improperly use the Filter and record sensitive information. Originally this was why I was writing the standard filter (so that there exists a readily accessible Filter with the appropriate scrubbing), but even that doesn't seem to be enough. What level of protection do you think we need against this? Should we make it not possible to use a filter that doesn't have this scrubbing place?

sdk/test/AzureTest/Scrubbing.swift
tjprescott: Not an overload--just remove the optionality in the parameter and return type.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    func test_scrubbingResponse_removeSubscriptionIDs() throws {
+        let dirtyHeaders = ["location": "[\"https://management.azure.com/subscriptions/72f988bf-86f1-41af-91ab-2d7cd011db47/providers/Microsoft.KeyVault/locations/eastus/operationResults/VVR8MDYzNzU0NDA3MTY0MzE2NTczMnwwNjZENTEwRTA4N0U0MTY5ODc1MDhDRDY3QUJDMzdGOQ?api-version=2019-09-01\"]"]
+        
+        let dirtyBody = """
+            "string": "{\"id\":\"/subscriptions/72f988bf-86f1-41af-91ab-2d7cd011db47/providers/Microsoft.KeyVault/locations/eastus/deletedVaults/myValtZikfikxz\",\"name\":\"myValtZikfikxz\",\"type\":\"Microsoft.KeyVault/deletedVaults\",\"properties\":{\"vaultId\":\"/subscriptions/72f988bf-86f1-41af-91ab-2d7cd011db47/resourceGroups/rgname/providers/Microsoft.KeyVault/vaults/myValtZikfikxz\",\"location\":\"eastus\",\"tags\":{},\"deletionDate\":\"2021-04-19T05:32:42Z\",\"scheduledPurgeDate\":\"2021-07-18T05:32:42Z\"}}"
+        """
+        
+        var shouldPass = false
```
tjprescott, line 70: There's still no reason for this declaration.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        
+        let cleanLocation = Filter.scrubSubscriptionId(from: dirtyHeaders["location"])
+        let cleanBody = Filter.scrubSubscriptionId(from: dirtyBody)
+        
+        shouldPass = (cleanLocation?.contains(regex: Filter.subcriptionIDReplacement) ?? false) && (cleanBody?.contains(regex: Filter.subcriptionIDReplacement) ?? false)
```
tjprescott, line 75:
```suggestion
        let shouldPass = cleanLocation!.contains(regex: Filter.subcriptionIDReplacement) && cleanBody!.contains(regex: Filter.subcriptionIDReplacement)
```

However, as I look at this, perhaps the criteria should actually be that we _don't_ find the original value rather than the fact that we _do_ find the replacement.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,117 @@
...
+    }
+    
+    /// Attempts to scrub subsctiption ID from optional Data that represents a string. Returns null if the input is nil or if the Data is not String-representable.
+    static func tryScrubSubscriptionId(from data: Data?) -> Data? {
+        
+        guard let stringData = String(data: data, encoding: .utf8) else {
+            return nil
+        }
+        return scrubSubscriptionId(from: stringData).data(using: .utf8)
+    }
```
tjprescott, line 75: These methods preserve the same ambiguity that made the original implementation suspect. Just refactor `scrubScubscriptionIDs` to early out if it is unable to get to an actual usable string instead of chaining optionals.

tjprescott (changes requested):
There are unaddressed comments from the previous review, and some new ones.

sdk/test/AzureTest/Scrubbing.swift
jairmyree: Im gonna make a comment about this on the updated version

sdk/test/AzureTest/Scrubbing.swift
tjprescott: The default for AzureTest should be to use this filter. I might not be understanding the scenario. We can hop on a meeting if needed

sdk/test/AzureTest/DVRSessionTransport.swift
jairmyree: Oh I changed something similar to this in a different file

sdk/test/AzureTest/Scrubbing.swift
jairmyree: That won't work because Filter comes from DVR and is a struct.
Classes can't inherit from structs and structs can't inherit from structs either.

sdk/test/AzureTest/Scrubbing.swift
jairmyree: Maybe we should have a call because I'm not understanding the ambiguity you're talking about. If I don't allow the function to take an optional then all that means is everywhere I use it I'd have to guard against nil which is definitely an option if you'd prefer that route but it seems like that way adds more unnecessary code. I want to understand better the ambiguity you are mentioning.

Podfile
```diff
@@ -95,14 +95,6 @@ target 'AzureStorageBlob' do
...
 end
 
-target 'AzureTemplate' do
-  project 'sdk/template/AzureTemplate/AzureTemplate'
-
-  target 'AzureTemplateTests' do
-    inherit! :search_paths
-  end
-end
-
```
tjprescott, line 105: Why is this being removed?

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        let path = testBundle.path(forResource: "TestData", ofType: "json")
+        fakeData = try! Data(contentsOf: URL(fileURLWithPath: path!))
+        chooseRecordedInteraction(number: 8)
+        
+    }
+
+    func test_scrubbingRequest_removeSubscriptionIDs() throws {
+        let cleanedRequest = Filter().scrubSubscriptionIDs(from: fakeRequest!)
+        let shouldPass = cleanedRequest.url?.absoluteString.contains(regex: Filter.subcriptionIDReplacement) ?? false
+        XCTAssert(shouldPass)
```
tjprescott, line 59: Same as with the `test_scrubbingResponse_removeSubscriptionIDs` test, we should test that the request no longer contains the GUID, not that it does contain the replacement.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        
+        let dirtyBody = """
+            "string": "{\"id\":\"/subscriptions/72f988bf-86f1-41af-91ab-2d7cd011db47/providers/Microsoft.KeyVault/locations/eastus/deletedVaults/myValtZikfikxz\",\"name\":\"myValtZikfikxz\",\"type\":\"Microsoft.KeyVault/deletedVaults\",\"properties\":{\"vaultId\":\"/subscriptions/72f988bf-86f1-41af-91ab-2d7cd011db47/resourceGroups/rgname/providers/Microsoft.KeyVault/vaults/myValtZikfikxz\",\"location\":\"eastus\",\"tags\":{},\"deletionDate\":\"2021-04-19T05:32:42Z\",\"scheduledPurgeDate\":\"2021-07-18T05:32:42Z\"}}"
+        """
+        
+        let cleanLocation = Filter.scrubSubscriptionId(from: dirtyHeaders["location"]!)
+        let cleanBody = Filter.scrubSubscriptionId(from: dirtyBody)
+        
+        let shouldPass = !(cleanLocation.contains(regex: "72f988bf-86f1-41af-91ab-2d7cd011db47")) && !(cleanBody.contains(regex: "72f988bf-86f1-41af-91ab-2d7cd011db47"))
```
tjprescott, line 73: Perhaps extract out the GUID to a constant. More maintainable via the DRY principle in case you every need to change it (for whatever reason).

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
tjprescott: I would also force unwrap things like `url` so you don't need `?? false`.

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,131 @@
...
+    static let subscriptionIDRegex = try! NSRegularExpression(pattern: "(/(subscriptions))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: .caseInsensitive)
+    static let graphSubscriptionIDRegex = try! NSRegularExpression(pattern: "(https://(graph.windows.net))/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}", options: .caseInsensitive)
+    
+    static var subcriptionIDReplacement = "00000000-0000-0000-0000-000000000000"
+    
+    /// Wraps `ScrubSubscriptionId` in a function to be passed as closure within initializer
+    func scrubSubscriptionIDClosure(key: String, val: String?) -> (String?) {
+        guard let val = val else { return nil }
+        return scrubSubscriptionId(from: val)
+    }
```
jairmyree, line 67: Originally I did not want to write this but since `scrubSubscriptionId` is not  static, I couldn't see another way to get the function inside the closure for "location" since it didn't have a matching parameter and return type. What's your thoughts on this @tjprescott ?

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -0,0 +1,131 @@
...
+            "x-ms-correlation-request-id": .remove,
+            "x-ms-ratelimit-remaining-subscription-reads": .remove,
+            "x-ms-request-id":.remove,
+            "x-ms-routing-request-id": .remove,
+            "x-ms-gateway-service-instanceid": .remove,
+            "x-ms-ratelimit-remaining-tenant-reads": .remove,
+            "x-ms-served-by": .remove,
+            "x-ms-authorization-auxiliary": .remove
+       ]
+        filterHeaders = standardHeaderScrubbing
```
tjprescott, line 53: These should not be in the SubscriptionIDFilter as they have nothing to do with it. These should be in another filter and then both filters passed to the DVRSession.

tjprescott (changes requested):
Looks a lot better. A few lingering comments.

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
jairmyree: Should I just remove the test json and only take the pieces that we want to use and hard code them in? We aren't using much of that json file at all.

sdk/test/AzureTest/Scrubbing.swift
tjprescott: I think it's fine since that's the signature for the closure enum

sdk/test/AzureTest/AzureTestTests/AzureTestTests.swift
tjprescott: Yeah, that's probably fine.

Podfile
jairmyree: I don't know why but the changes to the Podfile that we made a while back when we had our merge conflict keep coming back. So I have to keep on removing them. This is like the third time. If you look up in the file it was defined twice, and this one is the one with the error because `AzureTemplateTests` does not exist on our branch of AzureTest.

jairmyree (commented):
I've reviewed your comments and I've made changes and also left some comments of my own. Take a look when you get a chance @tjprescott

sdk/test/AzureTest/Scrubbing.swift
tjprescott: This doesn't look resolved.

jairmyree (commented):
I fixed the piece that I missed earlier

sdk/test/AzureTest/Scrubbing.swift
```diff
@@ -36,19 +36,7 @@ public class SubscriptionIDFilter : Filter {
...
-            "x-ms-client-request-id": .remove,
-            "x-ms-correlation-request-id": .remove,
-            "x-ms-ratelimit-remaining-subscription-reads": .remove,
-            "x-ms-request-id":.remove,
-            "x-ms-routing-request-id": .remove,
-            "x-ms-gateway-service-instanceid": .remove,
-            "x-ms-ratelimit-remaining-tenant-reads": .remove,
-            "x-ms-served-by": .remove,
-            "x-ms-authorization-auxiliary": .remove
+           "access_token": .remove
```
tjprescott: Any reason these aren't also in the same place as the other header removals?

sdk/test/AzureTest/DVRSessionTransport.swift
```diff
@@ -64,6 +64,20 @@ public class DVRSessionTransport: TransportStage {
...
+            "x-ms-client-request-id": .remove,
+            "x-ms-correlation-request-id": .remove,
+            "x-ms-ratelimit-remaining-subscription-reads": .remove,
+            "x-ms-request-id":.remove,
+            "x-ms-routing-request-id": .remove,
+            "x-ms-gateway-service-instanceid": .remove,
+            "x-ms-ratelimit-remaining-tenant-reads": .remove,
+            "x-ms-served-by": .remove,
+            "x-ms-authorization-auxiliary": .remove
+            ]
```
tjprescott, line 80: Two things I see here... 
First, replacements isn't being hooked up to anything. 
Second, your SubscriptionIDFilter also isn't hooked up!

This should look like:
```swift
let defaultFilter = Filter()
defaultFilter.filterHeaders = ["authorization": .remove, ...]
session.filters = [SubscriptionIDFilter(), defaultFilter]
```

tjprescott (changes requested):

sdk/test/AzureTest/Scrubbing.swift
jairmyree: It's because those headers are listed with the original Subscription Id issue so I handled them along with the IDs.

jairmyree: If you think they'd be better placed with the others I can move them over quick.

sdk/test/AzureTest/DVRSessionTransport.swift
jairmyree: Yea that's what I was talking about when I asked about where we're going with the filters and how much are we leaving open to developers. I was trying to figure out whether or not our goal was to connect it for devs or give them everything so they can connect it themselves.

sdk/test/AzureTest/Scrubbing.swift
tjprescott: Yes, even in that issue, they are separate bullets from Subscription ID.

sdk/test/AzureTest/DVRSessionTransport.swift
tjprescott: We should provide the default here in DVRSessionTransport. For this PR we don't need to worry about exposing an extension mechanism, but we should wire up the defaults. Extensibility can be a separate PR

jairmyree (commented):
The changes were made

tjprescott (approved):

PR #935, opened 2021-06-29
Scrub headings and commit chat test recordings.

tjprescott:
Uses the `filterHeaders` property of DVR to remove authorization headers, which permits us to commit the Chat test recordings.

DVR still needs to be updated to that header filters are done in a case insensitive way.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -95,14 +95,6 @@ target 'AzureStorageBlob' do
   end
 end
 
-target 'AzureTemplate' do
-  project 'sdk/template/AzureTemplate/AzureTemplate'
-
-  target 'AzureTemplateTests' do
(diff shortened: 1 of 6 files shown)


Podfile
```diff
@@ -95,14 +95,6 @@ target 'AzureStorageBlob' do
...
 end
 
-target 'AzureTemplate' do
-  project 'sdk/template/AzureTemplate/AzureTemplate'
-
-  target 'AzureTemplateTests' do
-    inherit! :search_paths
-  end
-end
-
```
tjprescott, line 105: This is unrelated, but this duplicate entry resulted in `pod install` failing.

jairmyree (approved):

PR #946, opened 2021-07-08
Approved: Chat GA swiftinterface

knvsl:
PR to address comments here: https://github.com/Azure/azure-sdk-for-ios/pull/948


examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  public func listReadReceipts(withOptions options: AzureCommunicationChat.ListChatReadReceiptsOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCore.PagedCollection<AzureCommunicationChat.ChatMessageReadReceipt>>)
+  public func sendTypingNotification(withOptions options: AzureCommunicationChat.SendTypingNotificationOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<Swift.Void>)
+  public func send(message: AzureCommunicationChat.SendChatMessageRequest, withOptions options: AzureCommunicationChat.SendChatMessageOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCommunicationChat.SendChatMessageResult>)
+  public func get(message messageId: Swift.String, withOptions options: AzureCommunicationChat.GetChatMessageOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<AzureCommunicationChat.ChatMessage>)
+  public func update(content: Swift.String, forMessage messageId: Swift.String, withOptions options: AzureCommunicationChat.UpdateChatMessageOptions? = nil, completionHandler: @escaping AzureCore.HTTPResultHandler<Swift.Void>)
```
tjprescott, line 62: Is this for editing content? If so, a name like `update(message messageId: String, content: String, ...)` seems more natural and follow the typical Azure conventions of specifying things from general to specific and identifying the target resource first.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+}
+public struct AddChatParticipantsOptions : AzureCore.RequestOptions {
+  public let clientRequestId: Swift.String?
+  public let cancellationToken: AzureCore.CancellationToken?
+  public var dispatchQueue: Dispatch.DispatchQueue?
+  public var context: AzureCore.PipelineContext?
+  public init(clientRequestId: Swift.String? = nil, cancellationToken: AzureCore.CancellationToken? = nil, dispatchQueue: Dispatch.DispatchQueue? = nil, context: AzureCore.PipelineContext? = nil)
+}
+public struct CommunicationIdentifierModel : Swift.Codable {
+  public let rawId: Swift.String?
```
tjprescott, line 149: Why is this public? Is this the same value that `communicationUser.id` would have?

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  public init(rawId: Swift.String? = nil, communicationUser: AzureCommunicationChat.CommunicationUserIdentifierModel? = nil, phoneNumber: AzureCommunicationChat.PhoneNumberIdentifierModel? = nil, microsoftTeamsUser: AzureCommunicationChat.MicrosoftTeamsUserIdentifierModel? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct AddChatParticipantsRequest : Swift.Codable {
+  public let participants: [AzureCommunicationChat.ChatParticipant]
+  public init(participants: [AzureCommunicationChat.ChatParticipant])
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
```
tjprescott, line 162: Why is this public? Your public method takes the `[ChatParticipant]` collection directly, so I would think this should not be part of the public API.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  public init(error: AzureCommunicationChat.ChatError)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
+public struct UpdateChatThreadRequest : Swift.Codable {
+  public let topic: Swift.String?
+  public init(topic: Swift.String? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
```
tjprescott, line 204: Again, this seems like it should be internal. It is not referenced in any other public call.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  case chatThreadDeleted(AzureCommunicationChat.ChatThreadDeletedEvent)
+  case participantsAdded(AzureCommunicationChat.ParticipantsAddedEvent)
+  case participantsRemoved(AzureCommunicationChat.ParticipantsRemovedEvent)
+}
+public struct UpdateChatMessageRequest : Swift.Codable {
+  public let content: Swift.String?
+  public init(content: Swift.String? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
```
tjprescott, line 263: Also no references in the public API. Should be internal.

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  @objc deinit
+}
+public struct SendChatMessageRequest : Swift.Codable {
+  public let content: Swift.String
+  public let senderDisplayName: Swift.String?
+  public let type: AzureCommunicationChat.ChatMessageType?
+  public init(content: Swift.String, senderDisplayName: Swift.String? = nil, type: AzureCommunicationChat.ChatMessageType? = nil)
+  public init(from decoder: Swift.Decoder) throws
+  public func encode(to encoder: Swift.Encoder) throws
+}
```
tjprescott, line 299: We've probably discussed this before, but why does create use a `SendRequest` object but update just does that inline and doesn't expose the `UpdateRequest` object?

tjprescott (commented):
Seems like there are some internals exposed.

examples/x86_64.swiftinterface
JoshuaLai: I think in this case, if memory serves, `CommunicationIdentifierModel` can only have one of `communicationUser` or `phoneNumber` or `microsoftTeamsUser`.  
Also when we make a request using this object it is optional, but when we are decoding a response it is required.

examples/x86_64.swiftinterface
knvsl: With update you can only update the string content so we do it inline instead of constructing the request. This will change in preview6 though because you can also edit metadata, in preview6 we do use UpdateChatMessageRequest

examples/x86_64.swiftinterface
knvsl: This will be public in preview6 btw (related to your other question) - I'll regenerate with updated autorest config & make this internal

examples/x86_64.swiftinterface
```diff
@@ -0,0 +1,507 @@
...
+  case typingIndicatorReceived
+  case readReceiptReceived
+  case chatMessageEdited
+  case chatMessageDeleted
+  case chatThreadCreated
+  case chatThreadPropertiesUpdated
+  case chatThreadDeleted
+  case participantsAdded
+  case participantsRemoved
+  public init?(rawValue: Swift.String)
```
JoshuaLai, line 398: Should this init throw instead of just being optional? I assume it is just optional in the case if a string comes in that we don't recognize?

knvsl: I am not actually sure why swiftinterface is generated this way - it does throw if we get an unrecognized event id
https://github.com/Azure/azure-sdk-for-ios/blob/6e984283fa291312a12ed561a61b1b69734bcc50/sdk/communication/AzureCommunicationChat/Source/Signaling/Events/ChatEvent.swift#L900

JoshuaLai: Oh interesting. I wonder if it is because it is an enum.

examples/x86_64.swiftinterface
tjprescott: If you are going to expose additional things in the future, you are probably better off exposing the UpdateRequest now instead of having a breaking change in a month.

examples/x86_64.swiftinterface
tjprescott: You should update this to use UpdateChatMessageRequest then. I have a hard time believing the ArchBoard would approve breaking change immediately after GA and this is entirely foreseeable.

juancamilor: it would not be a breaking change if this becomes an overload. The other SDKs that were GAd before went though the review already. With this said though, given that this has not been released, I agree that it is better to do it now.

knvsl: We received previous feedback to change arguments to be inline so that was the original motivation. This was before preview6 came out, I can update it now

tjprescott: You could do an overload, true, but it will may add customer confusion. Let me know how you'd like to proceed.

tjprescott (approved):

tjprescott:
This API is approved. Closing this PR.

PR #961, opened 2021-07-21
ChatClient set default applicationId to empty string + add missing signaling init step

knvsl:
- applicationId should be empty by default instead of the bundle identifier
- add missing trouter call to set activity state

### DIFF
--- a/sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
+++ b/sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
@@ -59,13 +59,32 @@ public class ChatClient {
             throw AzureError.client("Unable to form base URL.")
         }
 
+        // If applicationId is not provided bundle identifier will be used
+        // Instead set the default application id to be an empty string
+        var updatedOptions: AzureCommunicationChatClientOptions?
+        if options.telemetryOptions.applicationId == nil {
+            let apiVersion = AzureCommunicationChatClientOptions.ApiVersion(options.apiVersion)
+            let telemetryOptions = TelemetryOptions(
+                telemetryDisabled: options.telemetryOptions.telemetryDisabled,
+                applicationId: ""
+            )
+
+            updatedOptions = AzureCommunicationChatClientOptions(
+                apiVersion: apiVersion,
+                logger: options.logger,
+                telemetryOptions: telemetryOptions,
+                transportOptions: options.transportOptions,
+                dispatchQueue: options.dispatchQueue
+            )
+        }
+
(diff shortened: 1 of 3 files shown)


sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -59,13 +59,32 @@ public class ChatClient {
...
+        }
+
         let communicationCredential = TokenCredentialAdapter(credential)
         let authPolicy = BearerTokenCredentialPolicy(credential: communicationCredential, scopes: [])
 
         let client = try ChatClientInternal(
             endpoint: endpointUrl,
             authPolicy: authPolicy,
-            withOptions: options
+            withOptions: updatedOptions ?? options
```
knvsl, line 87: @tjprescott I realized that UserAgentPolicy() uses separate TelemetryOptions() so this doesn't cover the case where someone provides UserAgentPolicy with telemetry options that have nil applicationId. 

We only want to change the applicationId to an empty string if it is nil, I'm not sure how we can do this for UserAgentPolicy()?

tjprescott: The telemetryOptions you pass into your client contructor get passed into the UserAgentPolicy in your `ChatClientInternal` initializer (line 44).

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
```diff
@@ -79,6 +79,7 @@ class CommunicationSignalingClient {
     func start() {
         selfHostedTrouterClient.withRegistrar(trouterUrlRegistrar)
         selfHostedTrouterClient.start()
+        selfHostedTrouterClient.setUserActivityState(UserActivityState.TrouterUserActivityStateActive)
```
tjprescott, line 82: Does this have anything to do with the telemetry??

tjprescott (approved):

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
knvsl: No, this is unrelated & to enable connection status on trouter side. It doesn't change anything on SDK but we need it for an internal customer

knvsl: I can split this into separate PR but its not really related to any feature we are releasing, so just included it here.

tjprescott (approved):

PR #962, opened 2021-07-21
Remove Chat (and Calling and Common) from list of libraries in beta

knvsl:
@tjprescott Just to confirm will this also update the landing page of CocoaPods?
https://cocoapods.org/pods/AzureCommunicationChat


tjprescott (approved):

tjprescott:
@knvsl not certain this will fix the CocoaPods page, but right now iOS build agents are busted. Once they are running we will be able to see. Thanks for the catch!

PR #963, opened 2021-07-21
RPF: RecordableXCTestCase

### LINKED ISSUE #960: RPF: Wrapper Class
While writing the wiki article for RPF, it became clear that we still require too much from test authors to use the framework. We should provide a wrapper class to simplify the process and eliminate boilerplate code.

tjprescott:
A base class to eliminate boilerplate when writing recordable tests.

Closes #960.


sdk/communication/AzureCommunicationChat/Source/Signaling/Events/ChatEvent.swift
```diff
@@ -41,7 +41,7 @@ public struct SignalingChatParticipant {
     public let shareHistoryTime: Iso8601Date?
 
     // MARK: Initializers
-    
+
```
jairmyree, line 44: I'm guessing this is swift lint doing this, right?

sdk/template/AzureTemplate/Source/Dummy.swift
```diff
@@ -27,4 +27,3 @@
 import Foundation
 
 // Does nothing
-
```
jairmyree, line 30: Is there a reason for this to exist?

sdk/test/AzureTest/RecordableXCTestCase.swift
```diff
@@ -0,0 +1,65 @@
...
+        testName.removeLast()
+        transport = mode != "live" ? DVRSessionTransport(cassetteName: String(testName)) : URLSessionTransport()
+        try setUpTestWithError()
+        transport?.open()
+    }
+
+    /// Method which the test author can override to configure setup
+    open func setUpTestWithError() throws {}
+
+    override public final func tearDownWithError() throws {}
```
jairmyree, line 57: Is there a purpose for the overrides with no bodies? There are several of them in this file. If there's no logic specific to recording the tests couldn't the test author just override those functions themself when necessary?

sdk/test/AzureTest/TestSettings.swift
```diff
@@ -27,17 +27,17 @@
...
 public extension TestSettingsProtocol {
-    static func loadFromPlist() -> Self {
+    static func loadFromPlist() -> Self? {
         if let path = Bundle(for: self).path(forResource: "test-settings", ofType: "plist"),
             let xml = FileManager.default.contents(atPath: path),
             let settings = try? PropertyListDecoder().decode(Self.self, from: xml) {
             return settings
         } else {
-            fatalError("Unable to decode test-settings.plist!")
+            return nil
```
jairmyree, line 40: Why is returning nil a better choice than having a fatal error?

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientUnitTests.swift
```diff
@@ -45,8 +44,8 @@ class ChatThreadClientUnitTests: XCTestCase {
     private let settings = TestSettings.loadFromPlist()
 
     override func setUpWithError() throws {
-        let endpoint = settings.endpoint ?? "https://endpoint"
-        let token = settings.token ?? generateFakeToken()
+        let endpoint = settings?.endpoint ?? "https://endpoint"
+        let token = settings?.token ?? generateFakeToken()
```
jairmyree, line 48: Is this token and endpoint part not related to test recording? I notice that it's still in the file but AzureTest had its import removed.

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientTests.swift
```diff
@@ -592,7 +592,7 @@ class ChatThreadClientTests: XCTestCase {
     }
 
     func test_ListParticipants_ReturnsParticipants() {
-        let user2 = settings.user2!
```
jairmyree, line 595: Why the change to the force unwrap when the previous version wasn't optional? What benefits does settings gain from being made optional?

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientTests.swift
```diff
@@ -53,8 +53,8 @@ class ChatThreadClientTests: XCTestCase {
 
     override func setUpWithError() throws {
         // Initialize the chatClient
-        let endpoint = settings.endpoint ?? "https://endpoint"
```
jairmyree, line 56: If the logic is the same for both, why change settings to be able to be nil?

sdk/test/AzureTest/RecordableXCTestCase.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+// --------------------------------------------------------------------------
+
+import AzureCore
+import XCTest
+
+open class RecordableXCTestCase<SettingsType: TestSettingsProtocol>: XCTestCase {
+    public let settings = {
+        SettingsType.loadFromPlist()
+    }()
```
jairmyree, line 33: What are the implications of `settings` being nil?

jairmyree (commented):
I left some comments with questions

sdk/test/AzureTest/RecordableXCTestCase.swift
tjprescott: The whole point is to prevent the dev from overriding the superclass implementations (any of them) and instead provide a suitable hook that is sequenced into the desired template function.

sdk/template/AzureTemplate/Source/Dummy.swift
tjprescott: It's for engineering system testing. There has to be at least one source file.

sdk/test/AzureTest/TestSettings.swift
tjprescott: It's possible (though unlikely) that you could have a service that doesn't actually have or need any test settings.

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientUnitTests.swift
tjprescott: They are used to construct the client. They aren't needed by AzureTest, per se

sdk/test/AzureTest/RecordableXCTestCase.swift
tjprescott: It just means you can't access any settings. You may not need to.

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientTests.swift
tjprescott: In the base class, settings was changed to optional, so we have to make this change here.

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientTests.swift
tjprescott: A recordable test isn't required to have settings--we just provide a mechanism for them to do so.

sdk/communication/AzureCommunicationChat/Tests/ChatThreadClientUnitTests.swift
jairmyree: Ok I was just wondering :)

jairmyree (approved):

PR #967, opened 2021-07-26
Add AzureTest framework

tjprescott:
Adds AzureTest for creating record/playback style tests. Add RecordableXCTestCase to make it easier to add recordable tests.

Documentation provided here:
https://github.com/Azure/azure-sdk-for-ios/wiki/Writing-End-to-End-Tests-using-AzureTest


eng/scripts/prepare_chat_tests.py
```diff
@@ -5,36 +5,49 @@
```
benbp, line 5: Would this file be better placed in the ACS service directory?

tjprescott: That's what I would have preferred if I could have used `test-resources-post.py`. Absent that, I don't want to clutter the service directory and figured it would be best to keep this in the "normal" spot under `eng/scripts`.

sdk/communication/AzureCommunicationChat/AzureCommunicationChat.xcodeproj/xcshareddata/xcschemes/AzureCommunicationChat.xcscheme
```diff
@@ -34,23 +34,8 @@
...
-         <EnvironmentVariable
-            key = "AZURE_COMMUNICATION_USER_ID_1"
-            value = ""
-            isEnabled = "YES">
-         </EnvironmentVariable>
-         <EnvironmentVariable
-            key = "AZURE_COMMUNICATION_USER_ID_2"
-            value = ""
+            key = "SDK_REPO_ROOT"
+            value = "/Users/travisprescott/repos/azure-sdk-for-ios"
```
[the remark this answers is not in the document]
tjprescott: This is my Mac user directory and yes, it is intentional. Anyone who wants to run the test would need to update this with their own settings, but since the build artifact isn't "aware" of the source directory, it (seemingly) has to be hard coded like this for the recording functionality to work correctly.

PR #977, opened 2021-07-30
Fix Package.swift for AzureCommunicationCommon

tjprescott:
Remove AzureCore dependency in Package.swift to match podspec.


JoshuaLai (approved):
lgtm

PR #981, opened 2021-08-02
Fix matching issues with playback tests

### LINKED ISSUE #976: [BUG] ChatClientDVRTest requests fail to match the request in the cassette
Running the DVR-based tests for Chat in playback mode fails because the requests don't match what is persisted in the cassette. This seems to be an issue with how the request URLs are either filtered or persisted.

tjprescott:
Closes #976.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -52,7 +52,6 @@ target 'AzureCommunicationChat' do
     inherit! :search_paths
     pod 'OHHTTPStubs/Swift'
     pod 'Trouter', '0.0.1-beta.5'
-    pod 'MSAL', '1.1.15'
     if $use_local_dvr
         pod 'DVR', :path => $dvr_path
     else
@@ -64,6 +63,11 @@ target 'AzureCommunicationChat' do
     inherit! :search_paths
     pod 'OHHTTPStubs/Swift'
     pod 'Trouter', '0.0.1-beta.5'
+    if $use_local_dvr
+        pod 'DVR', :path => $dvr_path
+    else
+        pod 'DVR', :git => 'https://github.com/tjprescott/DVR.git'
+    end
   end
 end
 
--- a/sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
+++ b/sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
@@ -27,7 +27,6 @@
 import AzureCommunicationChat
 import AzureCommunicationCommon
 import AzureCore
-import AzureTest
 import OHHTTPStubs.Swift
 import XCTest
 
--- a/sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
+++ b/sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
@@ -39,9 +39,9 @@ class ChatClientUnitTests: XCTestCase {
     private let topic = "test topic"
 
     override func setUpWithError() throws {
-        let settings = TestSettings.loadFromPlist()
-        let endpoint = settings?.endpoint ?? "https://endpoint"
-        let token = settings?.token ?? generateFakeToken()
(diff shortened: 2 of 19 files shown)


Podfile
```diff
@@ -52,7 +52,6 @@ target 'AzureCommunicationChat' do
     inherit! :search_paths
     pod 'OHHTTPStubs/Swift'
     pod 'Trouter', '0.0.1-beta.5'
-    pod 'MSAL', '1.1.15'
```
jairmyree, line 55: What is this pod and there a reason why it is being removed?

sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
```diff
@@ -39,9 +38,9 @@ class ChatClientUnitTests: XCTestCase {
...
 
     override func setUpWithError() throws {
-        let settings = TestSettings.loadFromPlist()
-        let endpoint = settings?.endpoint ?? "https://endpoint"
-        let token = settings?.token ?? generateFakeToken()
+        let settings = TestSettings()
+        let endpoint = settings.endpoint
+        let token = settings.token
         let credential = try CommunicationTokenCredential(token: token)
         let options = AzureCommunicationChatClientOptions()
```
jairmyree, line 45: This appears in `setupWithError()` for just about every Chat test class. Would this be worth refactoring? If so, I'm not sure if that would be something you'd want to do or leave for chat.

sdk/communication/AzureCommunicationChat/Tests/TestSettings.swift
```diff
@@ -28,8 +28,22 @@ import AzureTest
...
 class TestSettings: TestSettingsProtocol {
-    var endpoint: String?
-    var user1: String?
-    var user2: String?
-    var token: String?
+    var endpoint = "https://endpoint/"
+    var user1 = "user1"
+    var user2 = "user2"
+    var token: String = {
+        let fakeValue = "{\"iss\":\"ACS\",\"iat\": 1608152725,\"exp\": 1739688725,\"aud\": \"\",\"sub\": \"\"}"
```
jairmyree, line 35: Just curious, what format is this?

sdk/communication/AzureCommunicationChat/Tests/TestSettings.swift
```diff
@@ -28,8 +28,22 @@ import AzureTest
...
+            .base64EncodedString()
+        let token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9." + fakeValue + ".EMS0ExXqRuobm34WKJE8mAfZ7KppU5kEHl0OFdyree8"
```
jairmyree, line 37: Same here, just curious as to what the token format is

jairmyree (approved):
I have some questions but they're just informational.

Podfile
tjprescott: It's not used by Chat so there's no reason for it to be listed as a dependency.

sdk/communication/AzureCommunicationChat/Tests/ChatClientUnitTests.swift
tjprescott: I'd leave it for the chat folks, but yeah, they could refactor it.

sdk/communication/AzureCommunicationChat/Tests/TestSettings.swift
tjprescott: JSON string representation of their token object.

sdk/communication/AzureCommunicationChat/Tests/TestSettings.swift
tjprescott: Some weird representation of their token. For our intents and purposes, it's just a string.

jairmyree (approved):

PR #1009, opened 2021-08-27
AzureCommunicationChat 1.0.2 patch

knvsl:
- **Fix displayName is optional to handle null displayName in certain trouter events**
- **Upgrade to Trouter package that contains plist fix**
- **Fix signaling client ignoring forceRefresh:**
  - Create token provider in Chat that stops connection if max retries are exceeded
  - Make AzureCommunicationChatClientOptions internal and inject error handler into public options, this is so user can provide a callback that's invoked if we stop the trouter connection
- Add unit tests to validate the trouter payloads
- Fix recipientId in TrouterPayloads for rest of events (was only applied to ChatMessageReceivedEvent)
- Fix readOn property in readReceiptReceived events was always nil due to not converting it properly
- Remove senderDisplayName from Typing/ReadReceipt events


sdk/communication/AzureCommunicationChat/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 1.0.2 (TODO add date)
+### Key Bug Fixes
```
tjprescott, line 4: This needs to be Bugs Fixed to align with updated Changelog guidance:
https://azure.github.io/azure-sdk/policies_releases.html#change-logs

sdk/communication/AzureCommunicationChat/Source/Generated/Options/AzureCommunicationChatClientOptionsInternal.swift
```diff
@@ -15,31 +15,31 @@ import Foundation
 // swiftlint:disable line_length
 
 /// User-configurable client options.
-public struct AzureCommunicationChatClientOptions: ClientOptions {
```
tjprescott, line 18: What's the difference between this model and the public version? If it's just the errorHandler you can just set it to nil for the internal usage.

tjprescott (changes requested):
A couple questions.

sdk/communication/AzureCommunicationChat/Source/Generated/Options/AzureCommunicationChatClientOptionsInternal.swift
knvsl: Thats correct its just the error handler. was trying to avoid modifying the generated code, but probably a better way. Is there a way I can get autorest to add the error handler to the options?

knvsl: ^AzureCommunicationChatClientOptions is autogenerated so I was trying to avoid manually adding error handler in the generated code

knvsl:
@tjprescott btw we targeted September 3rd to release the patch (earlier if it looks good before that)

CHILIU-MSFT:
Please make sure the follow scenarios work fine:
1) Invalid Skypetoken could not start Realtime Notification. Expect request stops within limited retries to service.

2) Valid Skypetoken could start Realtime Notification.

3) If a valid Skypetoken turns invalid after realtime notification is started, expect request stops within limited retries to service if there's no Skypetoken renew mechanism provided.

sdk/communication/AzureCommunicationChat/Package.swift
```diff
@@ -48,7 +48,7 @@ let package = Package(
         .package(
             name: "Trouter",
             url: "https://github.com/microsoft/trouter-client-ios.git",
-            .branch("v0.0.1-beta.5")
+            from: "0.1.0"
```
tjprescott, line 51: This should use `.exact("0.1.0")`. Otherwise a breaking change to trouter could be released that chat is not compatible with.

knvsl:
> Please make sure the follow scenarios work fine:
> 
> 1. Invalid Skypetoken could not start Realtime Notification. Expect request stops within limited retries to service.
> 2. Valid Skypetoken could start Realtime Notification.
> 3. If a valid Skypetoken turns invalid after realtime notification is started, expect request stops within limited retries to service if there's no Skypetoken renew mechanism provided.

Verified requests stop after the retries are exceeded & we stop the connection. If we use an invalid token startRealtimeNotifications completion handler will initially return success but if token is still invalid after max retries, the connection will be stopped.

sdk/communication/AzureCommunicationChat/Source/Generated/Options/AzureCommunicationChatClientOptionsInternal.swift
tjprescott: That makes sense and no, you can't add anything to the generated class.

However, you could make everything use just the public class and get rid of the internal class entirely.

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
```diff
@@ -36,11 +48,11 @@ class CommunicationSignalingClient {
     private var communicationHandlers: [ChatEventId: CommunicationHandler] = [:]
 
     init(
-        skypeTokenProvider: CommunicationSkypeTokenProvider,
+        communicationSkypeTokenProvider: CommunicationSkypeTokenProvider,
         logger: ClientLogger = ClientLoggers.default(tag: "AzureCommunicationSignalingClient")
     ) throws {
```
tjprescott, line 53: CommunicationSignalingClient is internal, yes? Otherwise this is a breaking change.

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
```diff
@@ -106,13 +111,64 @@ class CommunicationSignalingClient {
...
+    /// Called from getSkypetoken, handle token errors here.
+    var tokenRefreshHandler: TrouterTokenRefreshHandler
+
+    /// Current number of token fetch retries.
+    var tokenRetries: Int
 
-    init(skypeToken: String? = nil) {
-        self.skypeToken = skypeToken
+    /// Max number of token retries allowed.
+    let MAX_TOKEN_RETRIES: Int = 3
```
tjprescott, line 127: This is not idiomatic Swift for constants. It should just be `let maxTokenRetries`.

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
```diff
@@ -106,13 +111,64 @@ class CommunicationSignalingClient {
...
+                return token
+            }
+
+            // Fetch new token
+            credential.token { token, error in
+                // Let callback know we are attempting to refresh
+                self.tokenRefreshHandler(false)
+                guard let newToken = token?.token else {
+                    return
+                }
```
tjprescott, line 147: Why are you not doing anything with the error here? If token is nil, presumably there would be an error.

sdk/core/AzureCore/CHANGELOG.md
```diff
@@ -1,8 +1,12 @@
 # Release History
+
+## 1.0.0-beta.13 (Unreleased)
+### Bugs Fixed
+- Changed the format for application ID in `UserAgentPolicy` to remove the square brackets around it.  
+
```
tjprescott, line 6: This should not be part of your PR. This is the change on main.

tjprescott (changes requested):
There is something wrong with the rebase here. It includes changes to AzureCore which were already merged. Some other comments.

sdk/communication/AzureCommunicationChat/Source/Generated/Options/AzureCommunicationChatClientOptionsInternal.swift
knvsl: I think the issue is the generated options are used by other generated code (ChatClientInternal), so I can't remove it completely. next time the code is generated we'd have to keep deleting the generated version

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
knvsl: yes its internal

sdk/communication/AzureCommunicationChat/Package.swift
knvsl: Updated, thanks

sdk/communication/AzureCommunicationChat/Source/Signaling/CommunicationSignalingClient.swift
knvsl: Updated, to pass the error to chatClient

sdk/core/AzureCore/CHANGELOG.md
knvsl: reverted the cherry-pick, we won't worry about including this change in this particular patch

tjprescott (approved):

PR #1047, opened 2021-09-30
Add AzureTest as dependency in Package.swift

knvsl:


sdk/communication/AzureCommunicationChat/Package.swift
```diff
@@ -49,6 +49,11 @@ let package = Package(
             name: "Trouter",
             url: "https://github.com/microsoft/trouter-client-ios.git",
             .exact("0.1.0")
+        ),
+        .package(
+            name: "AzureTest",
+            url: "https://github.com/Azure/azure-sdk-for-ios.git",
+            .branch("main")
```
knvsl, line 56: Not sure if this is correct - is there version I should specify?

tjprescott: Nope, this is correct.

tjprescott (approved):

Issue #1124, opened 2022-01-13
Update SwiftPM mirror repos so the code reflects the latest code


angellan-msft:
**Describe the bug**
For SwiftPM-AzureCommunicationChat mirror repository, releases are tagged but the repo itself has not been updated, this can be misleading and confusing for people who visit that page. We need to update the actual repository with the latest release.

***Exception or Stack Trace***
N/A

**To Reproduce**
N/A

***Code Snippet***
N/A

**Expected behavior**
We expected the code of SPM mirror repository master branch to be updated to our latest release.

**Screenshots**
![image](https://user-images.githubusercontent.com/95160885/149241570-849c0fb0-cede-4cdb-80ba-c1b0533db38b.png)
From the above image we can see all codes are in "1.0.0-beta.12". 

![image](https://user-images.githubusercontent.com/95160885/149241946-ffd13ed1-2560-49e4-8db9-34e255439fd6.png)
The tags show that the latest release is "1.1.0-beta.2", which is newer than the repo itself.

**Setup (please complete the following information):**
N/A

**Additional context**
N/A

**Information Checklist**
Kindly make sure that you have added all the following information above and checkoff the required fields otherwise we will treat the issuer as an incomplete report
- [X] Bug Description Added
- [ ] Repro Steps Added
- [ ] Setup information Added

tjprescott:
This is true of _all_ the mirror repos and the change should apply uniformly.

benbp:
After talking with @tjprescott, I think the best resolution to this is for us to replace all code in the master branch with a small README that explains the mirror repo and redirects people to [azure-sdk-for-ios](https://github.com/Azure/azure-sdk-for-ios/) repository for things like issues and pull requests.

We'll need some sort of common readme, perhaps in the main repo, and a sync process for updating the mirror repositories or adding new ones.

@weshaggard thoughts?

weshaggard:
I'm fine with that as an approach. Let's make sure the partners are good with that.

angellan-msft:
I’m fine with it as well. Thank you.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows

benbp:
@angellan-msft this should update for the communication repos on the next release/sync to them, but let us know if you'd like us to manually update sooner.

PR #1138, opened 2022-01-26
Add release step to publish redirect README to master branch of SwiftPM repository

benbp:
Resolves https://github.com/Azure/azure-sdk-for-ios/issues/1124

### DIFF
--- a/eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
+++ b/eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
@@ -1,0 +1,56 @@
+param (
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string] $GitSourcePath,
+
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string] $GitDestinationPath,
+
+    [Parameter(Mandatory = $false)]
+    [ValidateNotNullOrEmpty()]
+    [string] $UpstreamBranch = "master",
+
+    [Parameter(Mandatory = $false)]
+    [ValidateNotNullOrEmpty()]
+    [string] $Repository = "azure"
+)
+
+$ErrorActionPreference = "Stop"
+
+. (Join-Path $PSScriptRoot/../common/scripts common.ps1)
+
+Push-Location $GitDestinationPath
+
+git checkout $UpstreamBranch
+
+$repoContents = (Get-ChildItem -Attributes Hidden,!Hidden -Exclude ".git").Name
+# If there is a diff from a single README.md file, clear the repo
+if (Compare-Object $repoContents @("README.md")) {
+    $repoContents | Remove-Item -Recurse -Force
+    LogDebug "Adding commit to clear $UpstreamBranch branch contents"
+    git add -u
+    git commit -m "Clear $UpstreamBranch branch contents"
+}
+
+Copy-Item -Path $GitSourcePath/eng/mirror.README.md -Destination README.md
+$repoUrl = git config --get remote.origin.url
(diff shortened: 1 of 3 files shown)


benbp:
@tjprescott do you know how we can double check any other partner teams on this change?

benbp:
@weshaggard is there a way to exclude template links from the link checker? Or should I move the entire markdown link block into the render statement?

weshaggard:
The way to turn off is to add an ignorelinks.txt, however you might also consider commenting out instead of having it look like a valid link.

tjprescott:
@benbp I think you've already cleared this with the only partner team. The other libraries we publish are owned by our team.

tjprescott (approved):

benbp:
> The way to turn off is to add an ignorelinks.txt, however you might also consider commenting out instead of having it look like a valid link.

I'm not quite sure I understand what you mean? The template text does not look like a link, but rather it's because it's in a markdown `[]()` link block. How would I comment that out without affecting the rendered markdown?

eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
```diff
@@ -0,0 +1,56 @@
...
+    [ValidateNotNullOrEmpty()]
+    [string] $GitSourcePath,
+
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string] $GitDestinationPath,
+
+    [Parameter(Mandatory = $false)]
+    [ValidateNotNullOrEmpty()]
+    [string] $UpstreamBranch = "master",
```
weshaggard, line 12: Should these be updated to be main?

eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
```diff
@@ -0,0 +1,56 @@
...
+    git commit -m "Clear $UpstreamBranch branch contents"
+}
+
+Copy-Item -Path $GitSourcePath/eng/mirror.README.md -Destination README.md
+$repoUrl = git config --get remote.origin.url
+# Convert ssh remotes to https urls
+if ($repoUrl -imatch ".*($Repository/SwiftPM-.*)") {
+    $repoUrl = "https://github.com/$($matches[1])"
+}
+$readme = cat README.md
```
weshaggard, line 42: Probably better to use Get-Content so we don't end up using different tools on different OS's.

eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
```diff
@@ -0,0 +1,56 @@
+param (
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string] $GitSourcePath,
```
weshaggard, line 4: Looks like you only use the readme file, so might be worth just passing in that path instead of the entire repo path.

eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
benbp: It seems like the swift package manager has no assumptions here, so yes. That said, while branch based dependencies aren't recommended, we are technically removing one possible swift dependency mode ([docs](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html)) with this change. @tjprescott any thoughts on this or are we ok enforcing version/commit based dependencies?

eng/scripts/Publish-SwiftMirrorRedirectReadme.ps1
benbp: I'll do a follow-on PR with these updates.

PR #1231, opened 2022-04-07
Add trouter events

LuChen-Microsoft:


angellan-msft (approved):

tjprescott (changes requested):
Holding this until I have time to review.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -99,6 +99,52 @@ public class ChatClient {
         )
 
         self.service = client.chat
+        self.signalingClient = try createSignalingClient(credential: credential)
+    }
```
tjprescott, line 103: Do you want client creation to fail if a signaling client cannot be created? What if they don't care about receiving realtime notifications? This seems like a significant change.

tjprescott (changes requested):

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -99,6 +99,52 @@ public class ChatClient {
...
+                signalingClient = try CommunicationSignalingClient(
+                    communicationSkypeTokenProvider: tokenProvider,
+                    logger: self.options.logger
+                )
+            } catch {
+                self.options.logger.error("Signaling client initialization failed.")
+            }
+        }
+        return signalingClient
     }
```
tjprescott, line 148: This looks fundamentally wrong as an async pattern. Line 147 will be executed immediately before the credential token closure completes. As an async function this needs a completion handler.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -235,64 +281,25 @@ public class ChatClient {
...
+
+        signalingClientStarted = true
+        communicationSignalingClient.start()
     }
 
     /// Stop receiving realtime notifications.
     /// This function would unsubscribe to all events.
     public func stopRealTimeNotifications() {
-        if signalingClient == nil {
+        guard let communicationSignalingClient = signalingClient else {
```
tjprescott, line 296: As FYI you don't have to rename this. This is perfectly fine:
```swift
guard let signalingClient = signalingClient else { ... }
```

sdk/communication/AzureCommunicationChat/Source/Signaling/Events/TrouterEvent.swift
```diff
@@ -47,7 +50,20 @@ public enum TrouterEvent {
     /// - Parameters:
     ///   - chatEventId: The ChatEventId.
     ///   - request: The TrouterRequest that contains the event data.
-    init(chatEventId: ChatEventId, from request: TrouterRequest) throws {
+    init(chatEventId: ChatEventId, from trouterRequest: TrouterRequest? = nil) throws {
```
tjprescott, line 53: I'd recommend not making `TrouterRequest?` have a `nil` default. You should be force to pass something here, and if it happens to be nil, so be it.

sdk/communication/AzureCommunicationChat/Source/Signaling/Events/TrouterEventUtil.swift
```diff
@@ -112,7 +112,7 @@ internal enum TrouterEventUtil {
     /// - Returns: A TrouterEvent.
     internal static func create(
         chatEvent chatEventId: ChatEventId,
-        from request: TrouterRequest
+        from request: TrouterRequest? = nil
```
tjprescott, line 115: Same. Recommend dropping the default.

tjprescott (changes requested):

tjprescott:
Additionally, please show the APIView that reflects this change.

LuChen-Microsoft:
> Additionally, please show the APIView that reflects this change.

https://apiview.dev/Assemblies/Review/7b285d2a43b94e83b7adb3681185e7d7

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -38,6 +38,8 @@ public class ChatClient {
     private let service: Chat
     private var signalingClient: CommunicationSignalingClient?
     private var signalingClientStarted: Bool = false
+    private var realTimeNotificationConnectedHandler: TrouterEventHandler?
+    private var realTimeNotificationDisconnectedHandler: TrouterEventHandler?
```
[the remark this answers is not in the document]
tjprescott: Ping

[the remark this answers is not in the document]
tjprescott: Since this is internal implementation, it can always be changed later.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -303,15 +315,35 @@ public class ChatClient {
...
+
+            if event == ChatEventId.realTimeNotificationDisconnected {
+                realTimeNotificationDisconnectedHandler = handler
+                return
+            }
+
             options.logger
                 .warning(
-                    "Signaling client is not initialized, cannot register handler. Ensure startRealtimeNotifications() is called first."
+                    "Signaling client is not initialized, cannot register handler."
```
tjprescott, line 331: Out of curiosity, does this actually have to be true? Couldn't you use the same pattern as with RT notification started/stopped and simply store the handlers on the ChatClient and then "connect" them once the RT notifications have started? 

It would be a non-breaking change to add this in the future, but worth considering.

sdk/communication/AzureCommunicationChat/Source/Signaling/Events/TrouterEvent.swift
```diff
@@ -65,7 +65,9 @@ public enum TrouterEvent {
         }
 
         switch chatEventId {
-        case ChatEventId.chatMessageReceived:
+        case .realTimeNotificationConnected, .realTimeNotificationDisconnected:
+            throw AzureError.client("Chat Event ID should not be \(chatEventId)")
+        case .chatMessageReceived:
```
tjprescott, line 70: Again, AzureError here implies that the customer can do something about this, which they can't. If this code is reached, it could only be due to a bug in the above code since it should exit by this point.
My recommendation would be to just do a logger.error here and return the relevant event. Because that's what you would do if lines 54-65 were deleted (which you totally could do if you moved the guard statement.

LuChen-Microsoft: It seems not worth adding the logger to the file. I changed to ignore the bug cases.

LuChen-Microsoft:
Verified the change.

tjprescott (approved):

tjprescott (changes requested):
Actually, one lingering issue--you need a Changlog entry.

sdk/communication/AzureCommunicationChat/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release History
 
+## 1.2.0 (Unreleased)
+### New Features
```
tjprescott, line 4: https://azure.github.io/azure-sdk/policies_releases.html#change-logs
This does not align with the changelog guidance. This should say "Features Added". Same for line 15.

tjprescott (changes requested):

sdk/communication/AzureCommunicationChat/CHANGELOG.md
LuChen-Microsoft: fixed them

tjprescott (approved):

PR #1154, opened 2022-02-08
Enable Push Notifications in Chat SDK

angellan-msft:

PR #1236, opened 2022-04-13
Chat Push Notification

angellan-msft:
I created a new branch for Chat Push Notification. 
This branch is based on Chat 1.1.0 GA release and contains all the code changes in the old PN PR (https://github.com/Azure/azure-sdk-for-ios/pull/1154).

APIView: https://apiview.dev/Assemblies/Revisions/993b3da970c24d24a05f035d674b7525


sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+internal enum RegistrarSettings {
```
tjprescott, line 29: Why is this an enum?

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
```diff
@@ -0,0 +1,56 @@
...
+    static let languageId = ""
+    static let platform = "iOS"
+    static let platformUIVersion = "3619/0.0.0.0/"
+    static let templateKey = "AcsIos.AcsNotify_Chat_3.0"
+
+    // Transport Settings
+    static let pushNotificationTransport = "APNS"
+    // Max TTL is 180 days
+    static let ttl = 15_552_000
+    static let context = ""
```
tjprescott, line 44: These are all constants, but where do these values actually come from? Why are they all static? Why are there empty strings (which cannot be modified since they are `let`)?

In short, I don't understand the need for this object.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarClientDescription.swift
```diff
@@ -0,0 +1,59 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+/// Client description for set registration requests.
+internal struct RegistrarClientDescription: Codable {
```
tjprescott, line 30: This seems redundant with `RegistrarSettings` except that this one is actually customizable. How are these used together?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -330,4 +338,109 @@ public class ChatClient {
...
+    }
+
+    /// Handle the data payload for an incoming push notification.
+    /// - Parameters:
+    ///   - notification: The APNS push notification payload ( including "aps" and "data" )
+    ///   - completionHandler:The event handler to handle the PushNotificationEvent
+    public func handlePush(
+        notification: [AnyHashable: Any],
+        completionHandler: PushNotificationEventHandler
+    ) {
```
tjprescott, line 403: Again, `completionHandler` isn't really a completion handler and this method does not make sense and causes a lot of internal implementation to be exposed as public.

Why are you not following the pattern already used for realtime notifications where you register and unregister closures for specific events? Reviewing that pattern, it does not seems to unnecessarily expose any internal implementation.

tjprescott (changes requested):

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationClient.swift
```diff
@@ -0,0 +1,134 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import AzureCommunicationCommon
+import AzureCore
+import Foundation
+
+/// ChatClient class for ChatThread operations.
+public class PushNotificationClient {
```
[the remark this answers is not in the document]
angellan-msft: Good catch. I've fixed it to internal.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
angellan-msft: This class contains the static values required in each registration request.
Based on the description of Registrar API ([EDFRegistrar Notebook (sharepoint.com)](https://microsoft.sharepoint.com/teams/EDFRegistrar/_layouts/15/Doc.aspx?sourcedoc=%7Bd54f197a-d828-4c78-bb08-42eb962b8211%7D&action=view&wd=target%28Home.one%7C03f29ff8-34f5-4e72-b688-7e0f89984628%2FRegistrar+API%7C53b2b5b3-3b4a-419d-8c41-69b82ec89ae7%2F%29&isSPOFile=1)),
we will need to include "nodeId", "clientDescription" and "transports" in our registration request. Most of them are static and pre-defined. Some fields are required although they have empty values - that's why you see some empty strings.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
angellan-msft: Please see the explanation for this class in the below comment.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarClientDescription.swift
angellan-msft: RegistrarClientDescription is the model class we use to init a ClientDescription object. RegistrarSettings contains the pre-defined parameters that should be put in the initializer.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
tjprescott: These constants should either be default values in the `RegistrarClientDescription` initializer or, if they are truly required constants, not in the initializer at all but simply defined on the `RegistrarClientDescription`.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarClientDescription.swift
tjprescott: Per discussion above, these objects should be merged together.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
angellan-msft: Sure. And do we want to keep these two classes?

internal enum RegistrarHeader: String {
    /// Content-type header.
    case contentType = "Content-Type"
    /// Skype token for authentication.
    case skypeTokenHeader = "X-Skypetoken"
}

internal enum RegistrarMimeType: String {
    case json = "application/json"
}

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
angellan-msft: OK I put them in the initializer.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -99,6 +103,10 @@ public class ChatClient {
         )
 
         self.service = client.chat
+        self.pushNotificationClient = PushNotificationClient(
+            credential: credential,
+            options: options
+        )
```
tjprescott, line 109: What if the user does not want or care about push notifications? Why are we creating a push notification client? If creating the client fails for some reason it means that all ChatClient functionality is closed to users.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -330,4 +338,109 @@ public class ChatClient {
...
+    /// - Parameters:
+    ///   - deviceToken: APNS push token.
+    ///   - completionHandler: Success indicates request to register for notifications has been received.
+    public func startPushNotifications(
+        deviceToken: String,
+        completionHandler: @escaping (Result<HTTPResponse?, AzureError>) -> Void
+    ) {
+        guard !pushNotificationsStarted else {
+            completionHandler(.failure(AzureError.client("Push notifications already started.")))
+            return
```
tjprescott, line 352: Questionable whether you want this to be an error. If you say "startNotifications" and they have already been started, a logical option would be to treat that as success. I'm not saying this has to change, but as error here seems... at least worthy of questioning.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -330,4 +338,109 @@ public class ChatClient {
...
+            completionHandler(nil, AzureError.client("decoding error"))
+        } catch let DecodingError.typeMismatch(type, context) {
+            print("Type '\(type)' mismatch:", context.debugDescription)
+            print("codingPath:", context.codingPath)
+            completionHandler(nil, AzureError.client("decoding error"))
+        } catch {
+            print("error: ", error)
+            // Pass the error into the contoso event handler
+            completionHandler(nil, error)
+        }
```
tjprescott, line 444: All of these print statements need to go.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarSettings.swift
angellan-msft: I removed "RegistrarSettings" and kept "RegistrarHeader" & "RegistrarMimeType".

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: This has been discussed in our meeting.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: Will remove the whole "handlePush()" API in this PR.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -38,6 +38,9 @@ public class ChatClient {
     private let service: Chat
     private var signalingClient: CommunicationSignalingClient?
     private var signalingClientStarted: Bool = false
+    private var pushNotificationClient: PushNotificationClient
+    private var pushNotificationsStarted: Bool = false
+    internal var registrationId: String
```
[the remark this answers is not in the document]
angellan-msft: From what I can remember, we decided to put "registrationId" in ChatClient to pass the ChatClientDVR test.

On 2/25/2022's chat discussion, you suggests that - "You could actually have ChatClient generate the GUID when you create the client, and then when you call startNotifications it uses that and generates a new one. It has the same effect but then you could change the value to something deterministic in your test and register the replacement."

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: Good question. Android initializes the PN client inside the Chat Client initialization process while RT Notification doesn't. Since PN and RT notification are in the same SDK. I think it is better to expect PN client has the same behaviour as RT client. So I will set PN Client as optional as well and initialize it when the contuse calls the related APIs.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -38,6 +38,9 @@ public class ChatClient {
     private let service: Chat
     private var signalingClient: CommunicationSignalingClient?
     private var signalingClientStarted: Bool = false
+    private var pushNotificationClient: PushNotificationClient
```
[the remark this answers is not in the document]
angellan-msft: Will set the PN client as optional as well.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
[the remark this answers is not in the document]
angellan-msft: And for "pushNotificationsStarted", we need this external to do some "pre-validation". For example, when Contoso calls chatClient.startPushNotification() and finds the PN has already started, Contoso does not need to initialize the PushNotificationClient at all.

sdk/communication/AzureCommunicationChat/Source/Registrar/RegistrarClientDescription.swift
angellan-msft: Merged.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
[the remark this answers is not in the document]
tjprescott: I think my advice from 2/25 still applies if the registrationId lives in the pushNotificationClient. For the second part, it would still work if `started` was part of the notification client.
```swift
func startPushNotification(...) {
  guard self.pushNotificationClient?.started == false else {
    "Push notifications already started".
  }
}
```

etc.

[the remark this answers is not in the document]
angellan-msft: Yes you are right. I put both of them inside PN client then.

tjprescott (approved):

tjprescott (changes requested):
Sorry, one thing missing -- Changelog entry!

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
tjprescott: It's worth considering whether RT notifications should be updated accordingly. But in a separate PR :)

tjprescott (changes requested):
Please retarget the PR to a feature branch.

tjprescott (approved):

PR #1244, opened 2022-04-22
[Bug] Thread wait moving outside of the dispatch queue

JoshuaLai:


tjprescott:
@JoshuaLai is there an issue associated with this fix?

JoshuaLai:
> @JoshuaLai is there an issue associated with this fix? If this bug affects token refresh is might be more prevalent in Azure Core. Can you provide some more info?

Spoke about this offline. I think we may want to do a deeper dive into this implementation. Going to merge this for now.

PR #1255, opened 2022-05-05
[CommunicationIdentifier] Introduce new rawId and creation of identifer

JoshuaLai:
The `CommunicationIdentifier` design provides a good abstraction of Azure Communication Services internal id format with better type safety, auto-complete and hides internal knowledge.

Create new `rawId` in `CommunicationIdentifier` protocol 
Introduce `CommunicationIdentifierHelper` help deserialize the mri and convenient way to create appropriate identifiers

Details [Internal wiki](https://skype.visualstudio.com/SPOOL/_wiki/wikis/SPOOL.wiki/31075/Design-change-request-RawId-support-for-(de)serialization)

### DIFF
--- a/sdk/communication/AzureCommunicationCommon/CHANGELOG.md
+++ b/sdk/communication/AzureCommunicationCommon/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Release History
 
-## 1.0.4 (upcoming)
+## 1.1.0 (upcoming)
 ### New Features
 - Introduce new `CommunicationIdentifierHelper` to create `CommunicationIdentifier` based on rawIds
 - Introduce new `rawId` getter property in the `CommunicationIdentifier` protocol 
+
+### Key Bug Fixes 
+- `PhoneNumberIdentifier` no longer allows the `rawId` to not correspond with the phone number
 
 ## 1.0.3 (2022-03-10)
 ### New Features
--- a/sdk/communication/AzureCommunicationCommon/CHANGELOG.md
+++ b/sdk/communication/AzureCommunicationCommon/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Release History
 
+## 1.0.4 (upcoming)
+### New Features
+- Introduce new `CommunicationIdentifierHelper` to create `CommunicationIdentifier` based on rawIds
+- Introduce new `rawId` getter property in the `CommunicationIdentifier` protocol 
+
 ## 1.0.3 (2022-03-10)
 ### New Features
 - Update `AzureCore` dependency to beta 15.
--- a/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
+++ b/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
@@ -25,9 +25,17 @@
 // --------------------------------------------------------------------------
 
 import Foundation
-
+/**
+ Helper class to easily create CommunicationIdentifiers
+ */
 @objcMembers
 public class CommunicationIdentifierHelper: NSObject {
+    /**
+     Creates a CommunicationIdentifier from a given rawId. When storing rawIds use this function to restore the identifier that was encoded in the rawId.
+     Parameters: rawId The rawId to be translated to its identifier representation.
+     Returns: Type safe CommunicationIdentifier created. Use the `isKind(of:)` to verify  identifier type
+     SeeAlso: ` CommunicationIdentifier`
+     */
     public static func createCommunicationIdentifier(from rawId: String) -> CommunicationIdentifier {
         let phoneNumberPrefix = "4:"
         let teamUserAnonymousPrefix = "8:teamsvisitor:"
--- a/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
+++ b/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
@@ -30,6 +30,15 @@ import Foundation
  */
 @objcMembers
 public class CommunicationIdentifierHelper: NSObject {
+    private static let phoneNumberPrefix = "4:"
+    private static let teamUserAnonymousPrefix = "8:teamsvisitor:"
+    private static let teamUserPublicCloudPrefix = "8:orgid:"
+    private static let teamUserDODCloudPrefix = "8:dod:"
+    private static let teamUserGCCHCloudPrefix = "8:gcch:"
+    private static let acsUser = "8:acs:"
+    private static let spoolUser = "8:spool:"
+    private static let dodAcsUser = "8:dod-acs:"
+    private static let gcchAcsUser = "8:gcch-acs:"
     /**
      Creates a CommunicationIdentifier from a given rawId. When storing rawIds use this function to restore the identifier that was encoded in the rawId.
      Parameters: rawId The rawId to be translated to its identifier representation.
@@ -37,29 +46,16 @@ public class CommunicationIdentifierHelper: NSObject {
      SeeAlso: ` CommunicationIdentifier`
      */
     public static func createCommunicationIdentifier(from rawId: String) -> CommunicationIdentifier {
-        let phoneNumberPrefix = "4:"
-        let teamUserAnonymousPrefix = "8:teamsvisitor:"
-        let teamUserPublicCloudPrefix = "8:orgid:"
-        let teamUserDODCloudPrefix = "8:dod:"
-        let teamUserGCCHCloudPrefix = "8:gcch:"
-        let acsUser = "8:acs:"
-        let spoolUser = "8:spool:"
-        let dodAcsUser = "8:dod-acs:"
-        let gcchAcsUser = "8:gcch-acs:"
-
         if rawId.hasPrefix(phoneNumberPrefix) {
             let phoneNumber = "+" + rawId.dropFirst(phoneNumberPrefix.count)
             return PhoneNumberIdentifier(phoneNumber: phoneNumber, rawId: rawId)
         }
-
         let segments = rawId.split(separator: ":")
         if segments.count < 3 {
             return UnknownIdentifier(rawId)
         }
-
         let scope = segments[0] + ":" + segments[1] + ":"
         let suffix = String(rawId.dropFirst(scope.count))
-
         switch scope {
         case teamUserAnonymousPrefix:
             return MicrosoftTeamsUserIdentifier(userId: suffix, isAnonymous: true)
--- a/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
+++ b/sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
@@ -1,0 +1,81 @@
+// --------------------------------------------------------------------------
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+//
+// The MIT License (MIT)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the ""Software""), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+@objcMembers
+public class CommunicationIdentifierHelper: NSObject {
+    public static func createCommunicationIdentifier(from rawId: String) -> CommunicationIdentifier {
+        let phoneNumberPrefix = "4:"
+        let teamUserAnonymousPrefix = "8:teamsvisitor:"
+        let teamUserPublicCloudPrefix = "8:orgid:"
+        let teamUserDODCloudPrefix = "8:dod:"
+        let teamUserGCCHCloudPrefix = "8:gcch:"
+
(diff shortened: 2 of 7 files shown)


sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
```diff
@@ -75,17 +80,36 @@ import Foundation
...
+    /**
+     Returns a Boolean value indicating whether two values are equal.
+        Note: In Objective-C favor isEqual() method
+     - Parameter lhs PhoneNumberIdentifier to compare.
+     - Parameter rhs  Another PhoneNumberIdentifier to compare.
+     */
+    // swiftlint:disable nsobject_prefer_isequal
     public static func == (lhs: PhoneNumberIdentifier, rhs: PhoneNumberIdentifier) -> Bool {
-        if lhs.phoneNumber != rhs.phoneNumber {
+        return lhs.rawId == rhs.rawId
```
JoshuaLai, line 99: If we make this change we will be changing some existing behaviour. 
Are we ok with this? 
Is this considered a breaking change?

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
```diff
@@ -114,31 +138,65 @@ import Foundation
...
+                            the id is the AAD object id of the user.
+     - Parameter isAnonymous: Set this to true if the user is anonymous:
+                                for example when joining a meeting with a share link.
+     */
     public init(userId: String, isAnonymous: Bool) {
         self.cloudEnviroment = .Public
-        self.rawId = nil
         self.userId = userId
         self.isAnonymous = isAnonymous
+        self.rawId = isAnonymous ? "8:teamsvisitor:" + userId : "8:orgid:" + userId
```
petrsvihlik, line 176: I suggest we keep the logic for constructing the `rawId` at a single place (right now it's partially duplicated in two constructors). 
plus, not sure if it's idiomatic for swift, but I'd prefer to have the logic outside of the constructor...or at least extracted to a method.

petrsvihlik (changes requested):

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
JoshuaLai: Unfortunately in swift because `rawId` is required. I cannot have a helper function called in the constructor until all the  properties of the class have been assigned. If we wanted to move it into a helper function it would need to be a static method and I don't think that is the right approach

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
```diff
@@ -114,31 +138,65 @@ import Foundation
         rawId: String? = nil,
         cloudEnvironment: CommunicationCloudEnvironment = .Public
     ) {
-        self.rawId = rawId
```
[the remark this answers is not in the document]
JoshuaLai: I'm not entirely sure what you mean here? Do you have an example of how you think it should be implemented? 
I can set up a convenience init but I am not sure how it will work in Objective-C, if it will work at all. I would need to test it out

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
```diff
@@ -0,0 +1,87 @@
...
+    func test_createPhoneNumberIdentifier() {
+        var phoneNumberIdentifier = CommunicationIdentifierHelper.createCommunicationIdentifier(from: "4:11231234")
+        XCTAssertTrue(phoneNumberIdentifier.isKind(of: PhoneNumberIdentifier.self))
+        phoneNumberIdentifier = CommunicationIdentifierHelper.createCommunicationIdentifier(from: "4:11231234:asdf")
+
+        XCTAssertTrue(phoneNumberIdentifier.isKind(of: PhoneNumberIdentifier.self))
+    }
+
+    func test_createCommunicationUserIdentifier() {
+        let acsScope = "8:acs:some_raw_id"
```
DominikMe, line 52: nit: this isn't a scope but the full rawId

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
```diff
@@ -0,0 +1,87 @@
...
+        var communciationUserIdentifier = CommunicationIdentifierHelper.createCommunicationIdentifier(from: acsScope)
+        XCTAssertTrue(communciationUserIdentifier.isKind(of: CommunicationUserIdentifier.self))
```
DominikMe, line 54: Can we do a full object comparison?

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
```diff
@@ -0,0 +1,87 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import AzureCommunicationCommon
+import XCTest
+
+class CommunicationIdentifierHelperTests: XCTestCase {
```
DominikMe, line 30: I had thought we could have the exact same test cases for each language with the exact same inputs and outputs instead of having an own set per language. For creating identfier from raw id, getting raw if from identifier and roundtrip.

What do you think?
JS tests: https://github.com/Azure/azure-sdk-for-js/pull/21699/files#diff-1145d95d7955948a6d8eee7b691a65bc6c7aeb65f3a135c5eb47dcaa224eaae6

sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
```diff
@@ -0,0 +1,89 @@
...
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+/**
+ Helper class to easily create CommunicationIdentifiers
+ */
+@objcMembers
+public class CommunicationIdentifierHelper: NSObject {
```
DominikMe, line 32: I'm never a fan of anything that ends in `Helper`, `Util` or `Manager` :-).

@tjprescott Any suggestions for good Swift-idiomatic naming?

Ideally we would have liked a static factory method on `CommunicationIdentifier` itself but can't do it because it's a protocol. Another alternative is a free function.

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
JoshuaLai: Are you thinking comparing every property inside `communciationUserIdentifier` and verifying that way

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
petrsvihlik: It would definitely be easier to manage in the long run. I also used Dom's set of tests in C#.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
petrsvihlik: are you talking about the case that someone initializes the `PhoneNumberIdentifier` with a `phoneNumber` and a non-corresponding `rawId`?

petrsvihlik: not sure if we want to support that case at all. perhaps we could add a validation to the constructor to avoid this case.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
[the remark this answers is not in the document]
petrsvihlik: I was thinking something like:

```swift
convenience init(userId: String, isAnonymous: Bool) {
     self.init(userId, isAnonymous, nil, .Public)
}

init(
        userId: String,
        isAnonymous: Bool = false,
        rawId: String? = nil,
        cloudEnvironment: CommunicationCloudEnvironment = .Public
    ) {
  /* set all properties, do the rawId serialization*/
}
```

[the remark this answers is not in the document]
petrsvihlik: >but I am not sure how it will work in Objective-C


seems to be possible
https://stackoverflow.com/a/36570905/1332034

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
petrsvihlik: I see. Then perhaps the convenience init could potentially make the code simpler.

sdk/communication/AzureCommunicationCommon/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Release History
 
+## 1.0.4 (upcoming)
+### New Features
+- Introduce new `CommunicationIdentifierHelper` to create `CommunicationIdentifier` based on rawIds
+- Introduce new `rawId` getter property in the `CommunicationIdentifier` protocol
```
tjprescott, line 6: This seems more like a 1.1.0 than a patch.

sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
tjprescott: Why not add a required init to the protocol that takes a rawId and then provide the contents of `createCommunicationIdentifier(from:)` as a default implementation via extension?

tjprescott: But I agree you don't really want this helper thingy as part of the public API.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
JoshuaLai: `are you talking about the case that someone initializes the PhoneNumberIdentifier with a phoneNumber and a non-corresponding rawId?`
Yes. We previously had unit tests to validate this one case. So as per this new PR we would remove this case. 
`not sure if we want to support that case at all. perhaps we could add a validation to the constructor to avoid this case.`
Yes this PR is make this change so it will no longer be possible

tjprescott: @JoshuaLai I think the archboard would definitely consider changing the behavior a breaking change... _unless_ the original behavior was considered a bug :)

tjprescott (changes requested):

sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
JoshuaLai: That was something I was trying. I initially was thinking something like this 
```
@objc public protocol CommunicationIdentifier: NSObjectProtocol {
    @objc var rawId: String { get }
    static func createCommunicationIdentifier(from rawId: String) -> CommunicationIdentifier
}
extension CommunicationIdentifier {
    public static func createCommunicationIdentifier(from rawId: String) -> CommunicationIdentifier {
        
    }
}
```
But because the `CommunicationIdentifier` needed to be exposed to Obj-C I can't create a function inside my extension nicely. 
Otherwise I would be forcing the identifiers to all require this method and every method would do the exact same thing.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
```diff
@@ -75,17 +80,36 @@ import Foundation
      */
     public init(phoneNumber: String, rawId: String? = nil) {
         self.phoneNumber = phoneNumber
-        self.rawId = rawId
+        if let rawId = rawId {
+            self.rawId = rawId
+        } else {
+            let strippedPhoneNumber = phoneNumber.replacingOccurrences(of: "+", with: "")
+            self.rawId = "4:" + strippedPhoneNumber
+        }
```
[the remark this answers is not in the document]
JoshuaLai: Previously we allowed developers to initialize without a `rawId` and it would just be optional. 
Now we are updating the logic, in case they do or formerly passed in `nil` we will still create a `rawId` on their behalf using the phone number that is passed in.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
DominikMe: We had supported to override the `phoneNumber` with a `rawId` previously and if rawId wsa provided the phoneNumber doesn't matter.
> unless the original behavior was considered a bug :)
So imo this is the case, and the change now is a bug fix.

> allowing the rawId to not correspond with the phone number almost seems like a bug to me.
It is a feature by design unfortunately because we were made aware of by our service teams that there are phone numbers that have a different format (no 4:).

sdk/communication/AzureCommunicationCommon/Tests/CommunicationIdentifierHelperTests.swift
JoshuaLai: Updated the unit tests to align better with other platforms

sdk/communication/AzureCommunicationCommon/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Release History
 
+## 1.1.0 (upcoming)
+### New Features
+- Introduce new `CommunicationIdentifierHelper` to create `CommunicationIdentifier` based on rawIds
```
petrsvihlik, line 5: Would be nice to have a more detailed explanation of the change, similar to https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/communication/communication-common/CHANGELOG.md#features-added

petrsvihlik (changes requested):

sdk/communication/AzureCommunicationCommon/Source/CommunicationIdentifierHelper.swift
JoshuaLai: @tjprescott @DominikMe Had a discussion about this implementation. I think in terms of functionality this approah that we have with the helper most closely aligns with the other platforms. 
That being said, the only reason we are doing this is because of obj-c support. I do think we should slowly start exploring having 2 implementations. One design to favour swift, and this implementation to favour obj-c. We can have in the documentation to guide the developers which approach they should use.

sdk/communication/AzureCommunicationCommon/Source/Identifiers.swift
```diff
@@ -30,11 +30,15 @@ import Foundation
  Common Communication Identifier protocol for all Azure Communication Services.
  All Communication Identifiers conform to this protocol.
  */
-@objc public protocol CommunicationIdentifier: NSObjectProtocol {}
+@objc public protocol CommunicationIdentifier: NSObjectProtocol {
+    var rawId: String { get }
```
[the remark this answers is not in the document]
JoshuaLai: Another suggestion that was brought up by @tjprescott that I really like is including a new enum or string to represent the `kind`. @DominikMe  I know we have a similar concept in the TS library.

JoshuaLai:
Closing this PR in favour of (https://github.com/Azure/azure-sdk-for-ios/pull/1312)

PR #1283, opened 2022-06-06
Add PushNotificationEncryptionDelegate

angellan-msft:
Updated APIView: https://apiview.dev/Assemblies/Revisions/eaa100d11a8a49a59bc77abe4d8731d5


sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -373,13 +373,14 @@ public class ChatClient {
...
-    /// Start push notifications. Receiving of notifications can be expected ~1 second after successfully registering.
+    /// Start push notifications. Receiving of notifications can be expected after successfully registering.
     /// - Parameters:
     ///   - deviceToken: APNS push token.
+    ///   - pushNotificationEncryptionDelegate: Manage encyrption keys used by push notification.
     ///   - completionHandler: Success indicates request to register for notifications has been received.
     public func startPushNotifications(
         deviceToken: String,
-        encryptionKeyPair: EncryptionKeyPair,
+        pushNotificationEncryptionDelegate: PushNotificationEncryptionDelegate?,
```
tjprescott, line 383: The delegate should not be passed to these calls. It should be set on the client directly.

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/CryptoUtils.swift
```diff
@@ -142,15 +142,37 @@ func decryptPushNotificationPayload(cipherText: [UInt8], iv: [UInt8], cryptoKey:
...
+    return SymmetricKey(size: .init(bitCount: 512)).serialize()
+}
+
+// Use the SymmetricKey class in CryptoKit framework to create encryption keys
+extension SymmetricKey {
+    /// Serializes a `SymmetricKey` to a Base64-encoded `String`.
+    func serialize() -> String {
+        return withUnsafeBytes { body in
+            Data(body).base64EncodedString()
+        }
```
tjprescott, line 155: This seems, well, unsafe. What is the status of consulting the Crypto Board for their review?

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationClient.swift
```diff
@@ -61,7 +61,7 @@ internal class PushNotificationClient {
 
     internal func startPushNotifications(
         deviceRegistrationToken: String,
-        encryptionKeyPair: EncryptionKeyPair,
+        pushNotificationEncryptionDelegate: PushNotificationEncryptionDelegate?,
```
tjprescott, line 64: While this is internal, I would recommend following the same pattern as the public API and not setting the delegate in a method like this.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationClient.swift
```diff
@@ -74,21 +74,25 @@ internal class PushNotificationClient {
...
+                    let encryptionKey = generateEncryptionKey()
+
+                    // If the Contoso doesn't want to implement the advanced version of PN, they will not pass in chatPushNotificationEncryptionDelegate. We can register two hard-coded fake keys for them.
+                    if pushNotificationEncryptionDelegate == nil {
                         self.aesKey = "0000000000000000B00000000000000000000000AES="
                         self.authKey = "0000000000000000B0000000000000000000000AUTH="
                     } else {
-                        self.aesKey = firstKey
-                        self.authKey = secondKey
+                        pushNotificationEncryptionDelegate!.store(
```
tjprescott, line 85: Do not force unwrap in production code.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationEncryptionDelegate.swift
```diff
@@ -0,0 +1,33 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import Foundation
+
+public protocol PushNotificationEncryptionDelegate {
```
tjprescott, line 29: I feel like this should reflect "ChatClient" somehow the way UITableViewDelegate does for a UITableView. `ChatClientPushNotificationDelegate` perhaps. @mpodwysocki, suggestions?

This will also run up against the guideline that says delegates are optional but closures are required.

tjprescott (changes requested):
Additionally, you need an APIView to show the changes to the API surface area.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationEncryptionDelegate.swift
```diff
@@ -0,0 +1,33 @@
...
+    func store(encryptionKey: String, expiryTime: Date)
+
+    func getEncryptionKeys() -> [String]?
```
tjprescott, line 32: Delegates must follow the guidelines for delegates:
**https://azure.github.io/azure-sdk/ios_design.html#delegates**

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: Not sure if I understand this comment correctly. Could you please elaborate more on "Set on the client directly"?

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/CryptoUtils.swift
angellan-msft: In order to convert a Symmetric key into a base64EncodedString, we will need to access the underlying data of the Symmetric key first.
Based on Apple Official doc, [func withUnsafeBytes<R> -> R] is the only way to achieve it. From my understanding, this is how Apple named it.
https://developer.apple.com/documentation/cryptokit/symmetrickey

And we've confirmed with the PNH team and got a response that they had this feature reviewed by Crypto Board before.

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: Let me clarify my understanding - We provide a protocol "PushNotificationEncryptionDelegate" and the Contoso needs to create a concrete class to adhere this Protocol, for example "AppGroupBasedDelegate". They will need to initialize the "AppGroupBasedDelegate" and pass it into chat SDK. So Chat SDK can call it to "persist key" when starting PN and "get keys" when decrypting PN.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationEncryptionDelegate.swift
angellan-msft: Thanks for sharing. I'm reading it right now.

angellan-msft: I just read through the example in this wiki and did some research on "Delegate" in swift. 
For me "PushNotificationAdapter" is just like an interface and don't need to be specifically defined as a "delegate".
Could you please provide a reason on why we should treat it as a "Delegate"? @tjprescott

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
angellan-msft: Could you please provide a reason on why we should treat it as a "Delegate"? Are there any guidelines? @tjprescott

angellan-msft: I provided a delegate property on the client side so Contoso can set the delegate as soon as they init the ChatClient. Code change pushed.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationClient.swift
angellan-msft: I set the delegate in chatClient and it will not be passed into pushNotificationClient.

sdk/communication/AzureCommunicationChat/Source/PushNotification/ChatClientPushNotificationProtocol.swift
```diff
@@ -0,0 +1,148 @@
...
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+//
+// --------------------------------------------------------------------------
+
+import AzureCore
+import Foundation
+
+public protocol ChatClientPushNotificationProtocol: AnyObject {
+    func onPersistKey(_ encryptionKey: String, expiryTime: Date)
```
jiminwen-msft, line 31: Is this a placeholder?

sdk/communication/AzureCommunicationChat/Source/ChatClient.swift
```diff
@@ -373,13 +374,12 @@ public class ChatClient {
         }
     }
 
-    /// Start push notifications. Receiving of notifications can be expected ~1 second after successfully registering.
+    /// Start push notifications. Receiving of notifications can be expected after successfully registering.
     /// - Parameters:
     ///   - deviceToken: APNS push token.
     ///   - completionHandler: Success indicates request to register for notifications has been received.
     public func startPushNotifications(
```
jiminwen-msft, line 381: Just wondering is it better to have a separate API called renewRegistration()? So instead of the whole process being invoked by client every 15 minutes, we just generate/rotate the keys. Or could we have a flag recording is it the first time start push notification, and avoid duplicate creation of objests?

angellan-msft: Yea on Android side we have renewRegistration(), but this "renew feature" is not within the scope in this release for IOS. For the second part of the comments, we have a flag "pushNotificationsStarted" inside PushNotificationClient to indicate if the PN has started. And thanks for sharing thoughts on how to generate object in a more efficient way when doing auto-renew - I will mark this and discuss further when we consider implementing this feature in next iterations.

jiminwen-msft (approved):
LGTM

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/CryptoUtils.swift
```diff
@@ -142,15 +142,37 @@ func decryptPushNotificationPayload(cipherText: [UInt8], iv: [UInt8], cryptoKey:
...
+    }
+}
+
+func splitEncryptionKey(encryptionKey: String) -> [String] {
+    guard var data = Data(base64Encoded: encryptionKey, options: .ignoreUnknownCharacters) else {
+        return ["0000000000000000B00000000000000000000000AES=", "0000000000000000B0000000000000000000000AUTH="]
+    }
```
tjprescott, line 162: Why would this not be an error? It seems returning an empty key is a recipe for auth bugs.

sdk/communication/AzureCommunicationChat/Source/PushNotification/AppGroupPushNotificationKeyHandler.swift
```diff
@@ -0,0 +1,94 @@
...
+    var sharedDefault: UserDefaults
+    var keyTag: String
+
+    public init?(appGroupId: String, keyTag: String) {
+        guard let shareDefault = UserDefaults(suiteName: appGroupId) else {
+            return nil
+        }
+        self.sharedDefault = shareDefault
+        self.keyTag = keyTag
+    }
```
tjprescott, line 39: To me this seems like it should be a throwing initializer. If this fails, it gives no indication why, when the reason is pretty clear--the UserDefaults don't exist for the app group ID. It would be better to throw and give the user an actionable error. If they do want this behavior they can just do `let handler = try? AppGroupPushNotificationKeyHandler(...)`

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationKeyHandler.swift
```diff
@@ -0,0 +1,148 @@
...
+// --------------------------------------------------------------------------
+
+import AzureCore
+import Foundation
+
+public protocol PushNotificationKeyHandler: AnyObject {
+    func onPersistKey(_ encryptionKey: String, expiryTime: Date)
+
+    func onRetrieveKeys() -> [String]?
+}
```
tjprescott, line 34: You should definitely add docstrings here since this is public API and you need to give guidance on the format of these strings.

tjprescott (changes requested):
There are a few questions and suggestions left, but this is very close to being ready to merge.

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/CryptoUtils.swift
angellan-msft: FYI - "something unexpected happens" is based on the fact that the Contoso wants to implement "advanced version of PN" and has already set the PNKeyHandler.

If the Contoso just wants to implement "simple version" and doesn't want to worry about encryption/decryption things, that's a different story. We will not treat this scenario as an error. SDK will register two fake encryption keys for Contoso,

sdk/communication/AzureCommunicationChat/Source/PushNotification/AppGroupPushNotificationKeyHandler.swift
angellan-msft: agree. This is relevant to my answer to the previous comment. I will add more error handling logics in key management part.

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/CryptoUtils.swift
tjprescott: Yeah, this is the same rationale as above.

sdk/communication/AzureCommunicationChat/Source/PushNotification/PushNotificationKeyHandler.swift
angellan-msft: Change pushed. API View updated.

sdk/communication/AzureCommunicationChat/Source/PushNotification/AppGroupPushNotificationKeyHandler.swift
angellan-msft: Change pushed. API View updated.

sdk/communication/AzureCommunicationChat/Source/PushNotification/AppGroupPushNotificationKeyHandler.swift
```diff
@@ -24,30 +24,38 @@
...
         // 1.Retrieve the existing key dictionary from app group
-        var keyDict = sharedDefault.dictionary(forKey: keyTag) as? [String: Date]
+        var prevKeyDict = sharedDefault.dictionary(forKey: keyTag)
 
         // If this is the first time we store the key, we'll need to create a new key dictionary
-        if keyDict == nil {
-            keyDict = [:]
+        if prevKeyDict == nil {
+            prevKeyDict = [:]
         }
```
tjprescott, line 52: You can just do this:
```swift
var prevKeyDict = sharedDefault.dictionary(forKey: keyTag) ?? [:]
```

PR #1312, opened 2022-07-12
[Feature] Introduce new method createCommunicationIdentifier(from rawId:)

JoshuaLai:
Introducing new method to serialize and deserialize the communication identifier the raw id. 
Follow up from feedback from PR(https://github.com/Azure/azure-sdk-for-ios/pull/1255)

Swift Interface: 
```
        let acsRawId = "8:acs:37691ec4-57fb-4c0f-ae31-32791610cb14_37691ec4-57fb-4c0f-ae31-32791610cb14"
        var communicationUserIdentifier = createCommunicationIdentifier(from: acsRawId)
        XCTAssertTrue(communicationUserIdentifier.isKind(of: CommunicationUserIdentifier.self))
        XCTAssertEqual(communicationUserIdentifier.kind, .CommunicationUser)
        XCTAssertEqual(communicationUserIdentifier.rawId, acsRawId)
```

ObjC Interface:
```
    NSString *acsRawId = @"8:acs:37691ec4-57fb-4c0f-ae31-32791610cb14_37691ec4-57fb-4c0f-ae31-32791610cb14";
    id<CommunicationIdentifier> identifier = [CommunicationIdentifierFactory createCommunicationIdentifier:acsRawId];
    XCTAssertTrue([identifier isKindOfClass:CommunicationUserIdentifier.class]);
    XCTAssertEqual(identifier.kind, IdentifierKind.CommunicationUser);
    XCTAssertEqualObjects(identifier.rawId, acsRawId);
    XCTAssertEqualObjects(((CommunicationUserIdentifier *)identifier).identifier, acsRawId);
```

PR #1335, opened 2022-08-16
Push Notification GA release

angellan-msft:
This PR is open to track any change needed for GA release.


sdk/communication/AzureCommunicationChat/Source/NotificationUtils/IdentityUtil.swift
```diff
@@ -68,3 +70,66 @@ func isGcch(id: String) -> Bool {
...
+            phoneNumber: parse(id: id, prefix: phoneNumberPrefix),
+            rawId: id
+        )
+    } else if id.starts(with: acsUserPrefix) || id.starts(with: spoolUserPrefix) || id
+        .starts(with: dodAcsUserPrefix) || id.starts(with: gcchAcsUserPrefix) {
+        return CommunicationUserIdentifier(id)
+    } else {
+        return UnknownIdentifier(id)
+    }
+}
```
petrsvihlik, line 135: This will need to get replaced with https://github.com/Azure/azure-sdk-for-ios/pull/1312/files#diff-a7d2b1adee8e4923bdf6478851842c5835a9a1bcfae3b364b7563eaa48070164 which is currently blocked by architects because of some issues with enums. We have a solid concept that we now need to test.

@JoshuaLai is about verify that it works in our scenario and if so, we'll get our PR approved by ARB and merged.

petrsvihlik (changes requested):

petrsvihlik:
@angellan-msft other than my only comment, the PR looks fine 👍

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/IdentityUtil.swift
angellan-msft: Thanks @petrsvihlik for the suggestion. I see. It would be better to reuse existing methods in Common library. 
However our team targets to release Push Notification GA on Sep 15 and it will take around 2 weeks to prepare everything before release. I discussed this with our PM and EM. We would like to wait until Aug 30 to see if this method in Common library is ready to use. If this Common PR is not merged before Aug 30, we have to go ahead with GA release. But we will still create a work item for this fix and work on it after GA release. Since it is an internal code change, we believe it will not be a breaking change. Let me know if you have any concerns. Thanks!

tjprescott (approved):

petrsvihlik (approved):

PR #1397, opened 2022-11-17
A patch release for AzureCommunicationCommon 1.1.1

AikoBB:
- Fixed the logic of `PhoneNumberIdentifier` to always maintain the original phone number string whether it included the leading + sign or not.

### DIFF
--- a/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
+++ b/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
@@ -1,6 +1,6 @@
 {
   "name": "AzureCommunicationCommon",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "summary": "Azure Communication Service common client library for iOS",
   "description": "This package contains common code for Azure Communication Services\nlibraries.",
   "homepage": "https://github.com/Azure/azure-sdk-for-ios",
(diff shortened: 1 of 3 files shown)


mjafferi-msft (approved):

tjprescott:
Why is this considered a bug? What benefit is there to not standardize the phone number?

AikoBB:
> Why is this considered a bug? What benefit is there to not standardize the phone number?

@tjprescott  While using the `createIdentifierFromRawId` by passing an e164 formatted number such as '+123456789` the response is a PhoneNumberIdentifier, but the phoneNumber property of the PhoneNumberIdentifier has two ++ at the beginning, like '++123456789' and it caused issue for our customers, so we decided to do not add plus sign internally and fixed it in this PR https://github.com/Azure/azure-sdk-for-ios/pull/1391

tjprescott (changes requested):
Preventing merging of this PR

petrsvihlik:
@tjprescott the problem mentioned by @AikoBB is just the tip of the iceberg. 

The main motivation behind #1391 was that the MRI format without a `+` sign wasn't canonical to ACS and IC3 services. Moreover, we found out that some of the IC3 infra works with `4:` MRIs such as `4:{guid}`, `4:{E.164}_{guid}`, `4:{guid}_{guid}`. So, if we ever wanted to support other than E.164-based MRIs we shouldn't assume the PhoneNumber MRI format and keep whatever is after the leading `4:`.

CC'ing @williamzhao87 (who originally called out the issue) and @DominikMe (who we've agreed on the approach with) to add any more context.

tjprescott (approved):

PR #1400, opened 2022-11-18
[Feature] Provide async wrappers for PagedCollection

emlynmac:
With swift `async/await` available in more recent versions of iOS, we can benefit from async wrappers at the core level in other components.

This PR introduces minimal functionality to address async paging of results.

### DIFF
--- a/sdk/core/AzureCore/CHANGELOG.md
+++ b/sdk/core/AzureCore/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 1.0.0-beta.16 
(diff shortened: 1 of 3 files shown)


mpodwysocki (approved):
Looks fine to me

tjprescott:
@mpodwysocki should we create an issue to track this:

> Are we going to use the overall AsyncIterator protocol? [AsyncIteratorProtocol | Apple Developer Documentation](https://developer.apple.com/documentation/swift/asynciteratorprotocol)

PR #1401, opened 2022-11-22
Fix Test result reporting for failed tests

weshaggard:


benbp:
Is this actively testing? Why is the whole client test yaml commented out?

PR #1402, opened 2022-11-23
Disabled flaky tests of the AzureCommunicationCommon

AikoBB:
The three test that are disabled runs properly locally, but during the pipeline run sometimes they fail which is the known issue. I am disabling these tests untill we will change the token autorefreshing logic and update the tests regarding the new logic. The ADO already created for it -> [User Story 3051451](https://skype.visualstudio.com/SPOOL/_workitems/edit/3051451): [SDK][Common][iOS] Add a backoff mechanism for the CommunicationTokenCredential

### DIFF
--- a/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.xcodeproj/xcshareddata/xcschemes/AzureCommunicationCommon.xcscheme
+++ b/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.xcodeproj/xcshareddata/xcschemes/AzureCommunicationCommon.xcscheme
@@ -37,6 +37,17 @@
                BlueprintName = "AzureCommunicationCommonTests"
                ReferencedContainer = "container:AzureCommunicationCommon.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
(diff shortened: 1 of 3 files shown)


petrsvihlik (approved):
since our changes haven't touch anything with regards to the autorefresh logic, I'm approving this. we'll solve the flaky tests in the next sprint

JoshuaLai:
If we are going to skip these tests should we skip all the Obj-C tests? 
`test_ObjCRefreshTokenProactively_FetchTokenReturnsError`

petrsvihlik:
> If we are going to skip these tests should we skip all the Obj-C tests? `test_ObjCRefreshTokenProactively_FetchTokenReturnsError`

Why to skip passing tests?

tjprescott (approved):

JoshuaLai:
> > If we are going to skip these tests should we skip all the Obj-C tests? `test_ObjCRefreshTokenProactively_FetchTokenReturnsError`
> 
> Why to skip passing tests?

Well we are not sure they are always passing. Just because they are passing this instance it may fail the next one. I think this is a pattern for all the ObjC tests.

PR #1435, opened 2023-01-04
Added Jazzy Config for ACS UI Chat

jpeng-ms:
As we are close to official release of ACS UI Chat, I am adding Jazzy config for it so API reference doc would be available after release

### DIFF
--- a/jazzy/AzureCommunicationUICalling.yml
+++ b/jazzy/AzureCommunicationUICalling.yml
@@ -3,7 +3,7 @@ author_url: https://azure.github.io/azure-sdk/
 github_url: https://github.com/Azure/communication-ui-library-ios
 module: AzureCommunicationUICalling
 module_version: 1.1.0
-readme: ../build/communication-ui-library-ios/README.md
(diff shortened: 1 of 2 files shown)


JoshuaLai (approved):

jazzy/AzureCommunicationUICalling.yml
```diff
@@ -3,7 +3,7 @@ author_url: https://azure.github.io/azure-sdk/
 github_url: https://github.com/Azure/communication-ui-library-ios
 module: AzureCommunicationUICalling
 module_version: 1.1.0
-readme: ../build/communication-ui-library-ios/README.md
+readme: ../build/communication-ui-library-ios/AzureCommunicationUI/sdk/AzureCommunicationUICalling/README.md
```
jpeng-ms, line 6: these links to readme might not be available until we merge chat to main on release day

tjprescott: That's fine as long as we don't try to regenerate the docs until then. So as long as UICalling doesn't release beforehand there won't be an issue.

tjprescott (approved):

PR #1456, opened 2023-01-26
Introducing a public cancel API for the CommunicationTokenCredential

AikoBB:
# Description
In the common sdk we are providing [token autorefresh](https://github.com/Azure/azure-sdk-for-ios/blob/main/sdk/communication/AzureCommunicationCommon/Source/Authentication/ThreadSafeRefreshableAccessTokenCache.swift) functionality by the [CommunicationTokenCredential](https://github.com/Azure/azure-sdk-for-ios/blob/45e35a9c220b28fe1b750043190359d35f9a8711/sdk/communication/AzureCommunicationCommon/Source/Authentication/CommunicationTokenCredential.swift#L35) and this credential object can be passed to multiple Chat or Calling client instances.  The SDK will make no assumptions about its lifetime and leaves the responsibility of its disposal(disposing of the credential will also cancel scheduled refresh actions when the proactive refreshing is enabled) to the developer and in other platforms they have public API for disposing it([java](https://github.com/Azure/azure-sdk-for-java/blob/354f5c7e9b8a8c0d035d1287ffffecaa57b2c37b/sdk/communication/azure-communication-common/src/main/java/com/azure/communication/common/CommunicationTokenCredential.java#L120), .[net](https://github.com/Azure/azure-sdk-for-net/blob/1395725b885c9a2523580f017bd1227a89738d4c/sdk/communication/Azure.Communication.Common/src/CommunicationTokenCredential.cs#L66), [python](https://github.com/Azure/azure-sdk-for-python/blob/498b0365dd218d3210d03b13993f2505e996b1a3/sdk/communication/azure-communication-identity/azure/communication/identity/_shared/user_credential.py#L141), [js](https://github.com/Azure/azure-sdk-for-js/blob/a9a944628e4d76f2e0cb617e751d87f6f0a3de25/sdk/communication/communication-common/src/azureCommunicationTokenCredential.ts#L58)) but in the iOS cancellation of the scheduled action is done only in the [deinitializer](https://github.com/Azure/azure-sdk-for-ios/blob/45e35a9c220b28fe1b750043190359d35f9a8711/sdk/communication/AzureCommunicationCommon/Source/Authentication/ThreadSafeRefreshableAccessTokenCache.swift#L147), which means it will be disposed only when the swift handles automatically deallocation of instances. This PR addresses inconsistencies in Swift language and was introduced public `cancel` API:
API view - [AzureCommunicationCommon (version 1.2.0) ](https://apiview.dev/Assemblies/Review/7296d5b56972409aa25807998d1a772d?diffRevisionId=1392f4b12890455a8f56d322a67c39f6&diffOnly=False&revisionId=192bb3a066bb4fe58023855c074c7ad5&doc=False)

### DIFF
--- a/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
+++ b/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
@@ -24,10 +24,6 @@
   "pod_target_xcconfig": {
     "DEFINES_MODULE": "YES"
   },
-  "dependencies": {
-      "AzureCore": [
-        "1.0.0-beta.15"
-      ],
-  },
+  "dependencies": {},
   "swift_version": "5.0"
 }
--- a/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
+++ b/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
@@ -24,6 +24,10 @@
   "pod_target_xcconfig": {
     "DEFINES_MODULE": "YES"
   },
-  "dependencies": {},
+  "dependencies": {
+      "AzureCore": [
+        "1.0.0-beta.15"
+      ]
+  },
   "swift_version": "5.0"
 }
--- a/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
+++ b/sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
@@ -24,10 +24,6 @@
   "pod_target_xcconfig": {
     "DEFINES_MODULE": "YES"
   },
-  "dependencies": {
-      "AzureCore": [
-        "1.0.0-beta.15"
-      ]
-  },
+  "dependencies": {},
   "swift_version": "5.0"
 }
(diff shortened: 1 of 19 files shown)


AikoBB:
@mpodwysocki @tjprescott The [Cancellable](https://developer.apple.com/documentation/combine/cancellable) is only available in iOS 13.0+, I was wondering if is there any way to provide backward compatibility or if it’s better to provide our own Cancellable protocol?

mpodwysocki (approved):

petrsvihlik (approved):

ostoliarova-msft (approved):

sdk/communication/AzureCommunicationCommon/AzureCommunicationCommon.podspec.json
```diff
@@ -24,6 +24,10 @@
   "pod_target_xcconfig": {
     "DEFINES_MODULE": "YES"
   },
-  "dependencies": {},
+  "dependencies": {
+      "AzureCore": [
+        "1.0.0-beta.15"
+      ]
+  },
```
tjprescott, line 31: Common should not take a dependency on AzureCore because it is used by Calling (ObjC based) and AzureCore is fundamentally not compatible with ObjC codebases.

AikoBB: Thanks, fixed

mpodwysocki (approved):
LGTM

PR #1569, opened 2023-06-30
LeapSecond700

epworth:


tjprescott:
There should not be changes to engineering system files in this PR.

tjprescott (changes requested):
What is the purpose of this PR? Why are there changes to engineering system files?

lucianopa-msft:
Looks like this PR doesn't do anything actually, it updates branch AzureCommunicationCalling_2.4.1 with main, but given that AzureCommunicationCalling_2.4.1 was merged on main already, I think this PR is moot

Issue #828, opened 2021-04-21
AzureCommunicationCalling: Guidelines Exceptions and Release Blockers


tjprescott:
These are the current deviations AzureCommunicationCalling has from the guidelines. 

**Blocking for AzureCommunicationCalling 2.6.0**
- [x] Support M1 processors ($ARCH_STANDARD (arm64))
- [x] `MediaDiagnosticsDelegate` and `NetworkDiagnosticsDelegate` methods should not end in `*Value`. [link](https://apiview.dev/Assemblies/Review/994dc64224614433a2685a0a75e184c0?diffRevisionId=909ebdc64aaa4241a9bdacc1683305eb&diffOnly=False&revisionId=f80562c8bad54b6d9674cd61504bf06f&doc=False)

**Blocking for AzureCommunicationCalling 2.7.0**
- [ ] SwiftPM support for AzureCommunicationCalling
- [ ] `RecordingInfo` must expose some kind of recording ID
- [ ] Deprecate `onIsRecordingActiveChanged` (tied to `RecordingInfo`?)
- [ ] AudioOptions.muted was deprecated in 2.5.1 but is "un-deprecated" in 2.6.0-beta.1 [link](https://apiview.dev/Assemblies/Review/994dc64224614433a2685a0a75e184c0?diffRevisionId=909ebdc64aaa4241a9bdacc1683305eb&diffOnly=False&revisionId=f80562c8bad54b6d9674cd61504bf06f&doc=False)
- [ ] `Call.add` event is not in `CallBase`, whereas all other event methods were moved there in 2.6.0-beta.1 [link](https://apiview.dev/Assemblies/Review/994dc64224614433a2685a0a75e184c0?diffRevisionId=909ebdc64aaa4241a9bdacc1683305eb&diffOnly=False&revisionId=f80562c8bad54b6d9674cd61504bf06f&doc=False)
- [ ] `CallBase.applyOutgoingVideo(constraints:)` is a non-standard name. Need resolution on final name before GA. [link](https://apiview.dev/Assemblies/Review/994dc64224614433a2685a0a75e184c0?diffRevisionId=909ebdc64aaa4241a9bdacc1683305eb&diffOnly=False&revisionId=f80562c8bad54b6d9674cd61504bf06f&doc=False)
[...]

PR #1584, opened 2023-07-20
AzureCommunicationCalling 2.6.0-beta.1

xumo-95:
https://apiview.dev/Assemblies/Review/994dc64224614433a2685a0a75e184c0


tjprescott (changes requested):
See APIView.

xumo-95:
@tjprescott Thanks for the comments. We would address them in the next beta release or when we promote the features to GA.  In that case can we proceed with this BETA release?

tjprescott (approved):
APIView is not approved due to deficiencies in the API which are being tracked in https://github.com/Azure/azure-sdk-for-ios/issues/828. However, since this is a beta, the release is approved.

PR #1602, opened 2023-08-25
Roll back to macOS 11

tjprescott:
Diagnose build system failure


eng/pipelines/templates/variables/globals.yml
```diff
@@ -1,5 +1,5 @@
 variables:
-  OSVmImage: 'macOS-12'
+  OSVmImage: 'macOS-11'
   DocWardenVersion: '0.7.1'
   XcodeVersion: '13.3'
```
mikeharder, line 4:
```suggestion
  XcodeVersion: '13.2'
```

PR #1603, opened 2023-08-25
Test macOS 13

mikeharder:


tjprescott:
We are able to overcome this without bumping the iOS version, so I will pursue that first.

PR #1468, opened 2023-02-09
[iOS] Added support for MicrosoftBotIdentifier

AikoBB:
# Description

Introducing a new type of CommunicationIdentifier - MicrosoftBotIdentifier and the logic is covered with tests.
Implementation of the [User Story 3133111](https://skype.visualstudio.com/SPOOL/_workitems/edit/3133111): [SDK][Common][Mobile] As a developer I can identify if a user is a Teams User or Bot)

PR #1630, opened 2023-10-10
[Communication][Chat] Revert reference on .cloudEnvironment to its previous name

cheukchuen:
`.cloudEnvironment` was being introduced in AzureCommunicationCommon version `1.2.0-beta.1`. This PR reverts the reference back to `.cloudEnviroment` since AzureCommunicationChat still references AzureCommunicationCommon version `~> 1.0`. This resolves the following error in our release pipeline:
> \- ERROR | xcodebuild:  AzureCommunicationChat/sdk/communication/AzureCommunicationChat/Source/Serializer/IdentifierSerializer.swift:126:30: error: value of type 'MicrosoftTeamsUserIdentifier' has no member 'cloudEnvironment'

### DIFF
--- a/sdk/communication/AzureCommunicationChat/Source/NotificationUtils/IdentityUtil.swift
+++ b/sdk/communication/AzureCommunicationChat/Source/NotificationUtils/IdentityUtil.swift
@@ -70,13 +70,3 @@ func isGcch(id: String) -> Bool {
     let gcchAcsUserPrefix = "gcch-acs:"
     return (id.starts(with: gcchTeamsUserPrefix) || id.starts(with: gcchAcsUserPrefix))
 }
-
-/// Parses out the id/phone number portion of a user id.
-/// - Parameters:
-///   - id: The string id.
-///   - prefix: The id prefix.
-/// - Returns: The part of the id after the prefix that corresponds to the user id or phone number of a user.
-private func parse(id: String, prefix: String) -> String {
-    let index = id.index(id.startIndex, offsetBy: prefix.count)
-    return String(id.suffix(from: index))
-}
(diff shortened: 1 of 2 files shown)


angellan-msft (approved):

sdk/communication/AzureCommunicationChat/Source/NotificationUtils/IdentityUtil.swift
```diff
@@ -70,13 +70,3 @@ func isGcch(id: String) -> Bool {
...
-
-/// Parses out the id/phone number portion of a user id.
-/// - Parameters:
-///   - id: The string id.
-///   - prefix: The id prefix.
-/// - Returns: The part of the id after the prefix that corresponds to the user id or phone number of a user.
-private func parse(id: String, prefix: String) -> String {
-    let index = id.index(id.startIndex, offsetBy: prefix.count)
-    return String(id.suffix(from: index))
-}
```
tjprescott, line 82: Why is this being removed?

cheukchuen: It's no longer being used, unless we want to bring everything back from https://github.com/Azure/azure-sdk-for-ios/pull/1468/files#diff-be387cd2134ba545b2160012a828b3b8c544eb948c33dc0228d57bd698a28689, including `getIdentifier(from:)`?

tjprescott: Is this something that will be readded when the references to `cloudEnviroment` are updated later on? If so, we should probably just leave this in. It's private after all.

cheukchuen: No, we'll be utilizing Common's `createCommunicationIdentifier(fromRawId:)` instead. Difference between the 2 versions is the support for MicrosoftBotIdentifier in `1.2.0-beta.1`

tjprescott: Okay, so this has been dead code already then?

PR #1667, opened 2023-12-12
Fix for "supports docs-only Jazzy and adds Vision docs (#1664)" commit

nabilat:
Adding missing property to enable doc-only jazzy uploads.


tjprescott (changes requested):
The actual script needs to be updated to not use this "no_jazzy_mode" property.

Issue #1672, opened 2024-01-04
[BUG] Push Notifications: 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.


mattials:
**Describe the bug**
Encountered the error message: "'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead." in ApplicationUtil.swift line 49 -> return UIApplication.shared

***Exception or Stack Trace***
'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.

**To Reproduce**
Followed the Microsoft Azure tutorial [link](https://learn.microsoft.com/en-us/azure/communication-services/tutorials/add-chat-push-notifications#2---advanced-version) for integrating push notifications into a chat application.
Encountered the error message: "'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead."

***Code Snippet***
ApplicationUtil.swift line 49 -> return UIApplication.shared

**Expected behavior**
Successful compilation without errors on ./add-chat-push-notifications in xcode 15.

**Screenshots**
<img width="1128" alt="Issue" src="https://github.com/Azure/azure-sdk-for-ios/assets/119877139/dfc12127-5838-478f-aa9e-1a9773105da4">

**Setup (please complete the following information):**
 - OS: iOS 17.0.2
 - IDE : xCode 15
 - pod 'AzureCommunicationChat', '~> 1.3.3'

**Additional context**
Encountered the same issue in both a direct integration into the chat project and when using the Microsoft-provided example project [link](https://github.com/Azure-Samples/communication-services-ios-quickstarts/tree/main/add-chat-push-notifications). The issue appears to be related to the AzureCore dependency, and a possible solution was suggested in the Apple Developer Forums [link](https://developer.apple.com/forums/thread/685369), recommending the use of @available(iOSApplicationExtension, unavailable). This change may need to be implemented in the Azure SDK to resolve the issue.

https://github.com/Azure-Samples/communication-services-ios-quickstarts/issues/70

tjprescott:
@angellan-msft can you please take a look.

angellan-msft:
@mattials , I just returned to office and will look into it next week.

mattials:
@angellan-msft @tjprescott I created a PR, I tested locally to fix the error, can you check it please 🙏 

https://github.com/Azure/azure-sdk-for-ios/pull/1673

angellan-msft:
[...]

PR #1673, opened 2024-01-08
Fix [BUG] https://github.com/Azure/azure-sdk-for-ios/issues/1672

mattials:
https://github.com/Azure/azure-sdk-for-ios/issues/1672

Use of @available(iOSApplicationExtension, unavailable) for Xcode 14 and 15

### DIFF
--- a/sdk/core/AzureCore/Source/Util/ApplicationUtil.swift
+++ b/sdk/core/AzureCore/Source/Util/ApplicationUtil.swift
@@ -25,9 +25,7 @@
 // --------------------------------------------------------------------------
 
 import Foundation
-#if canImport(UIKit)
-    import UIKit
-#endif
+import UIKit
 
 public enum ApplicationUtil {
     // MARK: Static Methods
@@ -41,9 +39,8 @@ public enum ApplicationUtil {
         return mainBundlePath.hasSuffix("appex")
     }
 
-    #if canImport(UIKit)
-
         /// Simple access to the shared application when not executing within an app extension.
+        @available(iOSApplicationExtension, unavailable)
         public static var sharedApplication: UIApplication? {
             guard !isExecutingInAppExtension else { return nil }
             return UIApplication.shared
@@ -52,6 +49,7 @@ public enum ApplicationUtil {
         /// Returns the current `UIViewController` for a parent controller.
         /// - Parameter parent: The parent `UIViewController`. If none provided, will attempt to discover the most
         /// relevant controller.
+        @available(iOSApplicationExtension, unavailable)
         public static func currentViewController(forParent parent: UIViewController? = nil) -> UIViewController? {
             // return the current view controller of the parent
             if let parent = parent {
@@ -69,6 +67,7 @@ public enum ApplicationUtil {
 
         /// Attempt to find the top-most view controller for a given root view controller.
         /// - Parameter root: The root `UIViewController`.
+        @available(iOSApplicationExtension, unavailable)
         public static func currentViewController(withRootViewController root: UIViewController?) -> UIViewController? {
             if let tabBarController = root as? UITabBarController {
                 return currentViewController(withRootViewController: tabBarController.selectedViewController)
@@ -80,6 +79,4 @@ public enum ApplicationUtil {
                 return root
             }
         }
-
-    #endif
 }


angellan-msft:
@mattials 
1.I'm able to see push notification coming in when using the sample app. Could you please provide a more detailed description on what steps you were going through and what error did you see on a particular step? And which XCode version are you using?
2.Thanks for suggesting the code change to "azure core" file. I have 3 questions -
1) How did you come to the conclusion that the bug was in azure core?
2) I noticed the IOS core CI pipeline failed with this suggested change. I believe this code change can't be put in use until the CI pipeline runs successfully.
3) I'm not sure if external contributor has the permission to merge any code changes. We will need some guidance from our IOS azure reviewer @tjprescott on this question. Maybe someone from Azure core should make this change (if this change is needed)?

mattials:
@angellan-msft,

Thank you for your prompt response. I understand that there might be some confusion regarding the necessity of the code change in the "azure core" file. Let me provide you with a detailed explanation and address the concerns raised

I always get the same error in your example https://github.com/Azure-Samples/communication-services-ios-quickstarts/tree/main/add-chat-push-notifications and following your documentation on advanced push notifications

Testing Environment:
I conducted tests on different machines with varying configurations to ensure the universality of the error and the effectiveness of the proposed fix. The tested environments include:

Apple M2, macOS 13.5.2, XCode 15.0
Intel i5 2019, macOS 13.5.2, XCode 15.2

##First
It will probably work in an already compiled app or using the simple notifications method, but the repository https://github.com/Azure-Samples/communication-services-ios-quickstarts/tree/main/add-chat-push-notifications doesn't compile, the first bug the repository gives with XCode 15 is:

DT_TOOLCHAIN_DIR cannot be used to evaluate LIBRARY_SEARCH_PATHS, use TOOLCHAIN_DIR instead
This has an easy solution, change the Podfile and it's not a concern.

I always get the same error on any machines trying to compile the test project https://github.com/Azure-Samples/communication-services-ios-quickstarts/tree/main/add-chat-push-notifications, if locally we apply the fix I have in this PR it works correctly.

MicrosoftTeams-image (2)

##Second
I have checked your pipe and the tests fail:

✗ test_ShouldBeCalledAgainAfterFirstRefreshCall, XCTAssertEqual failed: ("2") is not equal to ("1")
✗ test_ShouldBeCalledAgainAfterFirstRefreshCall, XCTAssertEqual fallido: ("3") no es igual a ("2")

The test relies on asynchronous execution using DispatchQueue.asyncAfter, and the timing might not be precise. If the token refresh takes longer than expected or if there are delays in the execution of the asynchronous blocks, the assertions may not be accurate.

The test involves asynchronous operations and multiple DispatchQueue calls. If there are shared resources or state changes that are not handled properly in a concurrent environment, it could lead to unexpected behavior.

tjprescott:
I'm reluctant to accept any change to `Azure.Core` if the issue could simply be the example documentation. Are you following the example verbatim? @angellan-msft do you know what the expected environment was when that sample was written? Are you able to repro the issue?

tjprescott:
The AzureCommunicationCommon units tests are known to be flaky. I've re-run the CI which should clear them up.

tjprescott:
Okay, the checks pass. @angellan-msft please let me know the results of your investigation.

angellan-msft:
@mattials Thanks for your detailed description. So the issue you met with is the advanced version of push notification, which I have not tested yet. What I tested is the basic version. I'll try to reproduce the issue today and see if this fix works. Will provide an update ASAP.

angellan-msft:
hi @mattials , could you try the below workaround to see if it can solve your issue:

**Set "Require Only App-Extension-Safe API" as "No" for Pod Target - AzureCore.**
Why we need this: [https://stackoverflow.com/questions/48122769/facebook-cocoapods-sharedapplication-is-unavailable-not-available-on-ios-app (Opens in new window or tab)](https://stackoverflow.com/questions/48122769/facebook-cocoapods-sharedapplication-is-unavailable-not-available-on-ios-app)

I just found that this workaround was documented in our internal wiki but not in public wiki. If it works for you, I would suggest this workaround rather than making extra code changes to azure core. Thanks!

mattials:
@angellan-msft @tjprescott 
Thank you for providing the suggested workaround. I have implemented the fix by setting "Require Only App-Extension-Safe API" to "No" for the Pod Target - AzureCore, and it appears to have resolved the compilation issue. I appreciate your prompt assistance.

While the workaround addresses the immediate concern, our team is committed to ensuring the long-term stability of our product. Therefore, we are interested in contributing to a more permanent solution by updating AzureCore. Could you please provide insights into when a stable version of AzureCore, beyond the current 1.0.0-beta.15, is expected to be released? This information will be valuable for our planning and integration efforts.

Thank you once again for your support, and we look forward to continued collaboration.

tjprescott (approved):

PR #1885, opened 2024-09-18
Infrastructure: OSMP API audience change

jeffwilcox:
The audience value for an Entra ID app call is
changing. This shifts to the new value.


weshaggard:
Replaced by https://github.com/Azure/azure-sdk-tools/pull/9004

PR #1948, opened 2024-11-08
Upgrade python < 3.11 to 3.11

JimSuplizio:


JimSuplizio:
I'm going to merge without waiting for approval. This change is really basic and the ios - client - ci run fully passes and python is only used in the analyze step.

PR #2051, opened 2025-03-12
Resolve 1ES Release warning

### LINKED ISSUE #2008: Address 1ES release warning
We want to eliminate [this warning](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4499458&view=logs&j=3e633120-f02c-5388-8b0c-169262c3e98e&t=a1fc4e68-3536-55e2-9cc0-649ab8b629a0&l=16) that is appearing on ALL go release pipelines. The example given is for CPP, and we haven't released the ios packages in a while. Still probably need to deal with this.

Reference [this PR](https://github.com/Azure/azure-sdk-for-python/pull/39480) for a summary of the necessary changes.

Need to:

- [ ] Eliminate any `download` or `checkout` usage in the `deployment` job
- [ ] Ensure that only the `release` job in the stage is actually a `deployment` with `templateContext` setting it to a release job. The rest should be moved to regular `job` instead of `deployments`

We need to update

- [ ] archtype-android-release

scbedd:
Resolves #2008 

Hilariously enough @weshaggard the first attempt to release _100%_ hit that SAML error you were mentioning on your way out today. I'm retrying it now!

[Example successful release](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4646258&view=results)

### DIFF
--- a/eng/pipelines/templates/jobs/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/jobs/archetype-sdk-client.yml
@@ -183,6 +183,8 @@ jobs:
     variables:
       - template: ../variables/globals.yml
 
+    condition: and(succeeded(), ne(variables['Skip.Test'], 'true'))
+
     pool:
       vmImage: $(MACVMIMAGE)
       name: $(MACPOOL)
--- a/eng/pipelines/templates/stages/archetype-sdk-client.yml
+++ b/eng/pipelines/templates/stages/archetype-sdk-client.yml
@@ -45,6 +45,9 @@ extends:
       policy: M365
       credscan:
         toolVersion: 2.3.12.23
+      sourceRepositoriesToScan:
+        exclude:
+          - repository: SwiftPM-AzureTemplate
 
     stages:
       - stage: Build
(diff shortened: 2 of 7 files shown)


eng/pipelines/templates/stages/archetype-sdk-client.yml
```diff
@@ -45,6 +45,9 @@ extends:
       policy: M365
       credscan:
         toolVersion: 2.3.12.23
+      sourceRepositoriesToScan:
+        exclude:
+          - repository: SwiftPM-AzureTemplate
```
weshaggard, line 50: Why did you exclude this?

weshaggard (approved):

eng/pipelines/templates/jobs/archetype-sdk-client.yml
```diff
@@ -183,6 +183,8 @@ jobs:
     variables:
       - template: ../variables/globals.yml
 
+    condition: and(succeeded(), ne(variables['Skip.Test'], 'true'))
```
weshaggard, line 186: This looks like an errand copy/paste.

scbedd: Wasn't errant. It's not a negative (this isn't set to true by default) but I needed to get through the failing unit tests.

eng/pipelines/templates/stages/archetype-sdk-client.yml
scbedd: To avoid compilation errors when having an additional checkout.

scbedd: I'll merge this now and see if I can reverse this to explicitly include it.

eng/pipelines/templates/jobs/archetype-sdk-client.yml
weshaggard: I couldn't tell where this applies and now see it is attached to the Job it just looks odd floating out here.

PR #2450, opened 2026-04-17
Remove Check-Enforcer

scbedd:
See title. `ios - core - ci` is now the required check.


tjprescott (approved):

scbedd:
This is not the strat. Looks like it's coming through eng/common and you can't avoid it rn. Just gonna disable workflow.

